PR #433, opened 2015-12-31
some work :D

leftyde:
vielleicht kannst du mit meinem OMT sachen noch was anfangen^^


leftyde:
okay irgendwie hab ich das mit den Pull Request noch net raus :D^^

Dream-Master:
@leftyde 
I think your recipes are a bit hard Open Modular Turrets is a fun mod. Yes i like hard recipes but i think will change them a bit. Some of your ideas are good and i will ad these to the script. Thanks for your work!

PR #478, opened 2016-02-20
Lower necessary sleep percentage to 33

Teteros:
Should we lower it a bit?
To be discussed.


Dream-Master:
no this i not add now. 
Wait for GTNH 2

Teteros:
Lowered to 25% as per GTNH#2

PR #593, opened 2016-05-14
Solve localization Issue

siriulx:
Maybe it will be better to avoid using specific language code in localization text. This change will make all the text display normally in different language environment, instead of only showing the keys.


Dream-Master:
Let me test it and after it i will merge it.

PR #661, opened 2016-08-10
EMT Buff

SanMan911:
Buffed the solars to 10x as discussed with @DreamMasterXXL earlier. Could take effect when IC2 recipes are changed accordingly.


Technus:
Even if that would not work since the solars can output max 2048 due to tier limit somewhere.

Technus:
Power gen rebalance - soon(tm)

Dream-Master:
When i am nerfing this mod we have to look into it again. Maybe have to change code wise.

Dream-Master:
I will go over it soon

PR #1027, opened 2017-01-20
Disable motorized boat

Ugachaga:
it does nothing in the current state, only possible bugs


Dream-Master:
@Ugachaga So the floating Boot without a Motor have the same speed ?
I try this Boot today not bug.

Ugachaga:
Yes, the same speed. Everything the same but motor model lol 
You can't open GUI or interact with Fueltank. It's just not implemented.
I've found tons of inflatable boat bugs, like stuck, desinc and a lot more. I'd pref just fully disable all inflatable boats, modpack has same indestructible IC2 boat, wich is more stable I believe.

I upload [beta25a](https://github.com/runescapejon/AdventureBackpack2/releases/tag/Beta-0.9fb25)

PR #1035, opened 2017-01-21
Focus Upgrade id conflict

Ugachaga:
```
[18:44:53] [Client thread/FATAL]: Focus Upgrade id 42 already occupied. Ignoring.
[18:44:53] [Client thread/FATAL]: Focus Upgrade id 43 already occupied. Ignoring.
```
I have no idea how to see/dump focus upgrade IDs, but changing it to 142-143 seems works. No more log warn and Forbidden Magic focus upgrades available ingame.

### DIFF
--- a/config/ForbiddenMagic.cfg
+++ b/config/ForbiddenMagic.cfg
@@ -35,8 +35,8 @@ enchantments {
 
 
 "focus upgrades" {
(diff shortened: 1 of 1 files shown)


Dream-Master:
Ok thanks. Will merge it when I am back on a computer.

PR #1088, opened 2017-02-05
Prevent IC2+NEI console spam

Ugachaga:
https://github.com/GTNewHorizons/NewHorizons/issues/1084
Seems IC2 doesn't like "*" as reference to itemArmor damage. Changing it to 1 works fine, no more console spam. You may also want to add recipes with damage 26 (discharged).


Dream-Master:
yes thanks will add that and changed all to meta 26

PR #1282, opened 2017-04-11
Fixed NBT for Copter/Jetpack

Ugachaga:
Updated config for gbeta9


Ugachaga:
may need to tune fuel values

Dream-Master:
Ok will include it thanks @Ugachaga

PR #1307, opened 2017-04-17
Change Malachite quest to 4 separate quests

hhheeexxx:
Test case. Maybe it makes sense to add an initial info quest "find a vein" and add a final quest with a gift. Total 6 quests instead of 1.


Dream-Master:
I will look into it tomorrow.

You add 2 empty Quests with no content to the code.

hhheeexxx:
Fix that

PR #1527, opened 2017-06-18
Finally got it fixed, Just the creative names now

Ethryan:
Again:
Done with the lang file in resources folder since minetweaker can't handle color codes with it's displayName function.

just some small changes where i replaced (Creative) with (Tier x).

And did a few changes


Dream-Master:
Why make the first macerator Quest retreival. I want all or most of the Bronze Age Quests be crafting Tasks @Ethryan

PR #1536, opened 2017-06-19
Maybe fix for quest 768

Ethryan:
Replaced them with
16 Heavy Fuel Cell and
32 Light Fuel Cell since the cracked versions have been removed

you were to fast dream, i missed this fix :)


Dream-Master:
@Ethryan thanks

PR #1539, opened 2017-06-20
Bugfix with vanilla ender chest

Ethryan:
Fix so that it isn't disabled, so that quest Dimensional Chest (603) can be completed. (Still uncraftable)

rename of new quest

### DIFF
--- a/config/EnderStorage.cfg
+++ b/config/EnderStorage.cfg
@@ -8,7 +8,7 @@ anarchy-mode=false
 checkUpdates=false
 
 #Set to true to make the vanilla enderchest unplaceable.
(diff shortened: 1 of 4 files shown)


config/EnderStorage.cfg
```diff
@@ -8,7 +8,7 @@ anarchy-mode=false
 checkUpdates=false
 
 #Set to true to make the vanilla enderchest unplaceable.
-disable-vanilla=true
+disable-vanilla=false
```
Dream-Master: You can found enderchests in the End so far i know. If you enable it The enderchest mod not working any more.

Issue #53, opened 2015-03-02
WIKI


Dream-Master:
Drobac i see you take care about that right ?

Drobac:
Guess ill write it eventually. We do need all the server specific information in one place

Dream-Master:
Yes sure. Feel free do do some stuff. I go nerfing Forestry stuff.

PR #1607, opened 2017-07-03
Correcting spelling error in quest #53 "Hammer time v2"

damug:
Change "tree ingots" to "three ingots"


Dream-Master:
@damug thanks

PR #1720, opened 2017-07-24
Infinity Items denerf

Ethryan:
I only changed 2 lines in the recipes.cfg, but it added everything from the latest gregtech update.

trying to fix it according to https://github.com/GTNewHorizons/NewHorizons/issues/1649

image
 an example, 
image
and how to craft infinity plates (the time is 200 secs not 190).


Dream-Master:
For me looks good @Drobac ?

PR #1723, opened 2017-07-24
Natura/Harvestcraft crossmod compatibility fixes

damug:
A couple of commits to allow natura berries to be juiced with soft hammer. Also, add natura berries to oredict for listAllfruit and listAllberry allowing them to be substituted as any other fruit or berry in other harvestcraft recipes.  Finally, require sugar to make lemonade, because otherwise it's just lemon juice.


Dream-Master:
@damug thanks

Methes:
These were not oredicted for a reason, as in the past you were able to use juicer and natura berries were easier to mass produce in early game.
Nowadays this shouldn't be a problem IMO, so good change.

PR #1725, opened 2017-07-25
Equalize infernal brick and sandy brick recipes

damug:
Apologies for being unclear earlier, but this is what I was trying to, but failed adequately explain when i said make the two recipes roughly equal since they are both largely cosmetic blocks in [https://github.com/GTNewHorizons/NewHorizons/issues/1700](url)

Feel free to disregard if you so desire.

### DIFF
--- a/scripts/Railcraft.zs
+++ b/scripts/Railcraft.zs
@@ -1329,10 +1329,10 @@ mods.forestry.Carpenter.addRecipe(60, <liquid:seedoil> * 5000,
 <Railcraft:backpack.apothecary.t1>, <Railcraft:backpack.apothecary.t2>);
 
 // --- Infernal Block
-recipes.addShaped(<Railcraft:brick.infernal:2> * 8, [
-[<ore:stoneNetherBrick>, <minecraft:soul_sand>, <ore:stoneNetherBrick>],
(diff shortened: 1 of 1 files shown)


Dream-Master:
Did you think magma cream is to expensive? Maybe have two recipe one with 8 one with 2 . The brick changes are ok.

Ethryan:
I actually think the removal of the magma cream in this recipe makes it too easy. 
Keep the old recipe that gives you 8 infernal bricks

just my opinion.

damug:
How about this: 4 nether bricks 4 nether sand and magma cream for 4 infernal bricks. Then to keep the recipes similar change Sandy brick to 4 bricks 4 sand and a bucket of water for 4 Sandy bricks. I will make another pull request after work today.

Dream-Master:
ok sure i will look at it then @damug

Methes:
I'm personally perfectly fine with removing the magma cream from the recipe.
It's a decoration block, you can get tons of others very easily so why not this one.
2x nether brick block + 2x soul sand -> 4x infernal
2x sand + 2x brick block -> 4x sandy brick.
Don't make it too complicated if you want people to use something else than cobblestone.

PR #1748, opened 2017-07-28
A couple small quest book edits.

damug:
Sifter requires assembler to make item filters, so I changed the quest to require it too.

Also changed it so the tier 3 carbon mesh/plate quests wouldn't unlock well before you could complete them in tier 2.


Dream-Master:
will look later.

PR #1771, opened 2017-08-01
Sprinkler from OpenBlocks made more usable.

OneEyeMaker:
Hello, @Dream-Master 
Default setting of this device make it useless at all. I bumped all parameters. Now this sprinkler can be compared with basic watering can from ExtraUtilities, but automated.
Because of that, recipe changed (now this gadget available on mid/end MV age).
With these changes I recommend you to add quest for this (now *useful*) device. It (*in my horrible opinion*) fits into Forestry branch after quest "Your first multiblock farm".

//If you want, I'll try to add this quest myself.


Dream-Master:
Yes why not good addition.

PR #1780, opened 2017-08-03
Turn Rye and Oats into Flour

damug:
Found it odd that there were no recipes to turn Rye or Oats into Flour in either the Macerator or with a Mortar and Pestle.  So I fixed that.

Thanks in advance.

### DIFF
--- a/scripts/Gregtech.zs
+++ b/scripts/Gregtech.zs
@@ -515,6 +515,14 @@ recipes.addShaped(<gregtech:gt.metaitem.01:2881>, [
 recipes.addShaped(<gregtech:gt.metaitem.01:2881>, [
 [<ore:cropWheat>, null, null],
 [<ore:craftingToolMortar>, null, null]]);
+// -
+recipes.addShaped(<gregtech:gt.metaitem.01:2881>, [
+[<ore:cropRye>, null, null],
+[<ore:craftingToolMortar>, null, null]]);
+// -
+recipes.addShaped(<gregtech:gt.metaitem.01:2881>, [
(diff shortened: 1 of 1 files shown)


Dream-Master:
@damug  are you sure that Rye and Oats not have thier own flour type?

damug:
Alright, I don't know what to make of this.  A fresh install of 1.5.0.7 has a working shapeless recipe for flour from wheat, oats, barley and rye (and potatoes) for me.  These recipes should be removed though since at the front of the gregtech.zs script that shapeless recipe is removed with the following line:  recipes.removeShapeless(<gregtech:gt.metaitem.01:2881>);.  It's almost like the gregtech.zs script didn't fully run.

That in mind, as soon as I start dickering with the .zs file that shapeless recipe no longer appears, and the only shaped crafting recipes are the ones from later in the gregtech.zs file for wheat and barley (and rye and oats after I add them).  

Either way, there's no macerator recipe for flour from rye or oats.  Also, apparently you can't use oredict entries in macerator recipes and I've fixed that.

To your actual question, Natura adds wheat flour and barley flour, but neither of these are used in this modpack.  I don't see any specific flour for rye or oats.  Extra trees used to add rye grain, but that was removed.  There was nothing similar for oats that I am aware of.

damug:
As I looked at all this there are also 2 different macerator recipes for flour from wheat and I can't understand the reason for both, or where both are coming from.

image

image

Not really related to these changes, but it is related to flour in general

Dream-Master:
Ok I will check it later.

PR #1939, opened 2017-09-10
Changed recipes for Compartments

OneEyeMaker:
Recipes for Compartments (from Binnie Core) was **_too_** hard.
For example: Diamond Compartment required Tungstensteel Frame Box (maybe, block id/meta changed after time; if so - **_please, use OreDictionary_**). 
Also these recipes used strange syntax (maybe outdated); this PR fixes it.
Where possible recipes use OreDictionary.

P. S. @Dream-Master 
When you add missing in GT components, please, check OreDictionary tags of existing components.
For example: *Diamond Frame Box* should be registered as ```frameGtDiamond```.

### DIFF
--- a/scripts/Binnies-Core.zs
+++ b/scripts/Binnies-Core.zs
@@ -9,22 +9,22 @@
 // --- remove recipes ---
 
 // --- Compartment
-recipes.remove(<BinnieCore:storage>);
+recipes.removeShaped(<BinnieCore:storage>);
 
 // --- Copper Compartment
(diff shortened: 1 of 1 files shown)


Dream-Master:
@OneEyeMaker is this tested? Are you sure forestry machines work with oredict?

OneEyeMaker:
Everything is tested on freshly-downloaded GTNH-1.5.1.2.
As I said - previous version of recipes was strange (I guess, outdated). 
And work of OreDictionary I checked twice - in code of MineTweaker and in game.

PR #1958, opened 2017-09-13
A few changes that were not made/forgotten/repetitive

MineAnPlay:
Reverted MixedOreOnlyYieldsTwoThirdsOfPureOre to false. This means furnace recipes for pure dusts (non impure/purified dusts) will give an ingots worth of material and EBF recipes will give 3 ingots instead of 2. So overall makes the game more enjoyable and makes the EBF processing a must have for people who want to get the most out of there ores.

### DIFF
--- a/config/GregTech/GregTech.cfg
+++ b/config/GregTech/GregTech.cfg
@@ -45,7 +45,7 @@ general {
     D:MagneticraftBonusOutputPercent=100.0
     I:MaxEqualEntitiesAtOneSpot=3
     I:MillisecondsPassedInGTTileEntityUntilLagWarning=100
-    B:MixedOreOnlyYieldsTwoThirdsOfPureOre=true
+    B:MixedOreOnlyYieldsTwoThirdsOfPureOre=false
     B:NerfCombs=true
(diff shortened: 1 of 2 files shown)


Dream-Master:
I don't merge this. Free face config is ok

MineAnPlay:
Why though? It makes it so that the EBF is a LOT more efficient and makes the game less repetitive. Like really you now need to get a lot more ores and they only give nuggets? This just makes the game a LOT more repetitive and boring. With this on false people can still get ingots from the old ways (though impure no longer gives ingots only the normal dust form gives it) and the EBF will be a new ore processing machine which would make people make separate EBFs for ore processing and would make people delve in the new recipes and also have fun and get more out of there ores.

SirAzazel:
I like overall thing with needing of free space, but 2/3 should be reverted to 1

bigbass1997:
@Dream-Master Can you give an actual reason you say no to https://github.com/GTNewHorizons/NewHorizons/pull/1958/commits/de34d0b6ee1010597a815fdaa44a751558809404 ?

PR #2181, opened 2017-11-08
Infernal Mobs edits

damug:
Remove choking, ninja, and vengeance effects from infernal mobs.

https://github.com/GTNewHorizons/NewHorizons/issues/2176

### DIFF
--- a/config/InfernalMobs.cfg
+++ b/config/InfernalMobs.cfg
@@ -632,7 +632,7 @@ general {
     B:"MM_Berserk enabled"=true
     B:"MM_Blastoff enabled"=true
     B:"MM_Bulwark enabled"=true
-    B:"MM_Choke enabled"=true
+    B:"MM_Choke enabled"=false
     B:"MM_Cloaking enabled"=true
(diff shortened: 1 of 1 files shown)


Dream-Master:
If we disable all mobs type we can remove to mod. I like to see fixes in code.

richardhendricks:
@Dream-Master We don't currently have a repo for Infernal Mobs under GTNewHorizons. Can you create one?

Dream-Master:
I will do asap

Dream-Master:
@richardhendricks done

draknyte1:
Fixing the Ninja modifier should be easy..
Vanilla has methods to check for a valid mob spawn area, which in this instance should be used to check for valid teleport areas. Look into things like the Zombie aid event for how the extra zombies are spawned in near the original and the logic is about the same.

Dream-Master:
Ok sounds cool. Hope we get a fix soon 😁

Namikon:
Closing this as the effects will not be disabled, but fixed in code to function properly

PR #2228, opened 2017-11-28
EMTools minetweaker script fix

RaggedCaster:
fully fixed all invalid damage values on ic2 charged items

### DIFF
--- a/scripts/Emt.zs
+++ b/scripts/Emt.zs
@@ -6,6 +6,8 @@
 
 
 import mods.gregtech.AlloySmelter;
+
+
 import mods.gregtech.Assembler;
 import mods.gregtech.FluidSolidifier;
 import mods.gregtech.FormingPress;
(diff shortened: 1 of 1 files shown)


Dream-Master:
You test it ? We use id 27 allready long time before.

RaggedCaster:
Yes. Most of the problems were caused by non-wildcard data value on infusion ingredients (like 1).
So the recipe asked for 0EU item that has an almost full durability bar = unobtainable and broken.
My tweak uses * (a wildcard) in place of all chargeable ingredient metadata (now you can use items with any charge in infusion). I also corrected the metadata of resulting chargeable items so the recipe outputs correct 0 EU items (empty durability bar).

RaggedCaster:
Yes, i tested it. BTW, TC infusion strips NBT so the issue came from using exact metadata in recipes. There was one recipe which used both a wildcard and an exact value metadata lapotron crystal. I recommend using wildcard metadata on all chargeable IC2 items.

Dream-Master:
ok fine. And Thanks @RaggedCaster

Dream-Master:
It not worked. Why the heck you not testing it @RaggedCaster 
image
image
Thaumcraft recipe not shown anymore so i can start from 0 find the error.

Dream-Master:
i try to fix it hope it works now.

Dream-Master:
@RaggedCaster please test with newest scripts and report to me.

PR #2371, opened 2018-01-07
Allow Recycling the Gregtech Sensor Card

codewarrior0:
Like the other cards, gives back half of the circuits used to craft it.


Dream-Master:
Will do it in gt code

PR #2428, opened 2018-01-15
Add recipes for Covered Fluix Cable using synthetic rubbers

codewarrior0:


Dream-Master:
@codewarrior0 thanks

codewarrior0:
I thought about using the EV-tier cable recipes, requiring both synthetic rubbers and a shield layer of either fluix dust or polyphenylene sulfide, but I figure fiber optic cables don't need an EM shield :stuck_out_tongue:

PR #2496, opened 2018-01-23
Update config for gbeta20

Ugachaga:


Dream-Master:
Yes that is true i forget the config :) @Ugachaga thanks

PR #2501, opened 2018-01-23
New Quests

Dream-Master:
by @benstuwie


Dream-Master:
@benstuwie i check it now and edit it a bit if necessary
Thanks for new quests

PR #2619, opened 2018-02-17
Spell checked various old and new quests. Quests 1224 and 1252 still …

Giovanni-NL:
…need to be checked.

### DIFF
--- a/config/betterquesting/DefaultQuests.json
+++ b/config/betterquesting/DefaultQuests.json
@@ -71065,7 +71065,7 @@
     {
       "questID": 773,
       "name": "Way one: Raw carbon mesh",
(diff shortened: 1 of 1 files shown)


Kortako:
The only issue I have with changing Epoxid to Epoxide (which apparently is correct, I didn't know that), is that Gregtech itself uses the term "Epoxid" everywhere.  So if you only change the quest text, new players will probably get confused, because you type Epoxide into NEI and nothing comes up.  I think if you're going to make that change, it needs to happen in Gregtech itself too, but I don't know how involved that is off the top of my head.

Giovanni-NL:
That's a good point.  I just noticed some inconsistencies between quests, and looked up the proper spelling for epoxide.  But if the in-game name happens to be epoxid, then it's probably better to have the quest text match with that.  I'll update that and submit another pull request.

PR #2784, opened 2018-04-05
Replace Pam's salt with Edible Salt

codewarrior0:
Prevents obtaining infinite sodium and chlorine using the Harvestcraft salt recipe. (Infinite Na is still available from centrifuged ashes, and infinite Cl from centrifuged stone dust.)

This patch requires the corresponding patch in the GTNH coremod: https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/25

Credit to @JasonMcRay for the artwork and original implementation.

### DIFF
--- a/config/harvestcraft.cfg
+++ b/config/harvestcraft.cfg
@@ -159,7 +159,7 @@ general {
     B:enableharvestcraftfish=true
     B:enablelistAllwaterfreshwater=true
     B:enablelistAllwatervanillawaterbucket=true
-    B:enablesaltfromwaterbucketrecipe=true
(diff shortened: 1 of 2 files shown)


richardhendricks:
Recommend adding Distillation Tower recipe:

1000L *normal* water (not accept distilled water) -> 250L distilled water, 1 pile of salt.
4s processing time, 256 EU/t

Adjust pile size and water usage as necessary

codewarrior0:
There is already a Distillation Tower recipe for 288L water -> 260L distilled water.

If that recipe were changed to also have a 10% chance to yield a tiny pile of salt, it would reflect the real-world concentration of salt in seawater.

Issue #2786, opened 2018-04-05
Quest 273


CalterOK:
#### Which modpack version are you using?
2.0.3.0
In the quest is necessary Show Queen Blood and iridium screws, but the recipes with blood require osmium screws. Iridium is also used with carmine (they can be added to the quest requirement 274).

PR #2787, opened 2018-04-05
#2786 v2

bartimaeusnek:
since Iridium Screws are only used for certain Staffs and Chrome is smeltable earier than Iridium, i split these quests. enhancement for d971625f6867311bfba01e9c2b148d01cd23a714 #2786


Dream-Master:
Ok good idea

PR #2904, opened 2018-05-06
Turn seeds into collectibles instead of them being infinite

DarkShadow44:
This is a change I've been playing with for a few months now. I don't like the idea of the seeds multiplying when you grow plants, I like to collect them. I guess it fits into a hard pack like this. With this change, every seed you get will only result in one plant - if you want more plants you need to get more seeds. Ofc you get the seeds back when you break the plants.

This is a fairly big change, and I'm not sure if you want it - but I find it useful and have been playing since the start with it. If you want a big farm, you actually need to collect a lot of gardens to find all the plants you want - but I didn't find this to be a problem, I have plenty of food and a big farm.

Changes in detail:
- harvest craft gardens don't spread
- harvestcraft gardens drop seed instead of vegetables
- vegetables can't be turned into seeds anymore
- Only seeds can be planted, vegetables can't
- seeds don't multiply when harvesting a crop, but you can't lose them either
- plants grow slower
- Seedoil is thus a lot harder to get (Can use peanuts for example)

This doesn't affect IC2 crops. Potatos and carrots are also unaffected and can still be planted, because there are no seeds for them.

### DIFF
--- a/config/HungerOverhaul/HungerOverhaul.cfg
+++ b/config/HungerOverhaul/HungerOverhaul.cfg
@@ -29,25 +29,25 @@
 
 delays {
     # Multiplier applied to the delay between breeding entities [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:breedingTimeoutMultiplier=2
+    I:breedingTimeoutMultiplier=4
 
     # Multiplier on the time it takes cactus to grow [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:cactusRegrowthMultiplier=2
+    I:cactusRegrowthMultiplier=4
 
     # Multiplier applied to the delay before children become adults [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:childDurationMultiplier=2
+    I:childDurationMultiplier=4
 
     # Multiplier on the time it takes cocoa to grow [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:cocoaRegrowthMultiplier=2
+    I:cocoaRegrowthMultiplier=4
 
     # Multiplier on the time it takes a non-tree crop to grow [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:cropRegrowthMultiplier=2
+    I:cropRegrowthMultiplier=4
 
     # Multiplier on the time it takes food to dry on Tinkers' Construct drying racks [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:dryingRackTimeMultiplier=2
+    I:dryingRackTimeMultiplier=4
 
     # Multiplier applied to the delay between chicken egg laying [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:eggTimeoutMultiplier=2
+    I:eggTimeoutMultiplier=4
 
     # Multiplier on the time it takes a WeeeFlower crop to grow [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 1.0]
     I:flowerRegrowthMultiplier=1
@@ -56,19 +56,19 @@ delays {
     I:milkedTimeout=20
 
     # Multiplier on the time it takes nether wart to grow [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:netherWartRegrowthMultiplier=2
+    I:netherWartRegrowthMultiplier=4
 
     # Multipier on crop growth time without sunlight (1 to disable feature, 0 to make crops only grow in sunlight) [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 2.0]
-    I:noSunlightRegrowthMultiplier=1
+    I:noSunlightRegrowthMultiplier=2
 
     # Multiplier on the time it takes a sapling to grow into a tree [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:saplingRegrowthMultiplier=2
+    I:saplingRegrowthMultiplier=4
 
     # Multiplier on the time it takes sugarcane to grow [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:sugarcaneRegrowthMultiplier=2
+    I:sugarcaneRegrowthMultiplier=4
 
     # Multiplier on the time it takes a tree crop to grow [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 4.0]
-    I:treeCropRegrowthMultiplier=2
+    I:treeCropRegrowthMultiplier=4
 
     # Multipier on crop growth time (except sugarcane) in the wrong biome (1 to disable feature, 0 to make crops only grow in correct biome) [vanilla: 1.0] [range: 0.0 ~ 3.4028235E38, default: 2.0]
     I:wrongBiomeRegrowthMultiplier=1
@@ -215,7 +215,7 @@ harvestcraft {
 
 harvesting {
     # Multiplier on the effectiveness of bonemeal; the smaller this is, the more often bonemeal will fail to create growth. Set to 0 to disable bonemeal completely. [vanilla: 1.0] [range: 0.0 ~ 1.0, default: 0.5]
-    S:bonemealEffectiveness=1.0
+    S:bonemealEffectiveness=0.5
 
     # Enables/disables harvest crops by right clicking them [vanilla: false] [default: true]
     B:enableRightClickHarvesting=true
--- a/config/HungerOverhaul/HungerOverhaul.cfg
+++ b/config/HungerOverhaul/HungerOverhaul.cfg
@@ -154,7 +154,7 @@ food {
 
 "getting seeds" {
     # Adds a crafting recipe to turn 1 wheat into 1 seed [vanilla: false] [default: true]
-    B:addSeedsCraftingRecipe=true
+    B:addSeedsCraftingRecipe=false
 
     # Each seed has an equal chance to drop (grass drops and via hoes) [vanilla: false] [default: true]
     B:allSeedsEqual=true
@@ -242,10 +242,10 @@ harvesting {
     S:producePerHarvestRightClickMin=2
 
     # Maximum number of seeds you get when harvesting a non-tree crop by breaking it (modifyCropDrops must be tree) [vanilla: 0] [range: 0 ~ 2147483647, default: 0]
-    I:seedsPerHarvestBreakMax=0
+    I:seedsPerHarvestBreakMax=1
 
     # Minimum number of seeds you get when harvesting a non-tree crop by breaking it (modifyCropDrops must be true) [vanilla: 0] [range: 0 ~ 2147483647, default: 0]
-    I:seedsPerHarvestBreakMin=0
+    I:seedsPerHarvestBreakMin=1
 
     # Maximum number of seeds you get when harvesting a non-tree crop with right click (modifyCropDrops must be true) [vanilla: 0] [range: 0 ~ 2147483647, default: 0]
     I:seedsPerHarvestRightClickMax=0
--- a/config/endercore/endercore.cfg
+++ b/config/endercore/endercore.cfg
@@ -26,7 +26,7 @@ enchants {
 general {
     # Disabling this option will prevent any crops added to the config json from being right clickable.
     # [default: true]
-    B:allowCropRC=true
+    B:allowCropRC=false
 
     # Prevent tick speedup (i.e. torcherino) on any TE that uses the base TE class from EnderCore
     # [default: false]
--- a/config/harvestcraft.cfg
+++ b/config/harvestcraft.cfg
@@ -81,7 +81,7 @@ dimensions {
 gardens {
     B:enableberrygardenGeneration=true
     B:enabledesertgardenGeneration=true
-    B:enablegardenSpread=true
+    B:enablegardenSpread=false
     B:enablegourdgardenGeneration=true
     B:enablegrassgardenGeneration=true
     B:enablegroundgardenGeneration=true
@@ -94,7 +94,7 @@ gardens {
     B:enablewatergardenGeneration=true
     I:gardenRarity=2
     I:gardendropAmount=3
-    B:gardensdropSeeds=false
+    B:gardensdropSeeds=true
     I:gardenspreadRate=100
 }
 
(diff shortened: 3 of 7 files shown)


Dream-Master:
I like the idea but will ask @bartimaeusnek too.

Methes:
No. If you want to play that way, sure, use those settings in SP.
I personally find plant grow slow as it is now in early game and food is already a big enough annoyance.

WarlordWossman:
So did you just remove all pams plants into seed recipes because I don't want you to break my glowflower farm. 
Besides that I do agree with @Methes tho, esp when you travel on a server because people build near spawn already it will be impossible to not starve.

bartimaeusnek:
@Dream-Master @WarlordWossman @Methes this would force the use of IC2 Crops at earler stages, if you really want a bigger farm early. IMO its a good idea, it re-balances crops and nerfes normal farms, but its also controversal to remove "vanilla" methodes.

DarkShadow44:
>  esp when you travel on a server because people build near spawn already it will be impossible to not starve.

I don't know exactly how it works with pam's gardens on a server, so dunno. They are always limited. But you can a lot of seed (especially wheat) from using a hoe against dirt.

> So did you just remove all pams plants into seed recipes because I don't want you to break my glowflower farm.

Why does your farm rely on turning plants into seeds? When a crop is planted, you will need seeds. When it's harvested, you get only the plants. When it's destroyed, you get the plants and exactly one seed. You can still make a farm with that, but you won't be able to increase the number of seeds you have.
However, I just noticed parm's harvest the nether doesn't respect the config. It has a bug, so this change doesn't affect nether plants.

> I like the idea the nerf did but what we can do on servers when the pams garden are allready be taken by others?

How did you handle that in the past? Because to get seeds to multiply you first need at least one garden. Maybe we could add some energy expensive recipe to duplicate seeds?

WarlordWossman:
>Why does your farm rely on turning plants into seeds? When a crop is planted, you will need seeds. When it's harvested, you get only the plants. When it's destroyed, you get the plants and exactly one seed. You can still make a farm with that, but you won't be able to increase the number of seeds you have.

So for example the ender IO farming station does right click actions which breaks the plant (giving you multiple things, however for glowflowers you get no seeds but multiple glowflowers which means I need an autocrafting table to turn them back into seeds).
Not sure how the exotic seed upgrade will work but I could imagine it also harvesting with right click, like the steves carts farmer that fully breaks wheat and replants the seeds after.
So this change would make pams crops incomatible with steves carts and ender IO farms, not sure if you can automate glowflowers at all since most farming devices right click rather than left click harvest things.
Golems might work, crops should, but I think that part is my main concern as it removes ways of automation.

@bartimaeusnek I am usually all for crops but in this case it will feel weird when you want to automate without IC2 crops. I always thought the idea was getting a really good plant if you want to crossbreed them, in the end I would probably be okay with forcing crops to automate but many people will be upset and some farming devices I already named will lose a lot of utility.

DarkShadow44:
With my changes the plant shouldn't break on rightclick, or when it does, you get exactly one seed. I don't see a problem here.

WarlordWossman:
Ah sorry, I meant leftclick, the farms leftclick/break the things.
Also try leftclicking a fully grown glowflower.

DarkShadow44:
Added a fix for nether seeds, since that didn't work properly yet.
Also needs a HungerOverhaul fix, please update HungerOverhaul to https://github.com/DarkShadow44/HungerOverhaul/releases/tag/v1.0.3

DarkShadow44:
Okay, there is a problem with the farming station from EnderIO, since that doesn't respect hungeroverhault config in the slightest. 

1. Problem: It can still plant harvestcraft crops, despite the player not being allowed to
2. Problem: It harvests glowflowers as seeds (Only glowflowers)
3. Problem It gets the wrong amount of drops from the plants

How do we want to handle this?
That would require a code change in harvestcraft respective harvestthenether.

Dream-Master:
ender io we forked so far I remember and pams have no source for 1.7.10

DarkShadow44:
I'm working on a fix, but it'll take a while.

codewarrior0:
@DarkShadow44 Do you know how this change affects Pam's cotton versus Natura cotton? Almost all recipes use the Pam's cotton, but that can be crafted into the Natura cotton and then into Natura cotton seeds. 

Also, do you know how this affects Thaumic Bases crops? Probably not important since you can just *craft* the seeds using alchemy, but it would be nice to know.

bartimaeusnek:
@codewarrior0 @DarkShadow44 cant we just change natura cotton aswell?

DarkShadow44:
@codewarrior0 This affects natura cotton as well. I removed the recipe that allows to craft the seeds from natura cotton - I missed one, just added that change.

I don't know about Thaumic Bases, didn't use it yet. Do these crops usually drop seeds? If you can tell me how I can use them in creative, I can test it. I can't plant them on normal farmland, that's all I know.

codewarrior0:
Some of the seeds can be planted on normal farmland, and some of them require a special block either under the farmland or to be planted on directly - I forget which is which. Grab the Creative thaum book and it should tell you.

bartimaeusnek:
@DarkShadow44 @codewarrior0 they drop 1 seed per plant afaik.

richardhendricks:
Redlons for sure multiply.  Most of the others are only 1 seed.

Not sure I like this change, tbh. It encourages people to get the OP food items even quicker, and makes the early part of the game more painful.

And frankly it doesn't help realism either, since you have to keep scavenging for seeds even though you have full grown fruit already.

If we keep this, how about adding extractor recipes to convert fruits/wheat/carrots/cotton/etc into seeds?

Methes:
All this does is it forces people to spend more hours exploring the world at the start of the game when they have nothing. Also no backpack so it's not like they can take all the treasures they find.
You'll be forced to find more lootboxes to hope you get the seed you want out of them. You want 81 cotton field? Better go away for an hour.
World generation is one of the most problematic aspect of this pack - servers don't keep up - so in MP everyone will lag as hell unless they pregenerate tons of starting land. In servers where everything is picked up already (delta, epsilon) well good luck.
It doesn't add any gameplay value, it's just another limitation for no reason at all.
Oh you're all about realism? This goes right against it. That's not how plants work. Would you like for trees to drop only one sapling too?
Don't keep extending the early game constantly. You wanna improve something? Add tier 0 ore finding tool cause spending 10 hours being a living ore scanner is stupid gameplay.
Extend late game, not early game. It's a drag already with new circuits, chemistry and other things that weren't present in the past.

draknyte1:
I agree that as an optional feature, this is nice for the niche of players who enjoy these changes. Overall, it offers no benefits to servers that aren't fresh, nor does it improve the early game grind that people already find tedious.

What you have to remember is that a change like this can drive players away very quickly if they're only just starting, definitely not a positive at all.

Food grind early on is already a massive PITA, so why make it worse than it already is?

codewarrior0:
Food changes are meh, even a casual player should have no problem staying fed with limited seeds. What really hurts about this is the cotton nerf, since you need a lot of it for things like lava gloves, backpacks, and pretty much anything that uses tanned leather.

Krispion:
Then why make the change? If you can survive fine without it and it has no basis in reality, what's the point? I don't see how it would make farming more fun as you're removing a fundamental aspect of the mechanic. It's also not forcing the player to automate like microcrafting nor is it really even possible to automate conventionally. You're trying to make something which is inherently not about resource gathering into a mechanic which requires resource gathering without compromise. Reduce seed spawn chance, limit yield from crops or gate seed crafting behind something but don't just remove the ability for plants to multiply. I would liken this to removing all oregen and forcing players to find all their metals from dungeons. It's harder, it's more tedious but it makes no sense.

Dream-Master:
I hope this here is a place for discussion so we will hear all pros and contras before we make a final decision.

sthegreat:
I also think that this wouldn't be a beneficial change. Changing the seed recipes to require gt tools is reasonable, but if you make seeds totally non-renewable it will only serve to further discourage players from setting up biodiesel compared to oil. It's also fairly unrealistic. And if you want ic2 crops to be more viable, in my opinion it is best to buff them, or to nerf other crops in growth rate or output instead of seed multiplication.

WarlordWossman:
I thought more about it and I think it doesn't really add anything.
If I don't travel around with no base I can easily survive on carrots, carrot juice, potatoes, baked potatoes, potatoes on a stick, natura berries etc.
And there is no real point in making the food items annoying to get. In general I wouldn't say that the pack is just trying to be tedious but instead it tries to be more complex in a way where you have to think about setting up in a convinient way. 
We can all search for seeds for a few hours, there is not much thinking involved. Technically you can even just kill yourself to refill your hunger bar + the quests give you TONS of food anyways.

I would like to see IC2 crops being more usefull but in the end this concept with the pams seeds doesn't add anything besides stretching the early game. There is already a lot of exploration to it as of now with quests like the one for the healing axe etc.

Some additional things to think about:
* I can afk at my base without losing any hunger which means the number of crops is not important since I generate food without losing food.
* There are enough (renewable) food items in the game so that players who are aware can completely dodge pam's food while newer players might struggle even more with it.
* It's not realistic that you don't get additional seeds.

As others said, we can tweak the rates a bit or add a GT knife to get seeds from the plantable food items but at the end of the day this change doesn't really add anything.
We decided against stuff like a thirst bar (second hunger bar with dif items) and hardened stone (takes 4 times as long to mine anything) and this change belongs into said catergory.

DarkShadow44:
Well, I see. My main motivation for that suggestion was that it's pretty easy to have everything from harvestcraft in masses even in tier 1. There's just no progression, no need to IC2 crops up or similar.
Do you have a suggestion for a change that doesn't remove seed multiplication like I proposed, but also combats this issue?

WarlordWossman:
I would say that spice of life already encourages you to not only plant a single food item but multiple ones for which you have to get the gardens. With the limitation of inventory space and the lunchbag/box having a set number of slots we shouldn't really touch those values I feel like.

In LV you already get the tin cans which negate the food variety and for steam age you can get a snow golem to get access to all the smoothie types without leaving your base to gather the snow balls.

Ideally the better food items (and you need to explore way more to get all the ingredients) should be more benefitial so that the focus is more on variety than multiplying the same crop. Realistically that's really hard to do without forcing the player to make them, maybe we could remove the "well fed" effect from all items that only give 1 hunger point or less but in the end the bulk of NEI pages and all the microcrafting just make the better foods annoying to deal with esp in a pack that is focused on progression.
Maybe you would bother with it if the better foods would be items you can eat more often before they get the diminishing returns instead of just filling up a massive chunk of your hunger bar.

So yeah the food aspect as of now isn't really a main focus of the pack, it's just a bit harder than eating baked potatoes for your whole lifetime.
Using your seed concepts fot NON food items like glowflowers and cotten might be worth thinking about but redstone/gold and string are renewable anyways throught spawners (which is more work than seed multiplying so that might make crops more attractive)

richardhendricks:
I dunno.  The only way to really tier the foods is to restrict the recipes for the Pams food crafting items.  Make a rolling pin require an LV lathe.  A mixing bowl require an MV (mumble-compressor maybe).  And the baking pan an HV bending machine.

Then move tin cans to EV and make them only 80% successful a canner.  

Add stainless steel cans at IV that are 100% successful with a canner.

There you go, progression.

Of course, players would hate every minute of it.

Zoko061602:
I think it wouldn't even make sense to nerf glowflowers and cotton, since you can get glowstone and string from thaumic bases crops (LV-MV), which shouldn't be affected by that config and doesn't even require GT processing, like glowflowers.

richardhendricks:
FYI, woven cloth requires cotton and string, lots of both if you are using them for bee frames.  Thaumic bases plants only give you string. You'll still need an extensive cotton farm.

Krispion:
I wouldn't mind moving up food cans to a higher tier of progression as richardhendricks suggested, those things are honestly OP. I can throw raw potatoes into a canning machine and I'll never have to look at another food crop again. Aside from the other gating suggestions for the pots and pans, the ideal would be a proper heating mechanic for Pam's foods like pizzas and whatnot but I'm not sure how that could be implemented within the scope of the pack. Another option would be nerfing food cans to only accept cooked foods and add better scaling to these (like multiple cooked potatoes for one can).

richardhendricks:
*Visions of a GT++ multiblock food processor...*

I like the idea of requiring cooked or higher food tiers vs low yield foods for tin cans.  Maybe a simple mechanic like "Do not can items worth less than 3 shanks" could work there.

LOL @ proper heating mechanic.  So now we will have to cook dough in a multismelter with a circuit to get the proper food output, and it will be hot and cause damage without BBQ gloves.

WarlordWossman:
>I dunno. The only way to really tier the foods is to restrict the recipes for the Pams food crafting items. Make a rolling pin require an LV lathe. A mixing bowl require an MV (mumble-compressor maybe). And the baking pan an HV bending machine.
Then move tin cans to EV and make them only 80% successful a canner.
Add stainless steel cans at IV that are 100% successful with a canner.

I think food cans from stainless steel would be okay but IV is literally insane :D
Maybe aluminium ones that have a breaking chance so it's not so far away but you need to deal with can loss if you want to go with less food variety.
For the more advanced foods I mentioned that nobody really bothers with crafting them if they are not epic beacon and you got the dyes from a lootbag. Gating rolling pin and mixing bowl will only further discourage ever making them while having the side effect of gating the healing axe if you don't manage to loot all the food you have to submit.

> I think it wouldn't even make sense to nerf glowflowers and cotton, since you can get glowstone and string from thaumic bases crops (LV-MV), which shouldn't be affected by that config and doesn't even require GT processing, like glowflowers.

We probably agree that some TC research is more work than just planting some seeds at your base.

>FYI, woven cloth requires cotton and string, lots of both if you are using them for bee frames. Thaumic bases plants only give you string. You'll still need an extensive cotton farm.

This is a good point but you CAN get natura cotton seeds by hitting grass blocks (that are not near water) with a hoe which makes them renewable so we could leave it like that rather than the simple way which is turning the cotton in to seeds.

Krispion:
>So now we will have to cook dough in a multismelter with a circuit to get the proper food output, and it will be hot and cause damage without BBQ gloves.

Oh no not like that, I just mean basic "smelt" processing of foods. It's not important anyway, was just a little weird to put tomatoes cheese and dough on a cold pan and have a pizza pop out. As I said it's a bit out of the scope of the pack as it would probably require adding intermediary items and it just isn't big enough a deal to bother.

codewarrior0:
Personally, if I were to start a new game with this change, the *only* way it would affect me is that I'd have to spend another hour or two hunting for textile gardens and Natura worldgen cotton to get a cotton farm going. Between food cans, the healing axe, ample amounts of quest/lootbag food, and suicide, this change really has no impact on my food situation.

I don't think Biodiesel lovers would notice either, since they'd rather breed Forestry trees than mess around with crops.

It also makes Seed Soup basically uncraftable.

WarlordWossman:
Yeah for sure, I think we all agree that the limited seed thing might sound nice on paper at first but will probably just stretch the game etc.
As I said a way to buff the better food types and moving tincans a bit further in the progression might already make the food progression better.

PR #2965, opened 2018-05-18
Downgrading Circuits.

bashduude:
Added "downgrade" for circuits.
Reason being - carpenter recipes require "old technology" circuits and its really annoying to go backwords and craft old unconvinient circuits.

At first i decided to add oredict to carpenter recipe, but it would not let me.

So i though: "Well this game already has Lv-tier circuits conversion,
why wont add another Circuits same possibility?"

This means you could craft any Tier circuit at any recipe you sit on right now, and can
always convert them "backwards" (and only backwards) to use in recipes such as
circuits for multifarms and engenes from forestry inside carpenter machine.

image

image

image

### DIFF
--- a/scripts/Gregtech.zs
+++ b/scripts/Gregtech.zs
@@ -1321,6 +1321,28 @@ mods.tconstruct.Casting.addTableRecipe(<gregtech:gt.metaitem.01:32375>, <liquid:
 
 
 
+// --- Circuit conversion
+// --- MV Tier
+
+
+// --- Yellow to Gold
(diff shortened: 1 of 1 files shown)


bashduude:
Yeah i thought of that, but it would be more "messy".

bartimaeusnek:
its just a small for-function in the core mod...

bashduude:
Oh, in the mod itself *derp*

Dream-Master:
yea we add this better to the Core mod.

Dream-Master:
Allready added via code. @bashduude

PR #2998, opened 2018-05-25
Fixing the Thick Neutron Collector (GTTweaker 1.7.0)

Ethryan:
Quick fix for the Thick Neutron Collector for GTTweaker 1.7.0 

Since the Neutron Reflector item doesn't stack


Ethryan:
All other script recipes seem to work without changes when i updated GTTweaker to latest version.

PR #3029, opened 2018-06-01
BQ Localisation/Rework/Key holders

bashduude:
I have started localising/(swapping hardcoded english into .lang files) Quest Book.
The problem is which keyholders to use?
I personally don't really care but one who create quests may find it difficult to manage later.
So we have to aggree on how we are gonna "call" them.
I guess person who add's and creates them can still do it via ingame BQ editor.
But he'll have to message me every time he adds/changes quest. So i turned it into proper .lang file.
Reason being - you can't make .lang quests in ingame editor.

for now it's
"name": "tier0.quest3.name",
"description": "tier0.quest3.desc",

first dot - name of the tab
second dot - number of quest (i number them from left to right, from top to bottom).
last dot - name/description of the quest.

User dont have to do anything to see proper localisation. It's all ready and set.
So only thing he has to do is change the language of the game to beloved one.
image

first image shows numeration
image

Second Img show's how quest look's for user
image

third img show's how quest look's for editor
image

and this img shows how quest looks in .lang file

### DIFF
--- a/config/betterquesting/DefaultQuests.json
+++ b/config/betterquesting/DefaultQuests.json
@@ -6,8 +6,8 @@
   "questDatabase": [
     {
       "questID": 0,
-      "name": "Your first night",
-      "description": "Your first task will be to find food, and a shelter for the night. The nights will be cruel, as this pack has hardcore darkness. It probably would be a good idea to collect some food now...\n\nNote: You don\u0027t need to keep this book, you can use tilde to open quests without the book.",
+      "name": "tier0.quest1.name",
+      "description": "tier0.quest1.desc",
       "isMain": true,
       "isSilent": false,
       "lockedProgress": false,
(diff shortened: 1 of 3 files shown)


richardhendricks:
Will doing that work well with coloring and other inline character stuff?  I am fine with it, it would make doing quests easier to handle it outside of the JSON file.

What about '?  I know the JSON file had lots of problems with leaving ' in the JSON, we have to use \u0027 hack.

richardhendricks:
Baashdude, I just checked in a bunch of changes.  Also, two quests had to be moved from EV to IV line, the ender quarry related quests.

Dream-Master:
@bashduude @richardhendricks as i said earlier the chinese make a script for using lan files. We better using this and write the stuff inside there
https://github.com/Yesterday17/GTNH-GuideBook-i18n

richardhendricks:
Yes ok.  I think after this release, we can lockdown the questbook to extract the EN_US language info and allow for translations.

PR #3259, opened 2018-07-05
GraviSuite HUD

Ugachaga:
Moves the GraviSuite HUD from the overloaded upper left corner to the bottom left corner (below the chat).
gravihud


Dream-Master:
@Ugachaga thanks

PR #3284, opened 2018-07-09
Some fixes and additions.

MineAnPlay:
Fixed a couple of errors and added the max values if one was set. (Bee Frame Tool Tips)


Dream-Master:
@MineAnPlay  thanks

Issue #3277, opened 2018-07-08
Early game storage (pre-AE)


Bryfer:
#### Which modpack version are you using?
2.0.4.6
#
#### If in multiplayer; On which server does this happen? 
Epsilon
#
#### Introduction
I have done a bit of research of peoples knowledge of Projectred Transportation over the weekend.
Most player does not even know it can be used as a storage in MV (almost LV tier)

I have also gone thru the quests for Projectred "Ghetto ME System" and these quest requires some rework, first of all this is the intro quest; Sounds really intressting to do these quests right?

Quest: Project red beginning # 127 (it should be
![image](https://user-images.githubusercontent.com/22825059/42423240-83b53ff0-82f7-11e8-981f-edca7d55024d.png)
This text should be in # 738 instead?
This text should be replaced by something like this

> Welcome to Projectred!
This quest will learn you the basic of Projectred and learn you how to set up a Ghetto AE system.
The Ghetto AE system will be really useful for sorting all of your first items

This quest also wants you to craft;
**16x** _Item Transport Pipes_
**16x** _Null-Logic Routing Chip_ (these can be a real pain to craft early game)

Should be changed to (Read more to know why)
12x Item Transport Pipes
2x Null-Logic Routing Chips

At this point you have no idea these quests will lead to a Storage system.

Quest: Can u eat those? # 737
This quest has a good explanation of all the different chips and what they do and are used for.
However crafting TWO of all of these is completely unnessesary!
**Change this quest to just a check box!!**

Quest: Routed Pipes # 738
Explains the different pipes and their functions
This quest wants you to craft 4 of each of the more advanced pipes (from the 16 mentioned earlier)
**change this quest to just a check box!**

Quest: Ghetto ME System # 739
Here is the first proper mention that this is actually a questline for storage, and also the last quest.
It explains how you setup a basic storage system using Projectred. And it will also give you a really handy Drawer Controller to use with it as reward 👍 
Here is a basic setup that I made that works, and uses a fraction of the items the quests wants you to craft..
![image](https://user-images.githubusercontent.com/22825059/42423362-3fd42768-82f9-11e8-8742-4fff0ab730eb.png)
![image](https://user-images.githubusercontent.com/22825059/42423377-66466a1e-82f9-11e8-86dd-0c2ef8b7ad1a.png)
![image](https://user-images.githubusercontent.com/22825059/42423379-6fc9ccde-82f9-11e8-9239-173a85cb63ac.png)

[...]

PR #3391, opened 2018-08-02
ProjectRed quests

Lainiel:
I've made some ProjectRed quests. They explaid what each chip does and show how to make a basic drawer controller storage system and how to automate desulfuring oil.


Lainiel:
Do not merge it yet. I have to make one more change.

Bryfer:
Whats the difference between these quests and the current quests in MV tier?
#3277 <- this one shows the most recent changes made to the quests

PR #3778, opened 2018-10-24
Fix for Issue 2967

shawnbyday:
I ended up fixing all of the ardite recipes. All of the recipes worked with the Tinkers Construct version of the Ardite Ingot but not the GregTech Ardite Ingot. I simply swapped this in the Tinkers-Construct.zs script. If you'd prefer these commented out and added to the GregTech.zs Extruder section let me know and I can do that instead. If you have any questions please let me know either on Issue 2967 or in Discord.

### DIFF
--- a/scripts/Gregtech.zs
+++ b/scripts/Gregtech.zs
@@ -1015,6 +1015,86 @@ Extractor.addRecipe(<gregtech:gt.metaitem.01:2896> * 2, <TConstruct:strangeFood>
 
 // --- Extruder Recipes ---
 
+// --- Ardite Tool Rod
+Extruder.addRecipe(<TConstruct:toolRod:11>, <gregtech:gt.metaitem.01:11382>, <TConstruct:metalPattern:1> * 0, 606, 120);
+
+// --- Ardite Pickaxe Head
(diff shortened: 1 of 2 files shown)


Dream-Master:
Are Ardite unificate to gt ardite?

Dream-Master:
yes tcon ardit ingots wont change because they not unificated. (there was some reasons to switch it of)
Are you sure we have to change the recipes? Or its better to add a exchange recipe for ingots etc ? @shawnbyday

shawnbyday:
Originally I submitted the ticket because the GT Extruder wouldn’t accept the GT ardite and there was no way to make the tcon ardite. The extruder accepts GT cobalt.

Alternatively I can submit a pull request that just adds the GT Ardite to the extruder but within the GregTech.zs and then leave the tconstruct.zs unchanged.

I actually started that way but then changed it to the tconstruct.zs since that’s where all the rest of the changes were.

I can rework into gregtech.zs extruder section if that is the preference, that’s no problem at all.

Dream-Master:
@shawnbyday sure do that and put all scripts here in this pull. After this i will merge it.

PR #3794, opened 2018-10-26
3780 fix quest 938 electric tree tap

shawnbyday:
Originally updated 'Damage' field from 26 to 27 but the quest still wouldn't turn in. This commit I removed the 'Damage' field all together which allows the electric treetap to be detected regardless of it's damage data. I think it works a little differently than most damage data in the sense that durability and charge seem to be interconnected somehow. If you have any questions you know how to find me lol.


Dream-Master:
@shawnbyday did you test it and it works ?

shawnbyday:
Yes tested on a few different durability. One full, one empty and one at like halfish.

shawnbyday:
Is there a way you prefer in a pull request to have it stated if/how something was tested? I mean I can add it but if there is already a format the GTNH peeps use I can follow.

Dream-Master:
idk all features in git so can't say for sure if there is a way do so. With tested I mean did the quest work if remove damge?

Dream-Master:
@shawnbyday thanks

PR #3804, opened 2018-10-29
Issue 3740 Ore Finding Rod Hints

shawnbyday:
Added detail to the ore finding wand quest to help players understand that the wand doesn't need to be held to work as well as some of the signs that it is working. Sounds and visuals.


Dream-Master:
@shawnbyday thanks

Issue #3808, opened 2018-11-01
Quest ID 1156 Not Completing (iron tiped bolts)


shawnbyday:
#### Which modpack version are you using?
2.0.5.1
#
#### If in multiplayer; On which server does this happen? 
Private server.
#
#### What did you try to do, and what did you expect to happen?
Made some iron tipped bolts. Expected quest to complete. This happened to me a while back but I'm opening issue because someone else was saying same thing tonight. When I first started playing GTNH it was broken, then it was fixed, now it is broken again.
#
#### What happend instead? (Attach screenshots if needed)
Quest didn't complete.
#
#### What do you suggest instead/what changes do you propose?
Have the quest detect the bolts? :D I'm actually going to look at it myself and do some testing on it with some different ideas and try to fix it if those ideas don't work. Worst case if I find out why I can update the quest description. Can assign to me if you want.

Dream-Master:
will check

Dream-Master:
@shawnbyday did you find out why it wont complete ?

shawnbyday:
@Dream-Master I haven't yet. I just got home about 8 minutes ago and I'm waiting for the game to load so I can do some testing :D I'll post back in a bit.

shawnbyday:
@Dream-Master so I setup the infrastructure to test this and it detected everything except the last part where it detects the iron bolts.

After that I opened the quest book with edit mode enabled and used the quest building interface to select the iron bolts I just made then got out of that and turned edit mode off and it detected the bolts.

After all of that I compared the current DefaultQuest.json file to the one I just did /bq_admin default save on and my merge tool reported that they were identical. So even though the files were the same before and after something I did by editing the file made it detect. It doesn't make sense.

I'm currently loading the game to try again in a fresh world to see if I can recreate.

PR #3813, opened 2018-11-01
Issue 3808 - Quest ID 1156 Iron Tipped Bolts Not Completing.

shawnbyday:
#3808 Fixed the quest so it will now complete correctly and I updated the quest text to reflect making the bolts with the smeltery instead of the fluid solidifier.

I've tested this in my locale (US) but I'm wondering if someone else can test in a different locale? I'm not sure it matters but quest detection was sort of odd.

What ended up fixing this was to set the quest to ignore NBT. Based on the diff below it looks like I changed other things when I didn't and I'm wondering if that is a localization thing (hence my above comment).

If you have any questions let me know.


shawnbyday:
Yea let me look at this again when I get a chance either later tonight after work or tomorrow sometime. I usually make my changes to the DefaultQuest.json in a text editor as opposed to the BQ admin interface. The question marks probably showed up because I did it from in game.

Based on the NBT format https://minecraft.gamepedia.com/NBT_format I didn't see anything that would cause an issue but I'll definitely defer to your hesitation on that. Has ignoring NBT for quests caused issues in the past? I'll maybe grab an NBT editor and look at my file to see exactly what the difference is that causes the quest to not complete just to be a little more thorough.

Dream-Master:
you can look into the nbt tag in edit mode and check it.

shawnbyday:
I'm still working on this. I spent most of my weekend at work. I did review the NBT data on the quest but the only thing that was different was my change saying to ignore. I was rushed though so i want to look again and I should be able to get that done tonight after work.

shawnbyday:
@Dream-Master the only way I'm able to get this to complete is by setting it to ignore NBT data. I even tried tossing the bolts on the ground and picking them back up.

You can maybe close my pull request if you'd like maybe someone else has some other things they can try on this. I believe I have it as an open issue as well.

Dream-Master:
yea i fix it now

PR #3951, opened 2019-01-04
Fix can't craft item what using Bedrockium plate

yukieiji:
Because don't use the ore dictionary,can't make Bedrock drum etc..


Dream-Master:
@yukieiji thanks

PR #4088, opened 2019-02-11
fix some bug and update translation

Kiwi233:


Dream-Master:
@Kiwi233  can you add the changes to the json file too. If i change it only in the lang file i guess it will be overriden next time i update the file. Thanks

PR #4197, opened 2019-03-05
Gregtech Stone Chisel Compatibility

damug:
https://github.com/GTNewHorizons/NewHorizons/issues/4188

Left Abyssal Stone in the Black Granite group instead of moving it to the Basalt Group by popular demand.

Sorry about the two pull requests, apparently I'm bad at github?

### DIFF
--- a/scripts/Chisel.zs
+++ b/scripts/Chisel.zs
@@ -269,8 +269,6 @@ mods.chisel.Groups.addVariation("glasswork", <chisel:glass2>);
 mods.chisel.Groups.addVariation("glasswork", <TConstruct:GlassBlock>);
 // -
 mods.chisel.Groups.addVariation("glasswork", <EnderIO:blockFusedQuartz:1>);
-// -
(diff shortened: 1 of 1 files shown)


Dream-Master:
why you remove tcon glass block? @damug

damug:
Werd wasn't messing with that section, will fix tonight though. Apologies.

Dream-Master:
np

richardhendricks:
@damug You can do new commits to your branch and they will be automatically added to this PR, you don't need to do another one unless the PR is approved.

richardhendricks:
How common is Abyssal stone in-game?  I can modify the GT stone gen so it can replace it when a GT stone blob would overlap (already doing hardened clay and quarried stone). Wouldn't matter much I supposed but normal oregen could then happen in the replaced blocks.

I could also research treating the abyssal and quarried stone as a replaceable block during oregen (where it would replace quarried stone with ore+marble and abyssal stone with ore+black granite).

Not important enough?

Dream-Master:
i merge it and rework it.

PR #4356, opened 2019-04-13
O like Oxygen Quest Rewrite

damug:
Slightly rewrote quest text to better direct player toward recipe that creates 2000 L Oxygen from 1 Cassiterite Sand, as well as some minor reworking of the rest of the quest text.


richardhendricks:
Can you add "or sugar from sugarcanes or ic2 sugar beets."

damug:
Added per your request.

Issue #4500, opened 2019-05-17
Need separate Genetics tab


richardhendricks:
Binnie's Genetics is used for bees, trees, and flowers.  Need to setup quest tab covering it.

Hyperic88:
And Butterflies too, although i never used them in large numbers , but they are excellent pollinators afaik :)

richardhendricks:
![image](https://user-images.githubusercontent.com/1214811/57984831-47e3c980-7a25-11e9-8f82-6e5923c839f1.png)
![image](https://user-images.githubusercontent.com/1214811/57984835-4adeba00-7a25-11e9-8b28-f34e8acdd051.png)
![image](https://user-images.githubusercontent.com/1214811/57984836-4f0ad780-7a25-11e9-8e01-b8513fac4d5f.png)
![image](https://user-images.githubusercontent.com/1214811/57984838-5336f500-7a25-11e9-9e11-62ed3721f0ff.png)
![image](https://user-images.githubusercontent.com/1214811/57984840-57fba900-7a25-11e9-91c8-3acc7575f01c.png)
![image](https://user-images.githubusercontent.com/1214811/57984842-5cc05d00-7a25-11e9-82d1-0670162f2d94.png)
![image](https://user-images.githubusercontent.com/1214811/57984844-61851100-7a25-11e9-9382-4cecc78a5fce.png)
![image](https://user-images.githubusercontent.com/1214811/57984846-6649c500-7a25-11e9-8083-5ce91ed5365b.png)

richardhendricks:
![image](https://user-images.githubusercontent.com/1214811/57984864-97c29080-7a25-11e9-82c8-291f70f40cd9.png)
![image](https://user-images.githubusercontent.com/1214811/57984866-9b561780-7a25-11e9-8e1d-d88a09dbc6d9.png)
![image](https://user-images.githubusercontent.com/1214811/57984870-a1e48f00-7a25-11e9-867f-dfcfabec3a55.png)
![image](https://user-images.githubusercontent.com/1214811/57984873-a7da7000-7a25-11e9-8046-0d1b5f9ef8b2.png)
![image](https://user-images.githubusercontent.com/1214811/57984876-ac9f2400-7a25-11e9-96b4-74d4ca696c79.png)
![image](https://user-images.githubusercontent.com/1214811/57984878-b163d800-7a25-11e9-83f7-992a2ad27165.png)
[...]

Issue #4373, opened 2019-04-19
AE2 quest issues


richardhendricks:
Meteorite quest needs to be main quest

Also the quests suck?

Nori SilverrageToday at 5:08 PM
Yes, but I'm not seeing a controller quest
kekzdealer🐝Today at 5:08 PM
AE tab has super weird quests

kekzdealer🐝Today at 5:08 PM
They ask for P2P tunnels and never ask for drives or controllers

Nori SilverrageToday at 5:08 PM
it also never details how to use the p2p tunnels
as I said, I'm a AE newb and I was hoping the quests would help

![image](https://user-images.githubusercontent.com/1214811/56446116-5fb11c00-62c6-11e9-8b5b-080a2f60193e.png)

NoriSilverrage:
Since the lovely Richard here went above and beyond posting this issue for me... I'll add some color.

I am a complete AE newb. Never used it and had barely heard of it before I started playing GTNH. I just got to a point where I can build the stuff and was hoping the quests would provide lots of help. But alas, while they tell you how to build the items, there is little help in creating a decent network.

For instance, in my reading online, the p2p tunnels are going to be crucial for a ore processing center. Yet the p2p quest here just says to build p2p tunnels.

If I was a expert at this I would offer to help with the quests, but it'll likely be weeks/months before I'm decent enough at it to say anything.

richardhendricks:
@kekzdealer @NoriSilverrage Can we get some more details here?

richardhendricks:
Meteorite quest is a main quest, dunno what exactly and where we need next.

NoriSilverrage:
So I realize 21 days have passed, but I've still only started building AE components. Haven't actually used the system. However I have watched a lot of videos and actually completed many of the quests.

**A few things I noticed** 

- The Find Them All quest isn't actually optional. It really should be option. Maybe one of those checkmark quests?

- The quest that asks you to build P2P tunnels tells you literally nothing about them. Not how to use them or why you "need lots of them". There is no follow up quest and there is no explanation on how to use the basic one and switch it to all the other versions of the tunnel.

- The quests that talk about making the crystals and dusts and fluix crystals give no idea on what amounts you are going to need for a basic setup.

- The cable quest should be split into 3 quests. One quest should be on how to get quartz fiber cables and explain that they do not carry networks, but do carry power and thus are used keep networks separate but power centralized.

[...]

PR #4518, opened 2019-05-19
#4500, adventure backpacks, bee stuff, #4373

richardhendricks:


richardhendricks:
@Dream-Master At this point, I am not going to be doing further quests for 2.0.7.3 unless it is a major blocker.

PR #4620, opened 2019-06-01
fixed damage problems with EMT Recipes

bartimaeusnek:
+now accept any damage value

Signed-off-by: bartimaeusnek <33183715+bartimaeusnek@users.noreply.github.com>


bartimaeusnek:
yes, but i only tested one infusion not all, the one i tested worked perfectly, so i assume it will work fine with every other recipe aswell.

PR #4879, opened 2019-07-28
Quests should check NBT for items that really matter

DeForce:


bartimaeusnek:
@DeForce did you check NBT on ANY GT Block or tool? if so, that will break the quest

DeForce:
It is actually simpler.
No NBT data in task items - we don't check for NBT.
We have NBT data in tasks - we check NBT

DeForce:
and yes, I did manually check couple quests, they did work.

bartimaeusnek:
sometimes there is NBT data on items, where it does not matter, i.e. motors, if we check for the NBT there, only the person who wrote/edited the quest will be able to complete it. same goes for a lot of BW/GT/GT++/TT machines, and some Thaumcraft stuff aswell. But we will see on Zeta how it goes.. just fyi

DeForce:
This is a good point, let me do a recheck, but I think if I find that data - we better remove it from task data cause it's not really valid data point for a quest

PR #4981, opened 2019-09-05
Move AE down to HV

Methes:
Titanium changed to Vibrant Alloy. Molecular Assembler still needs EV Assembler.

### DIFF
--- a/scripts/AE2.zs
+++ b/scripts/AE2.zs
@@ -114,7 +114,7 @@ val DiamondCircuit = <appliedenergistics2:item.ItemMultiMaterial:24>;
 
 val SteelPlate = <ore:plateSteel>;
 val AlPlate = <ore:plateAluminium>;
-val TitaniumPlate = <ore:plateTitanium>;
+val VibrantAlloyPlate = <ore:plateVibrantAlloy>;
 val RubberPlate = <ore:plateAnyRubber>;
 val CertusPlate = <gregtech:gt.metaitem.01:17516>;
 val NQuartzPlate = <gregtech:gt.metaitem.01:17522>;
@@ -508,9 +508,9 @@ recipes.remove(<appliedenergistics2:item.ToolPortableCell>);
 
 // --- ME Controller
 recipes.addShaped(MEController, [
-[TitaniumPlate, AdvCircuit, TitaniumPlate],
+[VibrantAlloyPlate, AdvCircuit, VibrantAlloyPlate],
 [DiamondCircuit, FluixBlock, DiamondCircuit],
-[TitaniumPlate, AdvCircuit, TitaniumPlate]]);
+[VibrantAlloyPlate, AdvCircuit, VibrantAlloyPlate]]);
 
 // --- Energy Cell
 recipes.addShaped(EnergyCell, [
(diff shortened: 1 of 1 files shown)


mitchej123:
Personally I'd like to see just the basics (storage) available at HV.  I'd rather everything else (autocrafting, quantum tunnels, etc) stay at their current tier.

I'd rather throw titanium (which I have a lot of) at those than vibrant alloy (which I have almost none of)

bigbass1997:
I don't know how this would play out exactly, but perhaps think about leaving in the titanium recipes, and making separate recipes that use vibrant alloy (tho perhaps even more expensive than just the alloy). This way once the player gets access to Titanium, it becomes a lot easier to make more AE stuff, and thus there's still a driving force to progress.

mitchej123:
> but perhaps think about leaving in the titanium recipes, and making separate recipes that use vibrant alloy (tho perhaps even more expensive than just the alloy)

I'm a fan; then it doesn't break existing setups

bartimaeusnek:
As i have already stated under the issue, i dont think we should move AE2 down a tier, simply because it encurages the use of storage interfaces on drawers, which hughly impacts performance. If we want to downtier AE2, we should definitely downtier the storage Cells aswell.

mitchej123:
Are we ignoring #votes then?

image

mitchej123:
If we're not going to use the results from #vote... let's remove the channel or make it 100% clear with a sticky/admin post they're "suggestions only" and the staff may choose to do something completely different at their own whim.

botn365:
the only tings that should be down tierd is the energy acepter and the interface tis will alow for a smal ae network in HV

Dream-Master:
i am agianst a complete change too. Small changes on the first items you need for ae is good idea . Make a small ae in hv and upgrade/make it bigger  it in ev.

Dream-Master:
and btw why vibrant alloy ??? why not stainless steel or another material

ScriptedPiky:
> and btw why vibrant alloy ??? why not stainless steel or another material

So Vibrant Alloy was my idea in the #github-discussion because it's a more 'difficult' material in the HV tier. Why not Stainless you ask? Well, Stainless Steel is obtained right as you progress into HV. So theoretically, the player can directly skip towards the beginning of AE with beginning HV level materials.

However, due to Vibrant Alloy now being the case, you now require three important things: Cleanroom, Freezer, and several other requirements before progressing onward into AE.

observeroftime02:
I would honestly prefer Vibrant Alloy, because it's a more complex material to obtain than simply stainless steel. The making of Vibrant Alloy is a bit more involved than Stainless. 

Look at this:
image

mitchej123:
as long as the titanium recipes stay whatever we do...

Dream-Master:
My Plan is to change only the 4 or 5 basic componen items for make a poor man's ae like power acceptor rack chest and cables.

ljjmrlion:
AE storage controller on a drawer controller sounds really nice for HV.

Dream-Master:
If i tiere them down why not put recipes in 9x9 grid of Avaritia table. Make recipes a bit more complex like 4x4 or 3x4 crafting.

Dream-Master:
or assembler recipes

shawnbyday:
This will also require updating the quests so that it has a new "requirement" for unlocking the AE tab.

Dream-Master:
i dont merge this. I close this now. We need a new and better concept for it.

PR #5062, opened 2019-10-24
BQ NBT Detection Issues

observeroftime02:
Hopefully fixes https://github.com/GTNewHorizons/NewHorizons/issues/5047
Changed all the IDs that Piky listed out, and set them to Ignore NBT Data. This should hopefully fix the quest not detecting the items when crafted.


observeroftime02:
No problem, glad to be of help.

PR #5348, opened 2020-01-16
Feature/2.0.8.3 quests

richardhendricks:
Update with all quest changes for 2.0.8.3.


richardhendricks:
Probably because of the sound fix, every requirement (4800+) had to be changed. You can look at the other commits to see what changed for each one.

PR #5364, opened 2020-01-17
climat control cfg  from epsilon

Dream-Master:
Need to check the cfgs @Dream-Master


Dream-Master:
@KaneHart fee free to post your configs. We can do maby more than 2 versions of the config put it in a folder and let people decide which ones they want to use/test

Issue #5354, opened 2020-01-17
Twilight Forest is dark


chochem:
#### Which modpack version are you using?
2.0.8.2
#
#### If in multiplayer; On which server does this happen? 
sp
#
#### What did you try to do, and what did you expect to happen?
expected the usual twilight gloom.
#
#### What happened instead? (Attach screenshots if needed)
It was fully dark (except around light sources).

![2020-01-17_00 10 13](https://user-images.githubusercontent.com/40274384/72573529-ecbcec00-38bd-11ea-9341-eaaa074dca35.png)

The config for the hardcore darkness mod already says:
Whether it should be dark in the Twilight Forest [default: false]
    B:"Dark Twilight Forest"=false

#
#### What do you suggest instead/what changes do you propose?
fix plz unless this is intended. The nether is unchanged and fine by the way.

Edit: Still the case in 2.0.8.3

Dream-Master:
will change back

chochem:
I fixed it by adding the dim (7) to the blacklist in the config. Then there is no hardcore darkness in the TF at all of course.

Dream-Master:
mit needed but an option too

PR #5370, opened 2020-01-18
Fix TF for #5354

Ethryan:


Dream-Master:
why this is necessary ? When tf is set to false @Ethryan

draknyte1:
> why this is necessary ? When tf is set to false @Ethryan
Because if you actually tested it, you'd know the config does not work. 🤔 

Is TF *always* Dim 7 though? That's what you really should check before merging this.

Dream-Master:
ah nice. Was in all dims except the tf. ok will merge it and upload a new file

PR #5489, opened 2020-02-02
Makes all blazes Special Blazes, so that snowball damage works when t…

richardhendricks:
…hey are infernal. Vanilla spawn ratio set to 20 so that special blazes are rarer.


richardhendricks:
All "vanilla" blazes will be infernal, per before.

PR #5527, opened 2020-02-06
Added quests, changed some quest text

Prometheus0000:
I hope i did it right?


Dream-Master:
@Prometheus0000 thanks for new quests.
I will merge it now but maybe change things a bit like add coin rewards or so.

PR #5529, opened 2020-02-06
feat(ordict): add all wooden fences to fenceWood

leagris:
Forestry mod already provides the fenceWood oredictionary name.

- Adds all other wooden fences to `<ore:fenceWood>`
- Use the oredict in recipes to reduce the number of same output recipes

Recipes updated to the `<ore:fenceWood>`:

- Carpenter's Barrier (Carpenter's Blocks)
- Builder's Backpack (Forestry)
- Piston (Minecraft)
- Bed (Minecraft)
- Firework Display (Steve's Carts 2)
- Stencil Table (Tinkers' Construct)


Dream-Master:
@leagris the idea is good but it would better to add fence oredicting to the core mod to load before scripts. I remeber we have some wired  issues in the past.

leagris:
> @leagris the idea is good but it would better to add fence oredicting to the core mod to load before scripts. I remeber we have some wired issues in the past.

Now offloaded oredict registration to the core mod.

*could not compile a working version of the core mod with current build.gradle/*
*See: https://gist.github.com/leagris/a6cd767d29de9ef8758a51053c35a773*

PR #5598, opened 2020-02-16
Fixed interDimensionMultiplier for SG...

KaneHart:
4x is default 10x you can't go across dims pretty much our one gate from overworld to the end cost 214,000 Power and you can only fit 4x50k under...

Not only that but this is so beyond end game 1 or 1000 power units under it make 0 difference....

So this at least lets you use it correctly...

### DIFF
--- a/config/SGCraft.cfg
+++ b/config/SGCraft.cfg
@@ -52,7 +52,7 @@ stargate {
     I:explosionRadius=10
     B:explosionSmoke=true
     I:gateOpeningsPerFuelItem=24
-    D:interDimensionMultiplier=10.0
+    D:interDimensionMultiplier=4.0
     D:maxEnergyBuffer=1000.0
     I:minutesOpenPerFuelItem=80
     B:oneWayTravel=true


Dream-Master:
@bartimaeusnek you set the vaues for SG?

Dream-Master:
@KaneHart its GTNH so nothing is cheap :)

mitchej123:
Impossible != Hard.  :)

mitchej123:
The other (probably better) option would be to edit the power modules so they allow more than 50k in at a time.

Dream-Master:
SG is open source so we can easy edit it.

KaneHart:
Yeah SG Open Source. Honestly if we go that route we should do it right. The Power Units should hold 100's times more energy but it also should be that they use way more energy...

Match Stargates Power Usage to what they require for the tier you unlock them at makes sense.

I'm fine with making them tons harder. My config change is to make them correctly now, but I be very happy to see it done right.

GTNH-Colen:
Why on earth fixes like this get stalled but heavily opposed nerfs can get pushed through in what feels like a day will never cease to amaze me.

Dream-Master:
because I think we have to change source make the energy buffer bigger instead lower the cost of energy.

Dream-Master:
@ElectroBot 
we have forked SG craft. Maybe you said me what values we can change in code.
I close this now

Issue #5574, opened 2020-02-14
Suggestion: Greatly reduce the amount of warp gained from Blood Magic research


Prometheus0000:
#### Which modpack version are you using?
2.0.8.3
#
Mentioned in #4843, the amount of warp in these researches is insane. Given that BM doesn't even naturally have such a mechanic, plus the fact that it's just too much (3 times the amount to get withers!) I suggest reducing it quite a lot. I know it (maybe) fits the theme somewhat, but it greatly discourages people from using it. Given that it's available from MV, but that people avoid it until they have pure tears (LuV), that seems to go against the tiering of the mod. If it's necessary to have lots of warp, maybe it could be concentrated on specific items? Like the ones to summon the demon invasion for example.
#
#### What do you suggest instead/what changes do you propose?
Reduce the amount of warp on these researches.

boubou19:
closing this as it's in the pack since a year now

Prometheus0000:
You mean rejected since a year ago.

Issue #5404, opened 2020-01-20
A very very cheap way to remove all permanent warp 2


mrTeroblaze:
GTNH 2.0.8.3
Warped Tumor is a Self-Infusion added by the Thaumic Horizons mod. It removes 50 of the player's Warp untill the player's death.  (c) FTB wiki.
So this handy infusion removes all warp from the player and death wont returning it back.
Tested both in MP and SP.

Dream-Master:
Ok nerf incomming :)

mrTeroblaze:
That is a bug, Nerfing this is unreasonable becouse there is not so many ways to get rid of warp untill IV. So tumor (and Soul Brazier that have same glitch) are really helpfull.

Dream-Master:
change the recipe will make it harder. Forbidden is forbidden for good reasons

ScriptedPiky:
The warp technically doesn't get removed upon death, however, if you close your MC client and load up a new one, all the warp will magically return back to you upon the next loading session. I had this occur on Zeta. No idea why returning on death doesn't properly return it but restarting & making a new client does.

mrTeroblaze:
> 
> 
> The warp technically doesn't get returned upon death, however, if you close your MC client and load up a new one, all the warp will magically return back to you upon the next loading session. I had this occur on Zeta. No idea why returning on death doesn't properly return it but restarting & making a new client does.

Tested it on my server - warp didnt come back even after client restart

ScriptedPiky:
> Tested it on my server - warp didnt come back even after client restart

Really? Well, that's incredibly strange. Does it return on server restart as well? Adjusting TC warp is a wacky as hell process then. So many bugs.

mrTeroblaze:
> 
> 
> > Tested it on my server - warp didnt come back even after client restart
> 
> Really? Well, that's incredibly strange. Does it return on server restart as well? Adjusting TC warp is a wacky as hell process then. So many bugs.

Yep, tryed all variants. 
So, as i mentioned above, it`s now useless to do any kind of magic research deeper than novice quests untill IV, becouse 2 methods of dealing with warp bugged and constantly using salts/spa is not the option.
Yes, yes, you can say: do only usefull researches, manage your way etc, But already at 30 warp you gain fancy lighting effect which is pain in the ass/

bombcar:
Ok this is *mostly* fixed but the "tumor removes ALL permanent warp" part still exists.

But it does return 50 upon death at least now.

[...]

Issue #5248, opened 2019-12-29
A very very cheap way to remove all permanent warp


nanmumc:
#### Which modpack version are you using?
2.0.7.5

#Only for servers

1. Place a Soul Brazier and make it works

2. Load the chunk and Logout

3. Make sure that the Soul Brazier can **stop working when you are offline** (eg. provides small amout of Mortuus essences)

4. Login and the permanent warp will disappear

richardhendricks:
Thaumic Exploration has a CC license, so we can change behavior so that Soul Brazier dumps all the warp back into you if it is not loaded and you connect.

calombardi:
I experienced this bug last night running version  2.6.0 on java 21 on a private mutiplayer server I run on my local machine. Please let me know any additional info neccesary.

Dream-Master:
> I experienced this bug last night running version 2.6.0 on java 21 on a private mutiplayer server I run on my local machine. Please let me know any additional info neccesary. @Dream-Master here

can you test with 2.7.0 beta 1 ?

calombardi:
yes I can tonight. I'll return with results.

calombardi:
@Dream-Master my client crashes attempting to run 2.7.0 1 beta. MultiMC Java 21

Dream-Master:
@calombardi this one should work <https://github.com/GTNewHorizons/DreamAssemblerXXL/actions/runs/11063768271>
with java 21. fresh install no copy over

calombardi:
@Dream-Master I'm doing something wrong, when you have time to discuss I'd like to help but I'm missing something and this isnt working.

EnderProyects:
Not solved

Issue #4961, opened 2019-08-27
Sigil of Compression can skip Machine Compression


yukieiji:
#### Which modpack version are you using?
2.0.7.5
#
#### If in multiplayer; On which server does this happen? 
Private Sever
#
#### What did you try to do, and what did you expect to happen?
Added Sigil of Compression to inventory
#
#### What happened instead? (Attach screenshots if needed)
Item is compressed such as Cobblestone, Gold, Iron.
#
#### What do you suggest instead/what changes do you propose?
Fix it.
This issue makes it easy to make BedRockium without TNT.

Methes:
So the Sigil worked as intended and that's a bad thing somehow?

Methes:
Never played with BM. How difficult is it to get the Sigil?
It requires player interaction and can't be automated.

yukieiji:
@richardhendricks This sigil doesn't use Life Essence.

boubou19:
the issue could be if it works with injections through player interfaces. If yes then you can hook up a cobble gen on the player interface and jackpot you have easy bedrockium.

Dream-Master:
nerf incomming :P

Dream-Master:
better Infinity

mitchej123:
Please vote:
:+1: - Leave as is
:-1: - Some form of nerf

Personally I'm a fan of magic options/alternatives.

AlexBucket:
Is it possible to modify the amount of LP it takes per compress?

noobyaran9:
> Is it possible to modify the amount of LP it takes per compress?

`I:"Sigil` of Compression"=250

Its possible

Dream-Master:
use more blood is a good idea and make the sigil a bit more expensive too. Not to bad i mean

PR #5604, opened 2020-02-17
#5574 Shuffled warp around. Sigils mostly had it removed. Late game s…

Prometheus0000:
…tuff had it increased, a lot. Added minor amount to a few critical items. Reduced amount to some critical items. Bazooka, Demon Summoning stuff, and Omega armor had a lot added. Hopefully more balance overall.

### DIFF
--- a/scripts/Blood-Magic-Thaumcraft.zs
+++ b/scripts/Blood-Magic-Thaumcraft.zs
@@ -1169,7 +1169,7 @@ mods.thaumcraft.Infusion.addRecipe("ENERGYBAZOOKAIII", <AWWayofTime:energyBazook
 [<Avaritia:big_pearl>, <Avaritia:Resource:6>, <DraconicEvolution:awakenedCore>, <Avaritia:Resource_Block>, <Avaritia:Resource:6>, <Avaritia:Resource:6>, <Avaritia:Resource_Block>, <DraconicEvolution:chaoticCore>, <Avaritia:Resource:6>], 
 "potentia 128, instrumentum 96, telum 72 , terminus 64, infernus 72, superbia 48, fames 32, nebrisum 16, ira 8", <AWWayofTime:energyBazookaThirdTier>, 20);
 mods.thaumcraft.Research.addInfusionPage("ENERGYBAZOOKAIII", <AWWayofTime:energyBazookaThirdTier>);
-mods.thaumcraft.Warp.addToResearch("ENERGYBAZOOKAIII", 15);
+mods.thaumcraft.Warp.addToResearch("ENERGYBAZOOKAIII", 20);
 
 // --- Blood Infused Wand Cap
 mods.thaumcraft.Research.clearPages("CAP_blood_iron");
--- a/scripts/Blood-Magic-Thaumcraft.zs
+++ b/scripts/Blood-Magic-Thaumcraft.zs
@@ -1139,7 +1139,7 @@ mods.thaumcraft.Arcane.addShaped("ENERGYBAZOOKAI", <AWWayofTime:energyBazooka>, 
 [<BloodArsenal:blood_infused_diamond_block>,  <AWWayofTime:energyBlasterThirdTier>, <BloodArsenal:blood_infused_diamond_block>],
 [<gregtech:gt.metaitem.01:11978>, <AWWayofTime:demonicTelepositionFocus>, <gregtech:gt.metaitem.01:11978>]]);
 mods.thaumcraft.Research.addArcanePage("ENERGYBAZOOKAI", <AWWayofTime:energyBazooka>);
-mods.thaumcraft.Warp.addToResearch("ENERGYBAZOOKAI", 50);
+mods.thaumcraft.Warp.addToResearch("ENERGYBAZOOKAI", 5);
 
 // --- Energy Bazooka II
 mods.thaumcraft.Research.addResearch("ENERGYBAZOOKAII", "BLOODMAGIC", "potentia 24, instrumentum 21, telum 18 , infernus 15, superbia 12, fames 9, nebrisum 6, ira 3", -14, 0, 4, <AWWayofTime:energyBazookaSecondTier>);
@@ -1154,7 +1154,7 @@ mods.thaumcraft.Infusion.addRecipe("ENERGYBAZOOKAII", <AWWayofTime:energyBazooka
 [<DraconicEvolution:draconium:2>, <DraconicEvolution:draconicCore>, <AWWayofTime:dawnScribeTool>, <DraconicEvolution:wyvernCore>, <AWWayofTime:duskScribeTool>, <DraconicEvolution:draconium:2>, <AWWayofTime:duskScribeTool>, <DraconicEvolution:draconicCore>, <AWWayofTime:dawnScribeTool>, <DraconicEvolution:wyvernCore>], 
 "potentia 96, instrumentum 72, telum 64 , infernus 64, superbia 32, fames 32, nebrisum 16, ira 8", <AWWayofTime:energyBazookaSecondTier>, 15);
 mods.thaumcraft.Research.addInfusionPage("ENERGYBAZOOKAII", <AWWayofTime:energyBazookaSecondTier>);
-mods.thaumcraft.Warp.addToResearch("ENERGYBAZOOKAII", 100);
+mods.thaumcraft.Warp.addToResearch("ENERGYBAZOOKAII", 10);
 
 // --- Energy Bazooka III
 mods.thaumcraft.Research.addResearch("ENERGYBAZOOKAIII", "BLOODMAGIC", "potentia 27, instrumentum 24, telum 21 , terminus 18, infernus 15, superbia 12, fames 9, nebrisum 6, ira 3", -16, 0, 4, <AWWayofTime:energyBazookaThirdTier>);
@@ -1169,7 +1169,7 @@ mods.thaumcraft.Infusion.addRecipe("ENERGYBAZOOKAIII", <AWWayofTime:energyBazook
 [<Avaritia:big_pearl>, <Avaritia:Resource:6>, <DraconicEvolution:awakenedCore>, <Avaritia:Resource_Block>, <Avaritia:Resource:6>, <Avaritia:Resource:6>, <Avaritia:Resource_Block>, <DraconicEvolution:chaoticCore>, <Avaritia:Resource:6>], 
 "potentia 128, instrumentum 96, telum 72 , terminus 64, infernus 72, superbia 48, fames 32, nebrisum 16, ira 8", <AWWayofTime:energyBazookaThirdTier>, 20);
 mods.thaumcraft.Research.addInfusionPage("ENERGYBAZOOKAIII", <AWWayofTime:energyBazookaThirdTier>);
-mods.thaumcraft.Warp.addToResearch("ENERGYBAZOOKAIII", 200);
+mods.thaumcraft.Warp.addToResearch("ENERGYBAZOOKAIII", 15);
 
 // --- Blood Infused Wand Cap
 mods.thaumcraft.Research.clearPages("CAP_blood_iron");
--- a/scripts/Blood-Magic-Thaumcraft.zs
+++ b/scripts/Blood-Magic-Thaumcraft.zs
@@ -47,19 +47,19 @@ mods.thaumcraft.Research.addArcanePage("BLOODALTAR", <AWWayofTime:Altar>);
 mods.thaumcraft.Warp.addToResearch("BLOODALTAR", 5);
 
 // --- Life Infuser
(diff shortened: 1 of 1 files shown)


Dream-Master:
What the other thinking ?

Prometheus0000:
The energy bazookas will be reduced to 5/10/15, because the damage numbers are a lie (it seems).

Dream-Master:
it was 5 10 20 before

GTNH-Colen:
There is already too much warp going around, reducing it will help out a bit. We have so many addons with hundreds and hundreds of researches, we really need to keep this in mind when deciding warp levels for research, it adds up quickly and playing with withers spawning behind you every 5 minutes gets old very fast.

Prometheus0000:
What about adjusting the Warp Theory configs? They allow a 1-200 range for effects. Even just using up to 100 instead of 80 would allow more room for non-witherness.

xTwi:
IMO buffing soap to let it cleanse normal "sticky" warp better but having the book give more permanent warp on progression researches (Blood Magic) instead of "sticky" so that its not "pure tear or nothing". So that soap will actually have a good use. You could also potentially apply a effect to Warp Ward potion effect that it would have a chance to reduce any kind of warp (maybe imbalanced) every minute by 1.

You can increase the Warp Theory all the way to 200 yes, but new effects would be nice to fill that gap.

Prometheus0000:
1) There's no setting for which type of warp, I have no idea how it decides which to give you.
2) I used 20 soap today, and it's around 50% or more chance to remove normal/sticky warp while under warp ward and in a pool of the salted water. I'm not sure it needs to be made higher, but making it cost less essentia would make it easier to get. Or getting more than two. Also, 10 HOURS for the coin quest? Seriously?
3) Having a warp ward potion would be nice, I'm not sure you could make one that just reduces the effects though. Even adding the effect to a potion would require coding I think.
4) Adding more effects would require coding.
5) The config is easy to change

scripts/Blood-Magic-Thaumcraft.zs
```diff
@@ -59,7 +59,7 @@ mods.thaumcraft.Infusion.addRecipe("LIFEINFUSER", <AWWayofTime:Altar>,
 [<minecraft:nether_star>, <gregtech:gt.metaitem.02:22977>, <gregtech:gt.metaitem.02:22977>, <AWWayofTime:imbuedSlate>, <TConstruct:heavyPlate:251>, <TConstruct:heavyPlate:501>, <TConstruct:heavyPlate:251>, <AWWayofTime:imbuedSlate>, <gregtech:gt.metaitem.02:22977>, <gregtech:gt.metaitem.02:22977>], 
 "victus 32, alienis 32, praecantatio 24, auram 16, fames 8, terra 8", <BloodArsenal:life_infuser>, 5);
 mods.thaumcraft.Research.addInfusionPage("LIFEINFUSER", <BloodArsenal:life_infuser>);
-mods.thaumcraft.Warp.addToResearch("LIFEINFUSER", 3);
+mods.thaumcraft.Warp.addToResearch("LIFEINFUSER", 50);
```
Rdjrjqouqcu: Is this supposed to add 50 warp? Maybe you meant to add only 5

Prometheus0000: It's supposed to be 50. It uses infinity and is completely broken. #2834

Prometheus0000: Actually is it fixed now? At the bottom of #3953 it mentions GT items being blacklisted, but can you use it to mess with other stuff?

Rdjrjqouqcu: ... Please at least include a warning or some other way to tell it apart somehow.  Maybe something like a quest or custom icons

Prometheus0000: I do think moving it to NE of the bound armor research might be appropriate, but I'm not 100% sure how to move them. And I'd probably want to add Teleposer research at the same time. Custom icon is no-go for me. Quest might be possible, but it would be UV, and wouldn't save you from getting the warp. I'm planning on doing quest rearranging/adding after this depending on the results.

DoomSquirter:
warp ate my cat
fuck warp

Prometheus0000:
Did you turn the world hole warp effect on?

But seriously, that doesn't really say whether you're for or against the PR, or what you might want changed or why?

scripts/Blood-Magic-Thaumcraft.zs
Rdjrjqouqcu: How about you restricted the research behind some other high level research that could optionally be required for the recipe. My main concern is I would be doing similar researches and nothing would make it apparent that this research is insanely warping. It would be easy to jump waay past 80 warp without realizing it which isn't a fun surprise.

Prometheus0000: It's used to charge the life imbued armor as well as repair things (I thought it charged the bound, I've never used either). So putting it behind that would make sense. But I have to restart the client every time I test it, and that takes forever, so I was putting it off.

yukieiji:
Against this, because:
1. GTNH has various items or machines to remove Warp, such as Pure Tear.
2. Sigil of Compression becomes OP(#4961) if warp is reduced and risk is reduced.

Prometheus0000:
Do you really think that ALL the sigils past the first 2/3 should have warp though? I'm mainly concerned over the amount of warp on them, since that's what you would do BM for. You can't skip to get to the ones you want either. Pure Tear is LuV, and instantly makes warp kinda pointless. Soul Brazier apparently has a range of 3 chunks (and now also requires LuV), and the tumor one was nerfed (but not as much for some reason even though it's got the same bug? It also seems complicated, and I don't want to use something broken). See #5248 and #5404. And the Purification Talisman is made from the tear. 

You shouldn't have to wait until you get a pure tear (LuV) to start BM (or getting use out of it), since the recipes start at MV.

About the sigil of compression, it was already voted to leave it the way it is in the very issue you linked. If you want it nerfed, make a PR to add bedrockium to it like richardhendricks said. I don't have a problem with that.

yukieiji:
There is also WandFocus: Flux Scrubber as a method to remove warp. This can be made at a very early stage(Removal of warp(Temp and Normal) is a probability after upgrading and decontaminating using WandForcus).

Although Sigil of Compression has been voted good for its current BM balance, this change may require some fix.

Prometheus0000:
I didn't know it did that. It seems annoying to purposely create flux to reduce normal warp. Isn't it just a more convoluted soap though?

I'm already at 45+4 after removing around 15 with the old balance, and I don't even have any warpy sigils researched. And 50 has some really nasty effects too. I counted _ninety-six_ warp in just the sigil research, enough to spawn withers, and then some.

Perhaps you could look at it from another angle, which is justifying _any_ warp in the BM research? After all, if the author of BM wanted a warp-like mechanic, they would add it themselves, right?

I agree it was voted based on the current balance, but I believe most of the discussion was about _using_ it, not getting it. Plus my previous comment.

yukieiji:
You don't need to create Flux, just look for the Tainted Biome and decontaminate it there.

If BM is required before IVTier(Osmiridium can be made with IVTier), the amount of warp proposed is considered appropriate.
However, the current requirement for BM is IVTier or later, and Pure Tear can be made without any problems.
Therefore, the current amount of warp that select research while accepting warp befor IVTier and can research without considering warp after IVTier is considered appropriate.

Also, if you want to ease to make of Sigil of Compression, it is necessary to consider recipes, research, LP consumption, compression blacklist etc.

Prometheus0000:
I count tier by power usage, so to me it's LuV. It's not about required or not, you shouldn't have to wait until pure tear to not have ridiculously high warp. I'm not even sure what you mean by 'the current requirement for BM'? Do you mean the quest on the 'Legendary Quests' tab? I don't think you're supposed to start BM _then,_ if you were, the altar/orb would be gated to LuV. Same for the other stuff you need on there you can start before then. And it's not about when you _need_ to get stuff from a mod to proceed, it's about when it unlocks. I'm not entirely certain about your 4th sentence, are you saying to be select about what you choose to research? Because I have been. And you can't for the sigils.

yukieiji:
I'm sorry, my English is not good.
TC research is your own choice, research and unlocks the content.
There is also a choice of unlocking content if you want to get it even if there is warp, and there is also a choice not to do research if you do not accept warp.Unlocking content is when the selected research is completed, not when the research is opened.

The current BM is not required for progress except for "Legendary Quests", otherwise it is only needed when you need it.
The reason that there recipe isn't a high tier material is thought to be for those who want to use BM even if they accept warp.

Prometheus0000:
I understand that it's your choice whether to research it or not, but I just feel the warp is too much, especially since a bunch of new effects were added with Warp Theory that are much worse than originally. If the effects were more spread out it wouldn't be as bad like I mentioned above. Even if it was just 1 point per research it would be too much. If you look at the number of vanilla warp researches per tab it's 2-5-2-9-3-4 (minus mod added researches), so this is way way more).

But there are apparently barely any people using the mod because of the warp and the over-nerfed recipes (that's a whole other discussion though), so I think bringing the level of warp down is a good thing. After all, I believe it contains more warp now than vanilla thaumcraft.

Anyways, I think we're talking in circles now, so I'd like someone else to speak up (hopefully).

yukieiji:
The current BM(I think all GTNH Magic also have) has a dilemma between accepting warp and making it convenient, and the disadvantage of accepting warp. I think it keeps BM balanced.

Your suggestion is great in reducing warp. However, the dilemma has disappeared and the result is likely to be unbalanced (like Sigil of Compression). therefore, it is thought that adjustment including recipe adjustment is also necessary.

I want to hear someone's opinions and suggestions about here as well.

daddy77gaming:
I think part of the issue is most everyone playing with TC and BM is not experienced enough in both mods to know which paths to take and which to avoid to reduce total amounts of warp gained. Meaning a lot of player will research and do other stuff that lets them gain warp thats not needed to progress to the points they want to reach. Experienced players can determine their paths without too much negative warp effects. Most players advocating warp is fine as it is now i feel know what they are doing, just think about players diving into these mods for the first or second time, they are going to suffer hard.

mitchej123:
BM opens up way too late due to the warp, some parts of it should be available earlier; keep anything OP/super powerful gated later.  I like having some alternatives to tech in magic... that don't suck.

Anarack:
I'm of the opinion that the wither affect from warp theory should be removed. Solves the hard limits for warp for all cases not just blood magic. Does not impact official servers as it doesn't work in claimed chunks anyway.

Dream-Master:
No the wither stays! maybe we need to add more weak creatures in lower warp levels and raise the level a bit a withe spawn

GTNH-Colen:
> I'm of the opinion that the wither affect from warp theory should be removed. Solves the hard limits for warp for all cases not just blood magic. Does not impact official servers as it doesn't work in claimed chunks anyway.

Honestly yea this is a pretty good idea, the game becomes basically unplayable with withers spamming at you constantly and while it can be entertaining on officials with the lightning spam on single player I can see players getting their bases utterly destroyed by this which imo is just a terrible mechanic. I basically just ended up saving all the warp for the pure tear and then instantly purged it to avoid this nonsense.

Rdjrjqouqcu:
> No the wither stays! maybe we need to add more weak creatures in lower warp levels and raise the level a bit a withe spawn

Would it make sense to tie in some witchery effects and creatures? I'm thinking the confusion effect (spawns holographic hostiles) would make a good warp effect. Possibly spawning ents would act as an upgrade to the nature warps, however their drop tables would have to be changed to get rid of the mutandis saplings.
In general i feel like the warp effect trigger values are too low. I feel like a simpler balance would be to double the current ones, then fill in the gaps. (Wither at 160, etc). Because there is ~600 warp in the book total, this would let you collect a much more reasonable amount of the research without needing a tear. This way, by chance or by intent, most people would be able to get the important researches before becoming too warped.

Either way, specifically about this PR, i think it makes sense to lower the warp cost of early Blood Magic as done here because BM is not that flexible in the early game. I dont mind having anything beyond bell jars and reagents, and maybe dusk runes be more difficult because that's where BM actually begins to pay off the investment.

Prometheus0000:
I just went to a clean world and used the cheat thaumonomicon, and it's currently 667-512-0. You can clear off the normal warp, but that takes a long time even if you have the items to do so. I also feel that we should push back the effects. Imagine AFKing while doing crops for a minute or two, and coming back to see an acceleration or random trees (changes blocks as well) effect has destroyed them all. I don't dare go try to breed the last couple crops because of this (falling animals, teleporting, and strange wind also destroy them).

yukieiji:
This is the PR of BM, but it is beginning to talk about warp as a whole.
So why not open a new issue and talk about the overall GTNH's warp balance there?

Dream-Master:
@yukieiji sure good idea.
But we need to stick on this pull to make it pull able and balanced for everyone/majority.

Dream-Master:
After we add new effects to warp Theory we can do new balancing

PR #5623, opened 2020-02-20
Update Railcraft.zs

TataTawa:
Cannot get High speed booster track in Assembler, so I changed the recipe equvelantly. I use ingotRedAlloy because it doesn't confuse with other recipes.
I tried a lot of recipes but all the complex recipes in the script cannot be crafted in game. So after several trials I finally found a recipe that can craft in assembler.


richardhendricks:
@TataTawa Can you create an issue to tie with this Pull Request?  That way we can capture it for the changelog.

Dream-Master:
I want this recipe have a control circuit. I dont add recipes without it anymore.

Dream-Master:
fixed in dv

PR #5664, opened 2020-02-29
Remove Nq pumping

0lafe:
Pt2 of overhaul of pumping. Moron added Nq in the first part, but the numbers for outputs from Nq pumping are incredibly high. Gives essentially super simple infinite Nq gens for IV/LuV/zpm, and easier UV. Also avoids smelting, making Nq infinite post IV


Dream-Master:
@0lafe you know better about high tier balancing than me.

GTNH-Colen:
This should at least be put into a discussion thread before a PR, even if I kinda agree.

Dream-Master:
@0lafe can we add another material in exchange ?

0lafe:
Yea, pt3 will be about adding a bunch of other fluids and things to planets. Just need a better understanding of how the generation correlates to what spawns in. Also needs some additions to the fluid scanner. Easier to fix this first though, no idea when I'll be able to add more, or someone else will pick it up

GeicowithaBango:
pls keep in mind to remove also nq+

0lafe:
Nq+ was also removed here

Dream-Master:
@0lafe so far i understand all gt materials can me used

0lafe:
Think so, it's more about figuring out how the config works. Changing the loss/cycle seems to have an impact on rate, and I don't want to be placing deposits I don't understand

GeicowithaBango:
is there any replacement for it tho?

0lafe:
Not yet, no

quedral:
so essentially your solution is to remove all fluid naq and enriched fluid naq from all planets? why was it added in the first place then?

GTNH-Colen:
Only concern I have around this issue is getting enough radioactive material to supply barts vat for endgame biowares. They eat a metric crap ton of the stuff, this was a nice solution to that issue.

0lafe:
>so essentially your solution is to remove all fluid naq and enriched fluid naq from all planets? why was it added in the first place then?

Moron added a lot of fluids and didn't quite know how the outputs of naq would work iirc

>Only concern I have around this issue is getting enough radioactive material to supply barts vat for endgame biowares. They eat a metric crap ton of the stuff, this was a nice solution to that issue.

What are some numbers for it? also which Nq's? You do still get quite a lot from quarrying the dd and such

draknyte1:
Thanks. 

Now I can finish the LNR without worrying about people having infinite fuel for it tiers too early.

Issue #5643, opened 2020-02-23
Discussion on thaumcraft warp and balancing it


Prometheus0000:
Refer to #5604

This is so the discussion on warp balance can be separated from the discussion on warp research specifically for BM.

People have brought up: altering the warp theory configs (currently the highest effect is 80, it can go to 200), adding new effects (such as weaker monsters than the wither, or more 'scary' stuff), whether having a wither spawn is balanced (would blow up your base, at least on SP), the fact that pure tear is all or nothing, adding other stuff to reduce warp somewhat (An easier version of the pure tear that reduces it by, say 5? Something like heart canisters that are equippable, stack, and reduce warp by 1/2/3/etc?), a warp ward potion (presumably easier to get than the salt, or longer lasting), the fact that machines that exist (soul brazier and the tumor) have serious bugs, the very high amount of warp due to large numbers of addons, that people don't use BM because of the warp, that people who aren't familiar with TC can easily screw themselves over with warp, and probably some I missed.

yukieiji:
#### Info on GTNH's warp #### 
・All Warp Amount: 
| Permanent | Normal　|
| ------------- | ------------- |
| 667  | 512  |

・Warp Effect: (+ TC Warp Effect)
| Warp Effect| Warp Amount |
| ------------- | ------------- |
| Fake explosion&Fake Creeper | 10  |
| Spawn bats | 15 |  
| Poison | 16 |
| Jump boost | 18 |
| Rain&Blood&Nausea | 25 | 
| Friendly Creeper | 26 | 
| Lightning |  30 | 
| Livestock rain | 32 | 
| Wind | 32| 
| Blindness | 43 | 
| Random teleport | 45 | 
| Acceleration&Decay&Random tree | 50 | 
| Chest scramble&Spawn Wither | 80 | 

・Item/Machine of reducing Warp
| Item/Machince name | Mod | Type and amount of warp to reduce |
| ------------- | ------------- |  ------------- |
| Sanitizing Soap | TC | Temp only |
| Tabacco Pile(Default) | Thaumic Bases | Temp only(probability) |
| Wand Focus:Flux Scrubber | Thaumic Bases | Normal and Temp(probability) |
| Warped Tumor | Thaumic Horizons | All(up to 50) |
| Soul Brazier | Thaumic Exploration | All |
| Pure Tear | Warp Theory | All |

・Item/Machine of prevent WarpEffect
| Item/Machince name | Mod | 
[...]

PR #5671, opened 2020-03-01
WT script

Prometheus0000:
Script for Warp Theory changes, see #5643.


yukieiji:
I felt that the amount of Essentia was a little small (1.25 times may be just right)
Because with Essentia Locus and Infusion Claw, Infusion can be automated at an early stage and could be the main method of removing warp.

Prometheus0000:
The Locus sounds like a pain to build/use. Have you actually set this kind of thing up for infusion? I think most people would just wait for TE. Keep in mind you still have to wait for the items to smelt down, plus getting them in the first place. I'm not sure you should assume people have either of the advanced furnaces. And doesn't it take another second for every 2-3 essentia to the time to complete? Does it need to be longer when you need to make a bunch? I'm not balancing for WAs here.

That said, if other people agree, then that's fine.

Prometheus0000:
Also, the difficulty is more in the items than the essentia anyway.

Prometheus0000:
@Dream-Master, do you think you could change the 3 instances of 'Pure? Tear' to 'Impure Tear'? They decided to change it to that in the other PR https://github.com/GTNewHorizons/WarpTheory/pull/2. It seems silly just to make another PR just for that.

Dream-Master:
@Prometheus0000 make a new one. Thanks

Dream-Master:
Done with https://github.com/GTNewHorizons/NewHorizons/commit/972ebd5078ac2534fe99cd0f247af713d0295811

PR #5690, opened 2020-03-04
Change Warp Theory Config

Prometheus0000:
See #5643 for the discussion


Dream-Master:
After we add more effects to warp theory we can make a new one

PR #5801, opened 2020-03-23
Craftweaker legacy

seitenca:


leagris:
I see this patch adds `as Int` to some research coordinates but no other coordinates. What is the intent with this?

PR #5877, opened 2020-04-05
Fun ross change

0lafe:
Makes the rocket tier needed for ross 4, as opposed to 3. This pushes lots of free IV machines until the very beginning of LuV. Changing the tier of the machine spawning is hard, so here's this for now


bartimaeusnek:
Make an RFC. Tgus us a rather big change to balance.

Issue #5974, opened 2020-04-24
More underground Fluids


Dream-Master:
I add now To barnadeE the unkown fluid as underground fluid and i found out nothing work on this planet so i fixed it too.

I like to see more usefull but not to op materials as fluids on planets and moons.
Any suggestions?

Prometheus0000:
I believe people were complaining about a lack of hydrogen, and pumping methane and cracking it or something, maybe add Hydrogen directly? Maybe add Sulfuric Acid or Hydrogen Sulfide to Io (it's volcanic right?). Also, maybe Carbon Monoxide and/or Carbon Dioxide as they would probably be realistic to find on planets. Maybe other molten metals? Possibly Distilled Water to Europa (it's full of water, isn't it?) .

feliser:
Milk

0lafe:
Methane would be nice, but H2 pre T7 or so seems kind of OP. H2 makes up a fair bit of the cost of many fuels.

botn365:
methane can alredy be found on titan

Prometheus0000:
Is H2 so OP? It's just easier than water electrolyzing it, right? Which is already infinite. This would just be faster.

Although what I really want is gas giant gas siphoning of various stuff, kind of like in Advanced Rocketry. You could have all kinds of stuff if that was possible. Maybe a new dim like the asteroids one that's supposed to be Jupiter's upper atmosphere or something?

botn365:
water electrolisis does come with a big energy cost and infrastructure so you cant realy compare it to water

feliser:
I agree with 0lfate

GTNH-Colen:
> Maybe other molten metals? Possibly Distilled Water to Europa (it's full of water, isn't it?) .

Perhaps molten iron is an option useful but not gamebreaking? Also would like to see hydrogen later on, perhaps a T7 or T8 planet as it can be quite useful.

quedral:
molten iron/ copper/tin?

iouter:
why not think about raw materials or products in chemical industry

SKYCATV587:
why not add a uua/uum in world ,it can just chance1,maxamount10(or5?),minamount0,even you find it ,it also can't change anything。Just as a symbol of luck

botn365:
uum would be realy OP maby even uua

Extremity4:
[...]

PR #5977, opened 2020-04-25
Nat gas removal

0lafe:
Removes Nat gas spawns in the OW. Kind of OP having a decent fuel that's pumpable 1 step before generation, due to rig rates.


Dream-Master:
Sure but why we don't move it to another planet or moon @0lafe ?

quedral:
why don't you adjust burn value instead of removing? If you have oil somewhere you will have natural gas.   If you remove natural gas then you have remove oil as well.

Again this is one of the stupidest suggestions that i've seen in months. Put this on discord vote and see what you get.

Prometheus0000:
To make it truly useful, you need to turn it into refinery gas, right? Maybe adjust those values to be more expensive?

Dream-Master:
no as the name said refinery gas need some processing. I am against it. But I am also against it remove natural gas complete. @Prometheus0000

Prometheus0000:
My understanding is that you pump Natural Gas (which has a burn value of 20,000 EU), and it's only use is burning or turning it into Refinery Gas (which has a burn value of 160,000 EU). You can use it for other stuff like HOG or cracking, but Refinery Gas' primary use is burning it as fuel as well. I'm saying maybe if the process of turning it from NG to RG was more expensive, it might be more balanced.

Extremity4:
I’m not too sure about the implementation, but I agree with the concept behind this nerf at least. Ideally, refinery gas should be inferior to nitro boosted diesel, benzene, or other more complex refined fuels due to how much simpler it is to produce. In practice, however, refinery gas is so incredibly easy to make and available in such massive quantities that anybody in their right mind can see that it makes more sense to use it over literally any other burnable fuel in MV-IV.
In my own experience in a SSP world, using natural gas allowed me to produce THOUSANDS of eu/t while in MV TIER, with nothing more than an lv chemical reactor and a few lv electrolyzers! It was so powerful that this tiny little setup carried me all the way to naquadah generators in IV. I set up a benzene line in HV and a cetane boosted biodiesel line later in IV just for the fun of it, but despite the higher level of complexity (vastly higher in the case of cetane) and more expensive machines these lines required they ended up being inferior to the MV natural gas setup!
I don’t know whether removing natural gas generation outright from the OW is the best way to go about dealing with it (perhaps gas fields can just be made much smaller?), but it seems clear to me that something needs adjusting here.

Dream-Master:
ok i close this now and open up a new one with addition an cahnges suggested in RFC issue
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/5974

PR #6003, opened 2020-04-28
Blood Magic Questline change

Prometheus0000:
I also reused three quests, so optimally they'd be reset. If that's possible, which it probably isn't? The T2, T3, and T4 sigils ones (228, 406, and 420). For the first two I added more sigils of that tier to complete, whereas the third has changed from the magnetism sigil (T3) to all the T4 ones. Also, BQ seems to have resorted some unrelated stuff possibly because of different BQ versions. I also changed the alchemy reagent quest to be inline with the amount of simple reagent made, and changed the reagents to something actually useful.


Dream-Master:
will check at the weekende. Less time during the week.

PR #6082, opened 2020-05-08
oredict the assembler crafting table recipe

jacobrham:
Oredict the crafting table recipe to support modded logs in the assembler. Something to be aware of: this replacement adds several (62) more recipe pages in NEI.


jacobrham:
> The Assembler as most Gregtech recipes map populated from MineTweaker only create recipes for the first entry in oredictionary.

Does that mean that GT machines won't make recipes based on the entire ore dictionary? I tested this script change on my machine and the recipes registered without issue.

Dream-Master:
@jacobrham @leagris but this should work. Lets test it.

Dream-Master:
@jacobrham please test the script and report if it worked. Thanks

Issue #6115, opened 2020-05-15
[Suggestion]Add BacterialVat quest.


yukieiji:
#### Which modpack version are you using?
2.0.9.0Beta2
#
#### What do you suggest instead/what changes do you propose?
Since Bacterial Vat is now needed to produce BioMedium, I believe a quest with instructions on how to use it, etc. is needed.
In fact, I don't even know how to make it work at all.

bartimaeusnek:
Step1: Build it.
Step2: Throw Items, Fluids and if nessecary radiation material in it. ( You will need a Petri dish with a culture, which would need a seperate Questline aswell)
Step3: keep output hatch at HALF FILLED LEVEL for optimal efficiency
Step4: ???
Step5: Profit

chochem:
yeah, as bart says, a questline for the BioLab would be great too.

It would be great if the quest for the Vat has some more explanations then what bart wrote above, like how precise is this *half filled* requirement, or what exactly means *optimal efficiency*. Was it x1000 (or x100?) for input+output fluid while keeping solids + recipe time fixed?? I am not sure, I havent used/tested it quite yet.

Some more info on the radio hatch would be great too. E.g. it only takes rods/long rods. Is there an overview of Sievert values?

chochem:
ok, I got some of the info for the quest from bot. Lets just collect a few pieces of info that need to be in the vat quest:

- keep output hatch at HALF FILLED LEVEL for optimal efficiency (how precise?)
- optimal efficiency means: 1000x input and output fluid while keeping solids + recipe time fixed.
- Sievert values are just the number of the element (whith nq+ at 140 and nq* at 150)

Prometheus0000:
- What about the biovat stuff? It seems really complicated with the DNA samples and plasmids and the different component/circuit/whatever they were called
- Presumably you would use a large output hatch and turn off the machine when it's below X amount in the hatch, possibly with a fluid regulator?
- Does it need to be EXACTLY half filled, I thought it was half or more?
- Does the multiplier scale, or is it just 1000x all at once?
- Recommend hatches for in/output for the 1000xing, so it can actually hold that much. Or was it like 1L at a time, I forget?
- Recommend a method of keeping it at half-filled that's been tested to work.
- What recipes are useful, and what are required, and what unlocks when.
- Is there a difference in normal/long rods? Time, radiation amount?
- Mention the glass tiers, and what requires what especially if there are followup quest for different fluids
[...]

PR #6164, opened 2020-05-26
Fix BacterialVat Quest and Add Glass-Tier Quest

yukieiji:
- Fix BacterialVat Quest sentences
- Move The Clean Lab Quest to IV Tab
- Add Glass-Tiers Quests


Prometheus0000:
You should try to put the issue number in a commit message or the comment so it links back to it so people know there's a related issue so it can be looked at and also marked as fixedindev. #6115. I still think that making raw growth medium should be a quest for the tier it's used in rather than saving it up for multiple tiers where it has no use. But I'd like to see what @Dream-Master and @bartimaeusnek say.

Issue #6161, opened 2020-05-26
Outline for future LuV-UV quests


chochem:
-- updated 07 June --

Ok, so @Prometheus0000 wanted to know a bit more details of the progression line for future quests starting from the current LuV questbook onward.

If only 2-3 quests are added it would be best to just turn the LuV tab into a LuV+ tab, but I am thinking a bit bigger here, so we should probably have dedicated ZPM and UV tabs. At most this should go up to UHV. But an extra Endgame Goals tab beyond that is a really good idea: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6172

My general idea would be that these quests do not ask for single block machines where multis are available, and generally focus on new machines and less obvious concepts.

related issues: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6128, https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6130, https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6115, https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6127

Edit: I also like the idea of integrating the *Fusion*-tab as well as the *An Age of Tech*-tab, as those only have a very small number of quests.

---
**LuV Tab**

- a quest early in the LuV tab to make purple rocket fuel (CN3H7O3) needed for T5/T6 rocket (probably better space race tab)

- a quest to make an LuV energy hatch as those are much more important than any machines in this tier. Should be behind the existing LuV-SC wire quest and behind LuV assembler. (add as requirement for starting fusion)

- a quest early in the LuV tab to obtain a small amount of trinium from deep iron (small ore!) and make a bit of naquadah alloy. (add as requirement for starting fusion)

Integrate quests from Fusion tab:
 - starting fusion. (move over) Correct typo: each energy hatch only has a capability of 10MEU (not 160MEU)

 - *Making power with your plasma in LuV*. move over and add the 'in LuV' to the name. Might also need an overhaul, the best option for a dynamo is a buffered LuV Dynamo (the quest talks about UV and demands a normal LuV dynamo) 

- *Turbine time in LuV*. move over and add the 'in LuV' to the name. Change the text to talk specifically about materials available at this stage. Best option imo: normal draconium turbine.

[...]

PR #6199, opened 2020-06-03
Quest additions+fixes

Prometheus0000:
A few fixes, Dream asked me to add the books, and mostly adding quest stuff from #6161.


yukieiji:
It's harder to see than the previous LuV quest line, is this okay @Dream-Master ?

prev:

after merging:


Also, why did you lower the tier of Radiohatch? It's a quest that's open to EVs, so I was going for the EV hatch.

Prometheus0000:
It seems I forgot to add my last changes, the mysterious dust quest in the bottom left.

Also, I had a huge mess of a time getting stuff to fit right, feel free to try to open it up more after I send my last changes.

I lowered it because it's supposed to be HV, that's when the vat opens up? Plus I don't think there's any meaning in it being EV, it doesn't change anything really.

Prometheus0000:
About the hatch, at hv it holds 1, but ev holds 2, but it still only uses 1 up at a time, so it'll just grab more from the stack (or lower amount) you put in. If you want to use long rods it would matter, but I think people would rather have the granularity that comes with the normal rods.

yukieiji:
The EV is the tier that will unlock radioactive materials such as Uranium, and the reward is a rod of Uranium, so the quest of RadioHatch is set to EV.
If you think about it, RadioHatch's tier could be just about anything. Can't I do the RadioHatch task in any of the tiers?

Prometheus0000:
Yes, it could be any of them, but the HV one is cheaper (and realistic to make unlike UV for example), plus they're actually worse as you go up the tiers for the reason I mentioned.

Would you actually make the EV one? I'd still use the HV one at UV, since there's no reason to make or use the higher tier ones.

Also, the first material on the list is Bismuth, from the Nether. Plus a bunch of misc ones I forget/don't know.

PR #6200, opened 2020-06-03
Forgot to commit final changes for previous PR, added them

Prometheus0000:


yukieiji:
I still think it's hard to see compared to the previous LuV quest lines and other tiered lines, but what do you think, @Dream-Master ?

prev:

After merge:

Prometheus0000:
Are you suggesting removing them or something? Like I said already, if you want to space them out, feel free once Dream has finished going over them, maybe he'll do it himself.

The problem is that there were too many things that had requirements that required me to contort the quests around, like europium, field emitters, and superconductors.

And some don't have any requirement really, so they're very awkward, like the circuit boards and ASMDs.

yukieiji:
I think you should make the flow easier to understand, rather than give it more space.
For example, as shown in the image.

Prometheus0000:
I admit that making things look nice is definitely not my strong suit. This looks mostly good, though you should probably raise the RNG quest up so you can tell you have to do all 3 quests.

Also, I didn't add quest rewards, so if you submit a PR while Dream is adding them, it'd be a huge mess, so you should wait.

PR #6202, opened 2020-06-04
Update LuV quest line

yukieiji:
- LuV energy hatch quest requires LuV assembler quest to open(to craft Ludicrous Voltage Coil)
- Imprints Circuit quest requires all Crystalprocessor quests to open
- Fix a few spaces
- Fix Crystalprocessor mainframe Quest
- Change the task of LuV Laser Quest


Prometheus0000:
Dream is still working on it, so... yeah, that's not good.

1) ok
2) I specifically made it _not_ like that so that you could get the reduced costs on the circuits as you go through the quests instead of having to make them all at the increased cost before getting the imprints
3) I don't see the spaces?
4) ?
5) Why? If you have a LPF, you shouldn't be forced to make one just to complete the quest

yukieiji:
> I specifically made it not like that so that you could get the reduced costs on the circuits as you go through the quests instead of having to make them all at the increased cost before getting the imprints

If you do that, shouldn't you only require one circuit in the ImprintsCircuit's quest and only one CAL's quest?


>Why? If you have a LPF, you shouldn't be forced to make one just to complete the quest

Crystal Processing Unit needed a LuV laser, so I changed it to that.

Prometheus0000:
Unlike previous versions of BQ you can complete the objectives on each page separately (instead of having to have everything at once), so there's no problem. Also, I realized it would be a problem after completing the quests, it originally required the 4th circuit, but I made it the 1st for the reason I mentioned.

The LuV Laser quest originally required you to get the Precision Laser Engraver (luv version), but I made it a checkbox task because the LPF has a PLE mode, and why would you want to make the single block when you have the multi already? You aren't making singleblock CRs instead of the LCR, right? But the main point is telling people that they have the option of doing either without making it two quests. So if you don't want to make the LPF (if you don't have it already) can make the singleblock if they really want to, or become aware that they can make/use it for this.

chochem:
> Crystal Processing Unit needed a LuV laser, so I changed it to that.

it doesnt, no. there is the LPF. LuV Laser should not be mandatory. this quest progression was just fixed a few weeks back, dont reintroduce that.

Dream-Master:
If I push code I stop working. I let you know when I start again

Prometheus0000:
I thought you meant when you said you were going to add quests earlier, you meant you were about to start working on it.

Prometheus0000:
I took a look ingame:

1) Still fine
2) If I realized the problem beforehand I might have made it have an equivalent quest for each, but that's too troublesome, and kinda pointless for the reasons I already mentioned. The next series will be Mats->Circuit->Completed Imprint for each quest in ZPM, but I don't think that would fit here
3) I think you mean you spaced it differently to account for moving the assembler quest
4) That 1 was really hard to see ingame
5) Still should be a checkbox task since the alternative is making it an OR with the LPF. But that's weird, since it's IV, and not new.

yukieiji:
@Dream-Master Okay, I'll reopen the PR after I heard from you.

Dream-Master:
why you close it ? Didnt you grab the lastest file ? @yukieiji

Prometheus0000:
He did:
After Dream updated it:

After Yukieiji altered it:


Though I'm not sure why you moved the stuff on the right over, but that's kind of irrelevant.

Dream-Master:
ok sure

Issue #5732, opened 2020-03-11
Skullfire Sword Rework


Houstonruss:
The Skullfire Sword (SFS) in its current state is extremely unlikely to be used for its intended purpose to get skulls to farm nether stars. Players will often go towards a ticon tool with as many beheading upgrades as they can get, and will be way easier to obtain than the SFS. This problem is cause by the necessity of the DireCrafting Table and additional matrix ingots. My suggestion is to incorporate the SFS into progression as a means of actually obtaining a DireCrafting Table. The sword should use materials from HV or earlier, and ideally fit the theme of the sword.

![Skullfire](https://user-images.githubusercontent.com/50430417/76452058-d8b9d600-63a6-11ea-9263-a6d1c657ac9b.png)

I have chosen some necrotic bones, lich bones, diamond lattices and a large Dark/Black Steel sword blade in an HV assembler as a way to get you to visit the Nether and kill some skeletons, Kill a lich, and get stainless before being allowed to get stars from the SFS. The amounts of diamond lattice and necrotic bones should be higher, but lich bones can be problematic if using too many of them.
No nether stars should be used in the recipe, it would defeat the purpose.

quedral:
actually not a bad idea.

Prometheus0000:
I say replace the Diamond Lattice with the Crystal Matrix Ingot so you have to kill the wither twice first. I don't think including nether stars in the recipe defeats the purpose, it's the large number of them.

Houstonruss:
> I say replace the Diamond Lattice with the Crystal Matrix Ingot so you have to kill the wither twice first. I don't think including nether stars in the recipe defeats the purpose, it's the large number of them.

Then you'd run into the problem of people still going for a beheading cleaver anyways and just ignoring a specialty sword that can't even 1-hit skeletons. You have to kill (depending on luck/looting) hundereds of wither keletons to even get the stars. Also it's worth noting that the salis mundus recipe is also a meme because it takes 2 nether stars aswell, which have the same beheading issue.

mitchej123:
+1 to making wither skellies less grindy and an alternative to ticon

Dream-Master:
I am ok with it but not a assembler recipe please.

PR #6409, opened 2020-07-19
Skullfire sword

Prometheus0000:
See #5732. Should be easier to make before the DCT now, also more magical-ish too.

image


Dream-Master:
@Prometheus0000 which sword is in the middle ?

Prometheus0000:
Fiery Sword. I confirmed that you can make it out of the fiery blood from the Hydra, and that the SFS crafts and shows up in the book correctly.

Issue #6424, opened 2020-07-22
Crafting Mud Balls causes buckets to duplicate


Qawler:
#### Which modpack version are you using?
2.0.9.0

#### What did you try to do, and what did you expect to happen?
While crafting mudballs with a water bucket and dirt I expected to get the two mudballs and one bucket

#### What happened instead? (Attach screenshots if needed)
Instead, I got 2 mudballs and 2 buckets back

chochem:
please use the search function. this is ticket no. 6 or 7 I think. see https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6354 for example.

Qawler:
Oh, I'm sorry

Dream-Master:
@Qawler if you find any new bugs feel free to open up issues. Thanks

Issue #6514, opened 2020-08-17
Dupe Bug with crafting Mud Balls in all crafting inventories giving x2 Buckets


KaneHart:
#### Which modpack version are you using?
2.0.9.0
#
#### If in multiplayer; On which server does this happen? 
Yes
#
#### What did you try to do, and what did you expect to happen?
Crafting Mud Balls with 1 Dirt 1 Bucket of Water in my crafting, normal crafting table even tinkers crafting table. I should get 1 bucket back but I get 2 Buckets Back instead of 1
#
#### What happened instead? (Attach screenshots if needed)
I dupe the buckets. 
#
#### What do you suggest instead/what changes do you propose?
I assume either there is a recipe / mechanic that does this already and it's duping or a typo somewhere.

![image](https://user-images.githubusercontent.com/425853/90359669-8781d280-e027-11ea-9e4e-ecd57f4b35c7.png)

GTNH-Colen:
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6424
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6354
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6256
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6230
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/5727
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/5591
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6089

I think it's safe to say this is a known issue at this point 😆

Dream-Master:
@KaneHart You Mud Ball duper :P

KaneHart:
Opps haha :)

KaneHart:
Can I please have the link to the actual open ticket for this bug and the fixed commit.

Prometheus0000:
Reopened #5591 as that appears to be the original issue.

Issue #6354, opened 2020-07-02
Crafting mud balls duplicates buckets 


909-xdd:
#### Which modpack version are you using?
2.0.9.0QF4
#
#### If in multiplayer; On which server does this happen? 
single player world
#
#### What did you try to do, and what did you expect to happen?
I tried to craft mud balls with water bucket and dirt.
#
#### What happened instead? (Attach screenshots if needed)
The mud balls got crafted but I got an extra bucket after.
#

chochem:
Duplicate.
Please use the search function next time. there is like 5 tickets already for this. For example https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6256.

Issue #6256, opened 2020-06-12
Bucket Dupe


mightydanp:
#### Which modpack version are you using?
2.0.9.0QF3
#
#### If in multiplayer; On which server does this happen? 
no
#
#### What did you try to do, and what did you expect to happen?
craft mud, give one bucket back
#
#### What happened instead? (Attach screenshots if needed)
gave 2 buckets back
#
#### What do you suggest instead/what changes do you propose?
fix it duping is bad

draknyte1:
Known issue.. did you look at the several existing tickets?

Prometheus0000:
#6230 #6089 #5727 #5591

PR #6523, opened 2020-08-19
Update Biome-o-Plenty.zs

KaneHart:
If you include a Bucket of Water and it also looks like Clay Bucket of Water you automatically get the bucket back without needing to script it in. This why it was duplicating. 
This not the case with all mods but it's the case here. You can also add a .noReturn() for those times you want it to be consumed.

This fixes:
#6424
#6354
#6256
#6230
#5727
#5591
#6089
#6514


KaneHart:
Redoing this this extra line I added bad me!

Issue #6230, opened 2020-06-09
Workbench Mud ball recipe. Buckets duplicating


kionter:
#### Which modpack version are you using?
2.0.9.0
#
#### If in multiplayer; On which server does this happen? 
Single player
#
#### What did you try to do, and what did you expect to happen?
Created the recipe for Mud ball on forestry's workbech. 1 Dirt, 1 water bucket. 
Placed the dirt and water bucket on the workbench inventory and when "creating" the mud I expected to have the completed Mud ball and an empty bucket. 

#
#### What happened instead? (Attach screenshots if needed)
An empty bucket returns to the workbench inventory (as expected) but also another empty bucket (duplicated) is returned to the character inventory. 
#
#### What do you suggest instead/what changes do you propose?
The final result should be 1 Mud Ball and 1 empty bucket.

kionter:
I'm unable to reproduce the same behavior after restarting the game and not getting duplicate buckets anymore.

Issue #6089, opened 2020-05-10
Bucket/Clay bucket bag/dupe


f0Re3t:
#### Which modpack version are you using?
v2.0.8.3
#
#### If in multiplayer; On which server does this happen? 
This happens both in a single player game and on the server
#
#### What did you try to do, and what did you expect to happen?
I tried to craft a Mud Ball from a block of dirt and a bucket of water
#
#### What happened instead? (Attach screenshots if needed)
The item is being crafted, but a bucket of water remains in the crafting grid, as well as giving an empty bucket to the inventory
#
#### What do you suggest instead/what changes do you propose?
Change the reuse function to giveBack or transformReplace

Dream-Master:
can you test with a newer version. I rember i fix some things. @f0Re3t

f0Re3t:
Yes, I checked the script files on v2.0.8.8 and you really fixed it, in the game it also works without bugs
@Dream-Master

Issue #5727, opened 2020-03-09
Mudball Bucket Duplication Glitch


ImVaskel:
#### Which modpack version are you using?
2.0.8.3
#
#### If in multiplayer; On which server does this happen? 
Modded Minecraft Club
#
#### What did you try to do, and what did you expect to happen?
I was making mud balls for the quest for the armor. I just expected to make some mud balls
#
#### What happened instead? (Attach screenshots if needed)
Instead I made the mud balls and then I had the amount of buckets in my inventory as dirt I used (eg 40 buckets for 40 dirt)
#

Dream-Master:
fixed. Try with 2.0.8.5 or 2.0.8.6

Issue #5591, opened 2020-02-16
Bucket Duping


Zystus:
so when I was playing GT:NH I decided to craft the mud armor just for the quest I had a bucket from a reward bag so I did the dirt+Water Bucket recipe when I crafted mud something strange happened the water bucket I had stayed and I had +1 empty bucket I had ONLY 1 bucket before I did this craft because of the crafts I now have 14 buckets I am playing in a Single Player world 

![image](https://user-images.githubusercontent.com/31573458/74605064-628ec000-50d5-11ea-9c9d-d7cbaf557e16.png)
![image](https://user-images.githubusercontent.com/31573458/74605069-6a4e6480-50d5-11ea-9d99-dd09b52b1b09.png)
![image](https://user-images.githubusercontent.com/31573458/74605071-6e7a8200-50d5-11ea-8c4e-d0f2b2f0f80c.png)
![image](https://user-images.githubusercontent.com/31573458/74605073-73d7cc80-50d5-11ea-8e4b-31adf8720278.png)

MadeofGold:
Tested , it works even in official servers (Delta)

Dream-Master:
need to test all other reuse recipes

EldrinnElantey:
The fix for this bug has already been added to the recipe script here https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/commit/47ca6245d5cdcaed51cfa6cd8373d3fdf61c1efe. Maybe it's time to close this issue?

PR #6525, opened 2020-08-19
Fixed Water Bucket Dupe with Mud Balls

KaneHart:
If you include a Bucket of Water and it also looks like Clay Bucket of Water you automatically get the bucket back without needing to script it in. This why it was duplicating.
This not the case with all mods but it's the case here. You can also add a .noReturn() for those times you want it to be consumed.

This fixes:
#6424
#6354
#6256
#6230
#5727
#5591
#6089
#6514


Dream-Master:
@KaneHart are you tested if the fix work ?

KaneHart:
@Dream-Master Yup both in SP and MP and works just fine!

Dream-Master (approved):

Issue #3909, opened 2018-12-16
The one ring op


0lafe:
#### Which modpack version are you using?
2.0.6.2
#
#### What do you suggest instead/what changes do you propose?

Probably just remove the ring. It's a fairly early loot drop and gives invincibility, feeds you, and makes mobs passive. Unless that's just warping stuff and you can do it other ways, the ring seems pretty OP. Apart from giving flight, that's better than infinity armor.

bartimaeusnek:
how about i throw blindness, nausea and some other annoying potion effect on it, in addition to the warp? then you will only use it in rare situations...

Dream-Master:
good idea

richardhendricks:
Nah, just have it give you permanent warp each time you put it on, like 5 points, on top of 5 points of temporary warp.  That way the problem naturally resolves itself.

Hakameda:
It gives blindness , nausea and poison already, and it doesn't make a difference. Can probably skip space suits too.

Hakameda:
It was at like 160 warping, it actually sees to reduce the frequency of regular warp events. Though double withers in the base was meh

0lafe:
>Nah, just have it give you permanent warp each time you put it on, like 5 points, on top of 5 points of temporary warp.

Then just put it on once and never remove it, never gain warp? Also, the whole warp part is almost a non factor once the talismen comes into play, in EV/IV, still well before infinity. Use it for a little, take it off and wait a little until you have no warp.

Anarack:
Make it give normal/sticky warp not perm, still removable just harder too and needs soap/arcane spa to do so, otherwise someone is going to brick their character with it.

Hakameda:
The current warp isn't really punishing anyway

richardhendricks:
It only gets removed by death @0lafe , and isn't bricking a character kinda the whole point?

0lafe:
Yeah it's comes off faster than a lubed up Winnie the Pooh stuck in a honey hole. If the warp goes away when the ring is gone, that's kind of an issue. I feel like to make an invincibility ring balanced at EVish, it would need to have so many downsides. Blindness is a good start, but how does that work with constant NV? Maybe don't give a ring that makes you invincible, or have it be a craft instead of a loot drop. Right now you pretty much get it for free, and end up making the talismen anyway, so it's almost free invincibility.

richardhendricks:
[...]

PR #6661, opened 2020-09-20
Disable One Ring

Prometheus0000:
Disables getting One Ring in Dungeon Chests

Because it's super OP, and mitch said it was already supposed to be disabled.


Rakdarian:
No.

Prometheus0000:
Not disabled enough? Should we remove the item from people inventories and chests too?

Or you could give more of a comment than just 'no'.

mitchej123:
Here's the last discussion on the topic - https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/3909

Prometheus0000:
If someone has the time to change it, that's great. But it doesn't seem that way right now, so let's disable it until then.

Rakdarian:
No as in, dont disable it.

Prometheus0000:
_Why?_

mitchej123:
+1 to disabling it.  Broken AF

Rakdarian:
Because its a fan favorite and enables variant ways of play. It also allows more casual players to move passed the early game better through exploration. Additionally it encourages early game exploration in general

Prometheus0000:
It makes you _invincible_. We don't have infernals and special mobs just so you can use the ring to not have to worry about them. GTNH != Casual. It's _supposed_ to be hard.

If you wanna re-code it so it raises your warp 1 every minute, plus other nasty effects, maybe it'd be worth keeping around. Any other nerf to it's functionality would basically be removing the invincibility, making it useless. Or making it a UHV recipe or something.

ScriptedPiky:
Agreed with the disabling until an actual rework is done for the One Ring. Another additional problem with the said item is practically ignore some of the warp effects on public or private servers running plugins, namely, town plugins that prevent mobs from spawning or entering their town, effectively negating a dangerous off-set of the One Ring. With a similar problem being that players can use said *invincibility*, not purely *invisibility*, to go off into life-threating areas without a single worry. Going to Venus or Cent. Bb that kills you instantly regardless of most armor or how much health without having the right suit on? Heck, the One Ring prevents that entirely, apparently. Aside from the dangerous warp increase, especially without abusing town plugins or such, the One Ring is too much of a benefit that should not offer *invinicibility* as a loot-item. It's kinda like applying a severely nerfed part of the Infinity Armor but to a single ring.

Dream-Master:
@Prometheus0000 you need to disable or remove the quest also if the ring not exists.

NoX-programer:
I voting against any nerf of a ring (in this case) including ++warp, but voting for hard recipe for game balance.

Prometheus0000:
? A recipe would be a nerf though.

NoX-programer:
The nerf of recipe is the acceptible, I'm sure. The nerf of functionality is the evil.

Rakdarian:
The core game also shouldnt be changed to cater to servers. If the downsides of one ring are mitigated by the fact that some servers offer overpowered protection thats the fault of the server, not the one ring. And personally servers should be anarchy anyways. You dont make the teleporting options easier because /home and /warp [location] exist and needs to compete, and you dont disable one ring because some servers offer overpowered protection

mitchej123:
One can enable it on their own server if they want it.  The pack has always catered to dream's vision and the official servers.

boubou19:
That's not even a matter of "servers offering overpowered protection". You simply become invincible, so ez bosses, no need to worry about oxygen, etc.

Rakdarian:
Actually if you make the damage from oxygen marked as the same as from the chaos dragon projectiles itll hit through one ring. I should also mention that runic shielding also gives protection agaisnt lack of oxygen on other planets so thats something that should be considered.

I was mainly addressing that if an argument is trying to say that because of how towny works with preventing warp effects from going off then the issue isnt with the ring, its with towny. Now if the warp effects arent enough of a discouragement thats another story. but im mainly concerned with the "Towny makes it overpowered so we should disable the ring" Id rather disable towny tyvm

mitchej123:
The ring makes the ring overpowered.  It's unbalanced & too easy to get, we should remove it.

Dream-Master:
ok i close this now. We change the chest llot and make a recipe

Dream-Master:
vote was clear

Dream-Master:
image

Issue #2070, opened 2017-10-09
[Bug] Automagy Glowing Amber Research missing aspect


DukeGrimm:
#### Which modpack version are you using?
2.0.1.0
#
#### If in multiplayer; On which server does this happen? 
Stone Legion
#
#### What did you try to do, and what did you expect to happen?
I was trying to research Glowing Amber. I placed the research on the research table. There should be two or more aspects you need to connect in the research grid in order to complete your research into that subject.
#
#### What happend instead? (Attach screenshots if needed)
The research grid opened but only had one aspect in the grid. T His did not prevent me from researching it. I was able to research it by adding one aspect to connect to the one on the grid.
#
#### What do you suggest instead/what changes do you propose?
The research recipe needs to have one more aspect added for glowing amber.

Dream-Master:
Automagy not neferd rn. Will do this in the future

SirAzazel:
Not a bug.
```java
(new ModResearchItem("NITORLIGHT", (new AspectList()).add(Aspect.LIGHT, 2), -5, 0, 1, new ItemStack(ModBlocks.translucent, 1, 32767))).createPages(new Object[]{"1", new ItemStack(ModBlocks.translucent), new ItemStack(ModBlocks.translucent, 1, 1)}).setParentsHidden(new String[]{"NITOR"}).setSecondary().registerResearchItem();
```

Dream-Master:
Ok but i will change the recipes anyway so for me a reminder.

SirAzazel:
Whats the point though? block doesnt do anything much, its just amber block.. that glows. thats all. no need to overcomplicate it, default recipe is already nerfed due to materials used for it being harder to make.

Dream-Master:
Automagy not nerfed yet will do later

PR #6669, opened 2020-09-22
Add Mark stale issues github action

Glease:
This essentially creates a stale bot that mark stale issues and close them if no activity happens after marking.

An issue is considered stale when
1. It is a balancing related change and it has not been updated in 90 days.
2. It is related to an official server and it has not been updated in 90 days.
3. It cannot be reproduced and it has not been updated in 180 days.

The bot will run 12 times per day, as we have quite a few issues to handle.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,30 @@
+name: Mark stale issues
+
+on:
+  schedule:
+  - cron: "30 */2 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 3 days'
+        days-before-stale: 90
+        days-before-close: 3
+        stale-issue-label: 'stale'
+        only-labels: 'addition,additional Mod,Add Tooltip,Delta,enhancement,Epsilon,Eta,EU Live Server,Idea,Nerf,New Quests,New recipe,New Tooltip,Open for discussion,RFC (request for comment),suggestion'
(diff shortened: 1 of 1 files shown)


jackowski626:
Point 1. and 3. are great imo, idk about 2.

Prometheus0000:
I'd feel better about this if it's down _after_ we work through them a bunch first. I tend to ignore closed stuff, and I imagine others are the same. If they go to search for an issue, they'll probably look at open issues, not closed ones. And then we'd be marking as duplicate and linking to a closed stale issue.

Glease:
Then we reopen the closed stale issue as it now gets some more activity <del>then get closed after another 90 days due to stale</del>

It is feature requests/can't reproduce after all. Normal bug reports are exempt from this process.

Prometheus0000:
OK, I thought the AND was an OR in 3, so it would apply to anything, but I guess not.

DualBallers:
Will the stale marking be removed automatically if the issue is 1) bumped, 2) referenced from other (duplicate) issue, 3) gets a commit?

boubou19:
we have issues opened years ago that still apply tho

Prometheus0000:
Maybe we could add a Don't Close label it would ignore?

Glease:
@DualBallers Getting commented will remove the stale marking next time the bot is ran, but I'm not sure being referenced will do the same. Chances are it will not as you can reference the issue at your own repo (security breach)
@boubou19 yes but this bot will ignore fixable bug reports.

Glease:
If you want to keep it open then add reminder tag as in #2070

GTNH-Afx237v7:
I agree with Prometheus, better run through a bunch and then let the bot do its work. As for item 3. I find 180 days to be a lot, would reduce it to 90 days as well.

Dream-Master:
180 days is fine many issues more years old

Glease:
@Dream-Master The label names are updated to match the new fancier ones. If you see it does not conflict with your intention and the suggestions accepted then you can merge it at any moment.

Dream-Master (approved):

PR #6686, opened 2020-09-24
disable WG transmutation

0lafe:
Should be a "fix" for OP transmutations from WG, until it's either balanced or removed differently

### DIFF
--- a/config/WitchingGadgets.cfg
+++ b/config/WitchingGadgets.cfg
@@ -32,7 +32,7 @@ modules {
 
 "ore/crucible" {
     B:"Enable clusters"=true
-    B:"Enable transmutations"=true
+    B:"Enable transmutations"=false
     S:"Not Trippling Cluster List" <
         Infinity
         Infinity Catalyst


bartimaeusnek (changes requested):
no.

0lafe:
Still haven't seen that entire rework of essentia gen, maybe then we can allow these transmutations lol

bartimaeusnek:
> 
> 
> Still haven't seen that entire rework of essentia gen, maybe then we can allow these transmutations lol

Go ahead, that would be the correct fix. do it instead of destructively turning off features you dont like.

0lafe:
I don't think that method is reasonable or achievable. I'm referring to what you always brought up as a counter point to how dumbly OP much of this is. I'm not really sure how you can see most of these transmutations being OK

mitchej123:
It should be balanced, not removed.

0lafe:
I agree, but the balance is difficult for this, and would need a LOT of stuff changed. I don't think leaving it in an OP state is better than removing it though. If lots of thaum got changed I could see it coming back in a balanced state, but I don't think those changes would ever happen

mitchej123:
I'm fine with having alternatives to tech, gogo magic!

Rakdarian:
Leave the magic alone unless you want to take the initiative and rebuild it from the ground up.

Issue #6562, opened 2020-08-31
Diamond Drill Recipe Broken


ShinoGFX:
#### Which modpack version are you using?

2.0.9.0

#### What did you try to do, and what did you expect to happen?

Attempted to craft the diamond chainsaw from the thaumcraft mods, I expected to be able to craft it. Right research, correct items, everything. But couldn’t craft it at all. Even in a creative world with everything unlocked quest achievement and research wise

#
#### What do you suggest instead/what changes do you propose?

Look into this and make the recipe actually craftable/Fix the crafting recipe. Thanks

GTNH-Afx237v7:
Can you attach the screenshots you showed in Discord?
The diamond chainsaw is from Electro Magic Tools. Recipe is changed via Minetweaker script. Requires an MV electric chainsaw, seemingly needs to be made out of titanium with thaumium chainsaw tip. Running this chainsaw through a name remover didn’t help either. I’m not too experienced with MT scripts, maybe the issue lies in the definition of required metadata of the electric chainsaw.

ShinoGFX:
![A644806D-C342-4839-A3C6-64C7AE5467F1](https://user-images.githubusercontent.com/48395343/91722952-a8d2da80-eb60-11ea-8242-bccbbd68733d.png)
![Uploading BC9EAEB1-32BD-4707-80DC-960743431C71.png…]()

ShinoGFX:
The wand is in there just for some reason isn’t showing. I’ve had that visual glitch for a while.

Prometheus0000:
This is the script:
// --- Diamond Chain Saw
mods.thaumcraft.Research.removeResearch("Diamond Chainsaw");
mods.thaumcraft.Research.addResearch("DiamondChainsaw", "EMT", "lucrum 12, ignis 9, metallum 6, terra 3", 0, -2 as int, 2, <EMT:DiamondChainsaw:27>.withTag({shearsMode: 1 as byte}));
mods.thaumcraft.Research.addPrereq("DiamondChainsaw", "ElectricMagicTools", false);
mods.thaumcraft.Research.setRound("DiamondChainsaw", true);
game.setLocalization("tc.research_name.DiamondChainsaw", "Diamond Chainsaw");
game.setLocalization("tc.research_text.DiamondChainsaw", "Harder, better, faster, stronger");
mods.thaumcraft.Research.addPage("DiamondChainsaw", "tc.research_page.DiamondChainsaw");
[...]

Issue #6474, opened 2020-08-02
Arcane Worktable and Infusion Altar do not accept the same tier circuit


GlodBlock:
#### Which modpack version are you using?

# 2090
#### What did you try to do, and what did you expect to happen?

The Arcane Worktable and Infusion Altar dont accept the same trie circuit,only the one showed in nei is workable.
#### What do you suggest instead/what changes do you propose?
Could you make all the same trie circuit acceptable.

Dream-Master:
ore dict us not possible with minetweaker. So I need to add 3 more recipe each recipe have circuits.

GlodBlock:
@Dream-Master 
i think you can just add recipe for every circuit to make the mainframe circuit of the same tire

Dream-Master:
Sure that is possible but make the script and loading time longer

Prometheus0000:
Have you decide we should do that, or not do that then? Also WTF is a trie circuit?

GTNH-Afx237v7:
Same tier, aka oredict for the circuit. Should not be too difficult to change in the MT script?

Prometheus0000:
We have to decide whether the extra loading time is worth it. And there's probably other recipes that would need the same thing. So it could add up.

I wish people would make use of the spellcheck at least.

GTNH-Afx237v7:
I wish it was stated whether this affects a certain recipe, or all, or a certain addon to TC...
For sure you can use oredict when doing arcane crafting terminal (see #6562 ). 
@GlodBlock Any more details?

GlodBlock:
All of recipes of TC worktable and infusion altar have the same PROBLEM.
Oredict doesn't work for them

Glease:
Dream's impression seems to be based on ancient versions we don't use any more. We can indeed use oredict in both of the recipes. 

Arcane workbench recipes using oredict: https://github.com/CyricV/TelluraPack/blob/79c946bef68bd90222d594fd6132682556afa842/scripts/ender_io.zs#L386-L388

All ingredients of Infusion recipes are implicitly oredict aware and will try to use oredict wherever possible. Just pass in any item with desired oredict name will work.

GlodBlock:
Well,it's good news for us.

Prometheus0000:
The electric goggles of revealing even has this done already. I'll go through them.

Prometheus0000:
Were there ones not in EMT or TE? I replaced all of those, but I want to do it all at once.

[...]

PR #6713, opened 2020-09-27
Add another IV-circuits for Thaumium Drill

MotiasF33:
maybe better avoid using things like circuits in the infusion?


Prometheus0000:
This is about oredicting circuits, right? Please look at #6474. This pr is completely useless. And remember to use search next time.

GTNH-Afx237v7 (changes requested):
I can take a look at this item however it is highly likely that this won’t work - detailing the ingredients borks because of different IDs. See also #6562.

Dream-Master:
yea something's wrong. I close this now. Feel free to make a new pr

GTNH-Afx237v7:
As I just learned it seems that Minetweaker does not accept oredict for infusion recipes...

Issue #6708, opened 2020-09-27
Dragon egg volumetric flask task


Florexiz:
#### Which modpack version are you using?
2.0.9.0
#
#### If in multiplayer; On which server does this happen? 
Eta
#
#### What did you try to do, and what did you expect to happen?
Quest require 1000mb volumetric flask with binnie.bacteria but its not counted
![image](https://user-images.githubusercontent.com/29033773/94374855-e9018a80-0117-11eb-8f01-f0c976c51fd2.png)
![image](https://user-images.githubusercontent.com/29033773/94374887-282fdb80-0118-11eb-93a4-29e09e724303.png)

(Quest was completed with staff member help)

PR #6718, opened 2020-09-29
fix One Dragon Egg.. Quest 

MotiasF33:
issue https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6708


MotiasF33:
rejected

PR #6736, opened 2020-10-02
minig pipe - conversely

MotiasF33:
before pr
изображение

### DIFF
--- a/scripts/IC2.zs
+++ b/scripts/IC2.zs
@@ -759,8 +759,10 @@ recipes.remove(<IC2:itemPartCFPowder>);
 
 
 
-// --- Mining Pipe Tip -> Mining Pipe
+// --- Mining Pipe Tip <-> Mining Pipe
 recipes.addShapeless(<IC2:blockMiningPipe>, [<IC2:blockMiningTip>]);
+//--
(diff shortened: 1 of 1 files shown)


repo-alt:
What is it for?

MotiasF33:
> 
> 
> What is it for?

for crafting: tip to regular mining pipe

repo-alt:
Indeed, and what would you need that for?

MotiasF33:
> 
> 
> Indeed, and what would you need that for?

After drilling rig working remain pipe - last from it is TIP , recipe is needed to continue using the used pipes

repo-alt:
Ah, I see, you posted the picture of the current state, which is of course wrong.

repo-alt (approved):

Dream-Master:
Please add the old recipe back so we have a conversion recipe for both pipes @MotiasF33

Dream-Master (approved):

Prometheus0000:
@MotiasF33 See https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/6790
Your PR caused the other one to not work. And you can't place down the mining pipe tip either?

> After drilling rig working remain pipe - last from it is TIP , recipe is needed to continue using the used pipes

How? They keep running anyway?

PR #6772, opened 2020-10-07
Wg transmutation fix

Prometheus0000:
For some completely inexplicable reason, OreDict (or is it material name?) tags that are more than one word don't work unless you put a space in, like 'Infinity Catalyst'. All the ones that didn't work had two 'words', like 'DeepIron' or 'Uranium235'. I have no clue why this fixes it, but it does, so yay. Keep this in mind if anything else is added to the list in the future.


Dream-Master (approved):

Rakdarian:
*oh no*

Dream-Master:
oh yes

mitchej123:
How about we balance this instead of just disabling them... :)

Prometheus0000:
This just fixes the _already existing_ blacklist. Deciding what goes on it belongs in an actual ticket.

Dream-Master:
yea for new addition material we can change the amount or so

Rakdarian:
So whats the blacklist now?

Prometheus0000:
The same? Just that it works. I mean seriously, just read what I wrote already.

PR #6816, opened 2020-10-13
Fix(remove) Mining Pipe

MotiasF33:
@Prometheus0000, recipe is already being managed >
https://github.com/GTNewHorizons/NewHorizonsCoreMod/blame/master/src/main/java/com/dreammaster/gthandler/GT_CraftingRecipeLoader.java#L74
i think need remove this recipe from MT

### DIFF
--- a/scripts/IC2.zs
+++ b/scripts/IC2.zs
@@ -759,9 +759,6 @@ recipes.remove(<IC2:itemPartCFPowder>);
 
 
 
-// --- Mining Pipe Tip -> Mining Pipe
-recipes.addShapeless(<IC2:blockMiningTip>, [<IC2:blockMiningPipe>]);
-
 // --- Electric Heat Generator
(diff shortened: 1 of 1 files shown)


Prometheus0000:
Huh, I didn't know you could view a file's history like that.

I tried commenting out the MT script, and there's no recipe, probably because the name of the second part is wrong
image
image

miningPipeTip - > miningTip

MotiasF33:
> second part is wrong

idk..
i'm change it to:
`GT_ModHandler.addShapelessCraftingRecipe(GT_ModHandler.getIC2Item("miningPipe", 1L), bits, new Object[]{GT_ModHandler.getIC2Item("miningTip", 1L)});`
and remove removing in mt but non results

MotiasF33:
can be remove code from gt 
and keep it in mt

PR #6846, opened 2020-10-23
Nerf warded glass

Prometheus0000:
.reuse() is not compatible with the arcane table, so I had it give out a stack, since it consumes the ward focus.


Dream-Master:
or we add a recipe via code

Prometheus0000:
It's a TC item, I think that would require ASM/mixins.

Glease:
See GTNewHorizons/NewHorizonsCoreMod#126. That PR will basically grant player a new warding focus when a craft is finished, so in MT you don't need to add `reuse` or add any special markings. Just add a normal arcane recipe as if you want the focus to be consumed.
```
mods.thaumcraft.Arcane.addShaped("WARDEDARCANA", <Thaumcraft:blockCosmeticOpaque:2> * 4, "aer 10, aqua 10, ignis 10, ordo 10, perditio 10, terra 10", [
[<ore:blockGlass>, <ore:plateThaumium>, <ore:blockGlass>],
[<Thaumcraft:blockWoodenDevice:6>, <Thaumcraft:FocusWarding>, <Thaumcraft:blockWoodenDevice:6>],
[<ore:blockGlass>, <ore:plateThaumium>, <ore:blockGlass>]]);
```
Something like this will suffice.

Dream-Master (approved):

Issue #6858, opened 2020-10-26
rubber boots repair


Eason134HK:
#### Which modpack version are you using?
2.0.9.0
#
#### If in multiplayer; On which server does this happen? 
delta
#
#### What did you try to do, and what did you expect to happen?
I repaired my rubber boots and the enchantment on my rubber boots was gone
#
#### What happened instead? (Attach screenshots if needed)
The enchantment on my rubber boots was gone!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
![image](https://user-images.githubusercontent.com/59776356/97152516-fd36b700-17ab-11eb-88eb-52c7148e7a48.png)
This is a screenshot after I enchanted my rubber boots last night
#
#### What do you suggest instead/what changes do you propose?
change the way of repairing hazmat suit and rubber boots to using anvil instead of crafting
also I hope to get compensation for my boots enchantment if possible

Prometheus0000:
It's an IC2 item, expect that to be super hard or at least totally not worth it.

Eason134HK:
> It's an IC2 item, expect that to be super hard or at least totally not worth it.

i didn't say and thing about i want it to be super hard i just want haste + feather falling and protection

Prometheus0000:
Hard to change the code. With asm/mixins. It's not even comprehensible to me.

Eason134HK:
> Hard to change the code. With asm/mixins. It's not even comprehensible to me.

ok but will i get compensation?

Prometheus0000:
Ask someone who can do that on discord. (not me). And I doubt it.

Dream-Master:
the minetweaker recipe was made for rubber boots without enchant. I huess there is no way to add it via mt the way it keep enchantments.

Glease:
Actually there is. See #6860.

Eason134HK:
> Actually there is. See #6860.

![image](https://user-images.githubusercontent.com/59776356/97265106-09c91700-1861-11eb-973d-4fb64b84662b.png)
i want to soulbound my wireless terminal will the NBT tag lose in the future upgrade?
![image](https://user-images.githubusercontent.com/59776356/97265312-69272700-1861-11eb-89e7-d75432d14166.png)

PR #6860, opened 2020-10-26
Preserve NBT tags when repairing rubber boots

Glease:
#6858


Prometheus0000:
It's cool you can do this, but maybe you should've done the same for the rest of the hazmat suit too?

PR #6892, opened 2020-11-03
add(Forestry) new uranium Tube to script

Dream-Master:


Houstonruss:
This really negatively affects early game power options. Players shouldn't be forced to go to space to use a forestry farm for ic2 crops.

Prometheus0000:
They aren't, get it from thorium.

PR #6935, opened 2020-11-14
add a discord button on the main menu

boubou19:
normal:

hovering the discord button:

discord link:


glowredman (changes requested):
I think the placement looks off, in my opinion one of the corners (preferably at the bottom) would be better

boubou19:
at the same level as the quit message?

glowredman:
For example (ofc with equal distance to both sides)

boubou19:
well, it does not scale well: if you use a little scale of the gui, then you'll have it more next to the button column than the left border, as it's defined by a position offset from the Multiplayer button. Here is an exemple of what i mean:

I changed my gui scale from auto to small to get this. But it seems to be an issue of the base menu, so i don't know if we can really change that. So i need to know if the space between the column and the discord button is good enough, or if i should put it more to the left.

glowredman:
Afaik it's possible to have a corner be 0/0, but the last time I used CMM was some years ago

boubou19:
So you want it full corner? not with some space between the discord button and the borders?

glowredman:
No, not *full* corner but idk 10 or so pixels from both sides. Alternatively not as an icon, but as an entry between language and mods - though ther might be a spce problem on large GUI scales

boubou19:
I put a border of 20 from the bottom left corner, but for the small scales, it seems weird:
small

normal

large

auto

glowredman:
I see what you mean, I think the smaller scales look weird in general though. Maybe you can make the distance larger for the smaller scales? All in all it looks better than in the beginning though. Some other people should probably look at this too.

boubou19:
Same gui scale (normal), just the size of the window changing:
maximized window:

reduced window:


It behaves too weird, so i think i'm gonna set the position of the button as an offset of one of the other buttons. The question is which one? I put the button next to the multiplayer because of the community aspect. But if you think it looks weird, then what other button should i use?

glowredman:
I think the best solution would be to make it a button like all the other, though as your last screenshot shows - there is no space.
I'd base it off of the "Quit" button, because the Discord butoon is not at some random height then.
Or:
If you can scale the icon, I'd make about as alrge as the text and put it left to one of the other buttons (the x coordinate would of course be influenced by the other buttons size). Again the "Quit" button would not neccessaryly the best fitting, but - i think - the best looking.

Prometheus0000:
What if they accidentally hit quit instead? It takes a long time to load in.

glowredman:
That is indeed a problem, but do you have a better idea than making sure there is some space between the two buttons?

Prometheus0000:
Look at some other packs and copy what they do?

boubou19:
I have 0 reference for the other packs, but if you give me some exemples, i can try to replicate it

Prometheus0000:
Maybe later, I'm testing fixing things.

glowredman:
From what I remember, most of the time they either integrate it into the "main" button set or put it in a corner. I'm starting up some packs now, I'll post the examples here

boubou19:
I tried to mess with it on the same level as the "quit" button, but when it looks fine on a scale, if i maximize the window or if i use the reduced window, then it looks weird. I guess we need to integrate it in the button column. 1 small button could look weird in the button column, but we could use the same line to put several buttons: 1 for the discord button, 1 for a github button, and maybe 1 for the official website? I would place them on top of the button column. What do you think?

Prometheus0000:
Small
image
Normal
image
Large
image

Prometheus0000:
Small
image
Normal
image
Large
image

glowredman:
**All normal@1080p**
FTB Infinity Evolved Skyblock
grafik
FTB Ultimate Reloaded
grafik
Minecraft Breakdown 3.2
grafik
Space Astronomy
grafik


@boubou19 Yes I think that's a good idea, though there are some questions then:
- Do 2/3 buttons with text fit in the same row?
- If not, can they (partially) be replaced by icons?
- If yes, does the rest look odd, becuase they *don't* have icons?

glowredman:
Also, I wouldn't put it on top. SP, MP and maybe options should always be the first ones. I think above or below mods would be fine

boubou19:
Which one do you prefer?

glowredman:
I think the top one looks better

boubou19:
then i'll make top one then :D

Prometheus0000:
The top one looks better, but certainly doesn't look as good as the other pack's stuff.

glowredman (approved):

glowredman:
Agree, though i guess getting to that standard would require some serious changes to the main menu. I would appretiate such changes, but I wouldn't invest time myself and I don't expect that from others either

boubou19:
small scale:

normal scale:

large scale:

auto scale:

hovering the gtnh website icon:

hovering the discord icon:

hovering the github icon:


github icon link is : https://github.com/GTNewHorizons/NewHorizons/issues
website link is : https://www.gtnewhorizons.com/
discord link is : https://discord.gg/EXshrPV

Prometheus0000:
I agree, but I was hoping we could just borrow the format from another pack, however they do it.

PR #6944, opened 2020-11-16
Remove added to NH Core mod metal block recipes

Ethryan:
https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/136


Dream-Master (changes requested):
Gt Brass and Tico Brass havent the same ingredients. Please rework it use only Tico brass to avaoid exchange between the two blocks.

Dream-Master (approved):

PR #6981, opened 2020-11-23
Update DefaultQuests.json

Dusttodust514:
I've noticed several inconsistencies in questlines as well as irks that I feel need reworked. Some of these include the inability to do basic math, multiple retrieval steps when no one 
processes materials in that fashion and some quest that I feel are just old....

So here is the start of my rework

tier 0 is mostly fine as well as Tips and Tricks not really much I can do with it and handholding is not the worst idea

Tier 0.5 - Stone Age

Upgrade 2.0 

Reduced cobblestone requirement to 128 players are going to be using cobble for cooked stone, and most ores you can mine with the GT flint pickaxe. With limited resources all this functions
to do is have someone mine mindlessly for 256 stone.

Getting Iron

Reduced required Ingots to 38

Players are going to want to beeline the tools, make the cauldron and get the orefinder rod. 72 ingots is excessive. 38 is the  minimum required for that

Wooden Door

made into a crafting task, multiple retrievals are bad

Divine Rod of the Gods?, iron and gold chest quests
removed component retrieval tasks


Prometheus0000:
Did you not grab the latest version?

Dream-Master:
@Dusttodust514 you cant come here and make decission by your own. Feel free to add a modified version to your pack of GTNH. The main quests in the low tiers are fine and balanced so often times that it is now enough.

dvdmandt:
I agree with the balance itself being mostly fine, but I also agree with the problem of intermediate steps. That has screwed me over so many times in multiple tiers. I've been thinking about introducing optional steps or something to BetterQuesting/SQP so that we could keep the handholding and explanations, but not totally screw people over if they start with step 3 before fully completing step 2 or similar.

Dream-Master:
feel free to contribute to quest file @dvdmandt 
@Dusttodust514 if you realy have some nice addition come back and make a pr

dvdmandt:
Need to implement support for it in the mods before I can start doing something about it in the quests file, and I need to figure out what it would even mean in some cases. .. I think I'll open an RFC issue instead of commenting on this issue though.

PR #7078, opened 2020-12-09
IC2.zs changing tier of electric treetap

Houstonruss:
basically changes materials from stainless steel to regular steel


Dream-Master:
All Ready added

PR #7137, opened 2020-12-20
Update Advanced-Backpacks.zs

Yakameen:
Added recipe to revert Adventure Backpacks back to the original style. Recipe is set for any adventure backpack in the middle surrounded by bonemeal.

Also fixed code so items inside the backpack are not deleted when making bat, squid, and iron golem backpacks like when making other backpacks.

### DIFF
--- a/scripts/Advanced-Backpacks.zs
+++ b/scripts/Advanced-Backpacks.zs
@@ -8,10 +8,7 @@
 import mods.gregtech.Brewery;
 
 
-
-
 // --- Remove Recipes --- 
-
 
 
 // --- Advanced Backpack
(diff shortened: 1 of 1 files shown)


Dream-Master:
@Yakameen are you tested it ?

Yakameen:
Yes, I went in game and put the crafting recipes through a bench. All items that are in the input backpack persist though the output. I did this because I wanted people on my server to be able to swap out backpack styles as they pleased.

This can also be done for the backpack in "Backpack.zs" to prevent items being deleted when upgraded, but that will take more time on my part.

Dream-Master (approved):

miaowwwwww:
image

this recipe is not fixed by this pr. i will write an issue if it is necessary.

Dream-Master:
> image
> 
> this recipe is not fixed by this pr. i will write an issue if it is necessary.

What do you mean? We not using scripts anymore .

miaowwwwww:
> > image
> > this recipe is not fixed by this pr. i will write an issue if it is necessary.
> 
> What do you mean? We not using scripts anymore .

sorry somehow missed this msg. i made a ticket about this.https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19313

Issue #3424, opened 2018-08-08
Thaumcraft mirror nerf in 2.0.4.7


DoomSquirter:
#### Which modpack version are you using?
on 4.6 now
#
#### If in multiplayer; On which server does this happen? 
n/a
#
#### What did you try to do, and what did you expect to happen?
n/a
#
#### What happend instead? (Attach screenshots if needed)
n/a
#
#### What do you suggest instead/what changes do you propose?

so I haven't seen the new recipe cept hear it's gated behind enderium for 4.7.

now if it's the regular mirror you are also now nerfing the magic hand mirror as well.  

if you are nerfing the mirrored glass recipe you are also nerfing all the other mirrors like essentia mirrors.

IMHO, this is a bad decision.  The magic mirrors are a simple item transport, UNIDIRECTIONAL, and speed is an issue tho can be rectified.  

In comparison to conduits/pipes, they lack the ability to tell when the destination is full, so their usefullness is limited w/o complex wireless redstone setups to turn off feed at source if destination is full or near full.

the fact that they show up well before anything else at least gives you a simple transport solution (that you have to work for via thaum).  

if the mirrors are now at the same level as transcievers, then I guarantee you NOBODY will make mirrors except the people wanting any of the downstream things that require it for crafting.  

If you feel MV is too early, then can I interest you in some HV recipe or something.  or better yet, make it a more expensive recipe but keep it at mv.

Dream-Master:
mv is definitely to early. It to stuff interdimensional

0lafe:
The other wireless transfer system is gated to EV too, but costs much more to make, as well as power to run. The mirrors on the other hand do all the functionality of the dimensional transceivers except RF transfer, for a cheaper construction cost, but an increased infrastructure cost. That sounds fairly fair to me. You don't really need much wireless transfer pre moon anyway. 

>if the mirrors are now at the same level as transcievers, then I guarantee you NOBODY will make mirrors except the people wanting any of the downstream things that require it for crafting.

I'd still make them, considering they cost a lot less. Also no dumb soul binding of witches needed.

0lafe:
Maybe switch it to titanium? Slightly earlier, cheaper, but still gates it to the rocket. Imo, giving interdimensional item transport when you start being interdimensional, is a good fit.

Dream-Master:
sure why not

CalterOK:
[...]

PR #7147, opened 2020-12-22
Mirror

Prometheus0000:
They only move essentia around, why are they gated so far? As mentioned in a previous issue 
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/3424#issuecomment-540146060


KiloJoel:
Strongly support this. I had been meaning to suggest this myself

KiloJoel:
I would personally say they would be better a bit earlier even. But any buff is a positive for this.

KiloJoel:
> 
> 
> I would personally say they would be better a bit earlier even. But any buff is a positive for this.

To make a specific suggestion: vibrant alloy instead of titanium

Prometheus0000:
Does anyone _other_ than kilojoel have any comments?

repo-alt:
Can you post a screen of the new recipe? metaitems are, kind of, not very idiomatic

Prometheus0000:
It's just enderium plate to titanium plate, as mentioned in the commit.

repo-alt (approved):

PR #7221, opened 2021-01-02
AddGTRecipesToIC2Machines = flase

Dream-Master:
changes need pull https://github.com/GTNewHorizons/GT5-Unofficial/pull/390


Glease:
No longer needed as this option is now hardcoded to be false.

Issue #7266, opened 2021-01-08
Adventure backpack 'bat'


anoldshoe:
#### Which modpack version are you using?
2.1.0.0
#
#### If in multiplayer; On which server does this happen? 
gtnh.moddedminecraft.club
#
#### What did you try to do, and what did you expect to happen?
Craft adventure backpack: bat (4345:2)
#
#### What happened instead? (Attach screenshots if needed)
crashes client
#

alphaNOVAircraft:
Can you post a Crashlog?

Yakameen:
@anoldshoe Yes, please post the crash log, I just crafted one and could not replicate the crash...

Florexiz:
Same happens on eta with squid backpack (most likely with every backpack)
[crash-2021-01-11_13.25.01-client.txt](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/files/5795146/crash-2021-01-11_13.25.01-client.txt)

Yakameen:
What was in the backpack when the crash happened?

Yakameen:
Fixed with https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/7299
Recommend Closing. Thank you @Florexiz for providing that crash report.

PR #7292, opened 2021-01-10
Update Backpack.zs

Yakameen:
Implemented fix from https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/7137

### DIFF
--- a/scripts/Backpack.zs
+++ b/scripts/Backpack.zs
@@ -28,6 +28,25 @@ val MagentaSmallB = <Backpack:backpack:14>;
 val OrangeSmallB = <Backpack:backpack:15>;
 val WhiteSmallB = <Backpack:backpack:16>;
 
+val SmallPacks = <ore:SmallBackpacks>;
+SmallPacks.add(SmallB);
+SmallPacks.add(BlackSmallB);
+SmallPacks.add(RedSmallB);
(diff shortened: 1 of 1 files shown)


Yakameen:
Actually, please pause this. I want to investigate the crash happening first.  https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7266

Dream-Master:
Sure set it to draft

Yakameen:
This is not affected by the same crash as adventure backpacks were affected by since the inventory is saved in a text file and referenced instead of stored directly in nbt.

Dream-Master:
@Yakameen did you test the whole stuff ?

Yakameen:
Yes, I tested crafting a small backpack, put stuff in it and upgrading it all the way to big also I added dye to it at each step with no issues.
The bags have NBT data that look like below. The code added simply copies the Backpack-UID between crafting recipes. There is no crash if the bag does not have an ID assigned to it. 
`<Backpack:backpack:202>.withTag({"backpack-UID": "325c7a8c-343c-4ccc-b08e-f9b1c5792c95", name: "item.backpack.big_red.name"})`

Ethryan (approved):
Good enough testing for me

PR #7461, opened 2021-02-11
fix autoclave of TC dusts / allow autocrafting of Galacticraft items

GTNH-Afx237v7:
fix molten void autoclave recipe of TC crystal powders only giving air shards instead of their respective shards
Change Arc Lamp and Frequency Module recipes to use GT battery instead of GalactiCraft battery to allow proper autocrafting

### DIFF
--- a/scripts/Galacticraft.zs
+++ b/scripts/Galacticraft.zs
@@ -113,22 +113,22 @@ recipes.remove(<GalacticraftCore:tile.refinery>);
 // --- Fuel Loader
 recipes.remove(<GalacticraftCore:tile.fuelLoader>);
 
-// --- Saleable Oxygen Pipe
+// --- Sealeable Oxygen Pipe
 recipes.remove(<GalacticraftCore:tile.enclosed:1>);
 
-// --- Saleable Copper Cable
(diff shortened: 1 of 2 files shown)


Dream-Master (approved):

Prometheus0000:
1) There's no point fixing typos in comments, they'll all get removed whenever we move to using the script mod
2) What's up with the glowstone torch thing, didn't I fix it already? Did you accidentally use a 3 month old version of this file or something?

GTNH-Afx237v7:
1) it just annoys me
2) no, I copied over from the main repo literally 2 mins before making the PR - did I do sth wrong?

Prometheus0000:
image
^was adjusted 3 months ago (change liquid to null) and again before that

Honestly I dunno what's up with it, but as long as it works, it's fine.

Dream-Master:
> image
> ^was adjusted 3 months ago (change liquid to null) and again before that
> 
> Honestly I dunno what's up with it, but as long as it works, it's fine.

A control circuits is always a good idea

Prometheus0000:
The thing is that I already edited this recipe, including the circuit, but github thinks it's a new change.

Issue #7455, opened 2021-02-10
Broken quicksilver recipe (Thaumic bases)


CatsLeftEar:
#### Which modpack version are you using?
MC1710
#
#### If in multiplayer; On which server does this happen? 
GTNH official ETA
#
#### What did you try to do, and what did you expect to happen?
I tried to use "making quicksilver" from Thaumic bases, there is a recipe to transmute silverwood to quicksilver.
#
#### What happened instead? (Attach screenshots if needed)
No such recipe in NEI and recipe doesnt work either with crucible or alchemical construct.
#
#### What do you suggest instead/what changes do you propose?
Remove this recipe from thaumonomicon (bad idea) or add recipe as it should be (good idea)
![изображение](https://user-images.githubusercontent.com/67397297/107508716-baa1ba80-6ba1-11eb-901e-699e4e94c9e5.png)
![изображение](https://user-images.githubusercontent.com/67397297/107508746-c3928c00-6ba1-11eb-9e38-f7f8a69634c8.png)

GTNH-Afx237v7:
Tested with 2.1.0.0, and it is indeed not working. You can't select this recipe in an alchemical construct.

There is obviously a Thaumonomicon page added

https://github.com/GTNewHorizons/ThaumicBases/blob/98230988b35fff69fff34a39861de5c7fe04361e/src/main/java/tb/init/TBThaumonomicon.java#L550

but I don't see an actual recipe in the code?

https://github.com/GTNewHorizons/ThaumicBases/blob/master/src/main/java/tb/init/TBRecipes.java 

Additionaly, there is no recipe in the ThaumicBases.zs script file.

Maybe this was forgotten at some point?

GTNH-Afx237v7:
Add this to Thaumic-Bases-01-Main.zs
I tested it and it works fine.

```
// --- Quicksilver
Crucible.addRecipe("TB.Quicksilver", <Thaumcraft:ItemResource:3>, <Thaumcraft:blockMagicalLog:1>, "ordo 1, venenum 1");
```

Dream-Master:
why not pack it into the code ?

Prometheus0000:
Not everyone knows how.

GTNH-Afx237v7:
as i just learned my way of doing things on Github is not the best so for now i'll let it sit here...

Prometheus0000:
[...]

PR #7465, opened 2021-02-11
Add script QS recipe

Prometheus0000:
See #7455 for why.


Dream-Master:
Does it fix the problem or not ?

GTNH-Afx237v7:
Yes it does. Adding it to the script file works fine, I tested it with 2.1.0.0.

Dream-Master:
but if you use it for charcoal woks it too?

Dream-Master (approved):

GTNH-Afx237v7:
Yes, just tested and it works fine, too

PR #7495, opened 2021-02-17
Fluorite from calcite recipe

Leche-Caliente:
I was told that if I wanted to make a fix for the current fluorite-from-limestone recipe, I could write the code myself. This therefore is the simplest fix I could cobble together considering that I was figuring out modding for the first time. I don't claim that this recipe makes much sense, but it certainly makes more sense than the limestone recipe that is already in the pack. If someone else wants to put more thought into and code up a different process that makes more sense, I would be happy to see them do it.

### DIFF
--- a/scripts/GT++.zs
+++ b/scripts/GT++.zs
@@ -233,4 +233,4 @@ Assembler.addRecipe(<miscutils:personalCloakingDevice-0.0>, <IC2:itemPartIridium
 
 
 // --- Fluorite Dust ---
-ChemicalReactor.addRecipe(<miscutils:itemDustFluorite>, <gregtech:gt.metaitem.01:2823> * 4, <gregtech:gt.integrated_circuit:2> * 0, 600, 120);
+ChemicalReactor.addRecipe(<miscutils:itemDustFluorite>, <gregtech:gt.metaitem.01:2823> * 4, <gregtech:gt.integrated_circuit:2> * 0, 600);
--- a/scripts/GT++.zs
+++ b/scripts/GT++.zs
@@ -233,4 +233,4 @@ Assembler.addRecipe(<miscutils:personalCloakingDevice-0.0>, <IC2:itemPartIridium
 
 
 // --- Fluorite Dust ---
-ChemicalReactor.addRecipe(<miscutils:itemDustFluorite>, <gregtech:gt.metaitem.01:2823> * 4, <gregtech:gt.integrated_circuit:2> * 0, 600, 120)
+ChemicalReactor.addRecipe(<miscutils:itemDustFluorite>, <gregtech:gt.metaitem.01:2823> * 4, <gregtech:gt.integrated_circuit:2> * 0, 600, 120);
--- a/scripts/GT++.zs
+++ b/scripts/GT++.zs
@@ -7,6 +7,8 @@
 
 import mods.gregtech.Assembler;
 
+import mods.gregtech.ChemicalReactor;
+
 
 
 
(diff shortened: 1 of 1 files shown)


Leche-Caliente:
I had thought that I could link to an issue after posting, but it doesn't seem that I can. This pull request should be linked to the "Limonite to fluorite ore recipe" which already has some discussion.

botn365 (changes requested):
this removes and chalange geting fluorine with already has a lot of ways of geting it 
and defnetly not this dirt cheap witch almost creates a infinit loop of calcite production and would allow for realy easy mass produtcion of fluorine at the cost of only some carbon oxygen and calcium
and to make this would be a lot easyer then the only other way of infinit fluorine production witch is crops 
but crops actual takes many hours of grind to get any good fluorine production and scales pret bad
while this with a few LCR and a few Keu/t can make buckets of fluorine per second

Leche-Caliente:
> this removes and chalange geting fluorine with already has a lot of ways of geting it
> and defnetly not this dirt cheap witch almost creates a infinit loop of calcite production and would allow for realy easy mass produtcion of fluorine at the cost of only some carbon oxygen and calcium
> and to make this would be a lot easyer then the only other way of infinit fluorine production witch is crops
> but crops actual takes many hours of grind to get any good fluorine production and scales pret bad
> while this with a few LCR and a few Keu/t can make buckets of fluorine per second

All I did was add a recipe which does the same thing which you can already do with limestone, except you don't have to break blocks to do it. I even reduced the yield by half over the current recipe.

If the limestone recipe is unbalanced, fine, it should be removed, but if it isn't, this recipe shouldn't be unbalanced either.

botn365:
what you did is make it automatble and realy easly automatble
the only reason it in its curent form isent OP is becous its a lot of work and cant be automated
it not being automatble means it is not a reliable source of flourine and it needign a good amount of work makes it imposible to get large amounts of fluorine from it for mass production
but this recipe takes all what makes this not a OP source of fluorine and trows it out the windom adding this would totaly break any balance to fluorine

Leche-Caliente:
> what you did is make it automatble and realy easly automatble
> the only reason it in its curent form isent OP is becous its a lot of work and cant be automated
> it not being automatble means it is not a reliable source of flourine and it needign a good amount of work makes it imposible to get large amounts of fluorine from it for mass production
> but this recipe takes all what makes this not a OP source of fluorine and trows it out the windom adding this would totaly break any balance to fluorine

The current recipe is balanced around tedium, which is a poor way to balance things. If you think that this recipe is unbalanced, than you should help me make a new recipe to replace the old one, which is balanced without being tedious. I made it very clear that I was welcome to suggestions for a better process, but that my modding knowledge isn't up to making anything more involved myself.

botn365:
no its not balanced around tedium its balanced around to be a quick boost of fluorine not to be a main source of it there are other options for that
and there a;ready is a option to make fluorite ore in the chem plant from limestone/sandstone so no this will just break balance completly

Leche-Caliente:
> no its not balanced around tedium its balanced around to be a quick boost of fluorine not to be a main source of it there are other options for that
> and there a;ready is a option to make fluorite ore in the chem plant from limestone/sandstone so no this will just break balance completly

No, it is supposed to be a quick books of fluorine, and it is balanced by tedium in order to get it to that point. This is obvious, because if it was not tedious --as I made it in my change-- it becomes unbalanced, as you have clearly expressed.

The chem plant is not HV, which is the context of this issue.

Prometheus0000:
> The chem plant is not HV, which is the context of this issue.

It's the TS casing requirement that is not HV.

Leche-Caliente:
> > The chem plant is not HV, which is the context of this issue.
> 
> It's the TS casing requirement that is not HV.

True, a bit pedantic, I think it's clear what I meant.

Prometheus0000:
It wasn't clear to me, which is why I went and checked what you were talking about.

PR #7552, opened 2021-03-02
CraftPresence GUI key is changed to None

### LINKED ISSUE #7351: The craftpresence mod defaults to `/~, messing up the QB shortcut
#### Which modpack version are you using?
2.1.02c
#
#### What did you try to do, and what did you expect to happen?
Use the QB
#
#### What happened instead? (Attach screenshots if needed)
Get the stupid CP thing
#
#### What do you suggest instead/what changes do you propose?
Remove CP. But seriously, make the default _some other key_. Or no key. And also add the ability to config it _from the key config menu_ if possible.
@jackowski626

jackowski626:
Pressing escape when prompted to enter a keybind has set it to "none". "none" keycode is 0.
Fixes #7351


Dream-Master (approved):

yukieiji:
I checked with v2.1.0.7a, and it was not set to None even with 0, and '@' was assigned in the 108 key layout(Japanese).

Prometheus0000:
Even if you set it to none, it activates with special keyboard keys, like the calculator button. Apparently this mod is not alone in that kind of thing, and I don't know if jack could fix that. Try setting it to something unused instead, or removing the mod.

yukieiji:
It may not be the effective way, but how about assigning a large number (a number that exceeds the number of keys, e.g. 200)?
I tried it and it didn't log errors, warnings, crash, etc.

Prometheus0000:
Did you do that through the cfg rather than the GUI?

yukieiji:
Yes, Loaded by "Sync Config" after configuring in the cfg.
KeyBindCraftpresence

Prometheus0000:
@jackowski626

yukieiji:
I've tried putting in various appropriate values.
200->OK
255->OK
256->replace 200
300->replace 41
2000->replace 41
1000->replace 41

Glease:
GT++ has specific patches in place to prevent a erroneous key binding from crashing the game with NPE. It might be at play here. Try load the pack with only craft presence and see how it goes.

Prometheus0000:
It sounds like it has a 0-255 limit. I'd make a PR, but I don't know if it would cause a problem with some keyboards or regional settings, so I'm going to leave that up to jack.

yukieiji:
I've tested it in a CraftPresence only environment, and it seems to be fine.
KeybindCraftpresence_test

PR #7590, opened 2021-03-09
Quest changes

Prometheus0000:
BTW, it seems every time I add or remove a quest it's readjusting all the quest IDs and/or their positioning. Until we get back to the amount before deleting some. This causes big diffs that are in reality, minor. Maybe we should change that system?


Dream-Master:
I will dig through your pr today

config/betterquesting/DefaultQuests.json
```diff
@@ -8560,8 +8560,12 @@
           "consume:1": 0,
           "requiredItems:9": {
             "0:10": {
-              "id:8": "gregtech:gt.blockgem3:4",
+              "id:8": "betterquesting:placeholder",
```
Dream-Master: placeholder ? looks like a bug

Dream-Master (approved):

PR #7630, opened 2021-03-14
Add TC oreberries to Pulverizer recipes

rv-jgentile:
@Dream-Master I added the recipes locally because I had a backlog of oreberries I wanted to macerate, so I figured I might as well put up a PR. I wasn't sure what the old duration/voltage was but as I write this sentence I realize that I was just lazy and can fire up an older version to check them out, so I'll circle back after doing that.

* All recipes are 15s @ 600EU
* Aluminum oreberries crush to tiny aluminium piles.
* Used refs from [tconstruct in scripts/Tinkers-Construct.zs](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/blob/ba4f01a32df60ec05d5c11168f4174aee231da09/scripts/Tinkers-Construct.zs#L14) and [Gregtech.zs](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/blob/4c9222ffa366fb3dd78995af11c97fd12479a9b2/scripts/Gregtech.zs#L504)
* Some formatting fixes thanks to the ZenScript IDE Extension I grabbed.
* First time touching ZenScript so don't go easy on me.
  
    
 _Edit: Double checked in 2.0.0.0, recipes were indeed 15s @ 600EU._

### DIFF
--- a/config/GTNewHorizons/dreamcraft.cfg
+++ b/config/GTNewHorizons/dreamcraft.cfg
@@ -165,8 +165,8 @@ modules {
     # Set to true to show login message with modpack version [default: true]
     B:LoginMessage=true
 
-    # Version of the Modpack [default: 2.1.0.4]
-    S:ModPackVersion=2.1.0.4
+    # Version of the Modpack [default: 2.1.0.5]
+    S:ModPackVersion=2.1.0.5
 
     # Set to false to prevent the OreDict register for SpaceStones and SpaceDusts [default: true]
     B:OreDictItems=true
--- a/scripts/Gregtech.zs
+++ b/scripts/Gregtech.zs
@@ -2,7 +2,7 @@
 
 
 
-// --- Imports --- 
+// --- Imports ---
 
 import minetweaker.item.IIngredient;
 import minetweaker.item.IItemStack;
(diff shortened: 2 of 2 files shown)


Prometheus0000:
They're berries rather than nuggets though. Also, I'm not sure why you changed the version of the pack? Or was that dream?

rv-jgentile:
> 
> 
> They're berries rather than nuggets though. Also, I'm not sure why you changed the version of the pack? Or was that dream?

The version bump was dream @ https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/commit/eb0000f06ab465a526a9238c3ddd04f6715ad97e - not sure how it ended up in this PR though. 
  
Totally right about nuggets vs. berries, I'll update the comments.

yukieiji:
Since v2.1.0.0 does not have these recipes within ZenScript, I assume they were added by some code. Therefore, I think they should be added in the code....

Dream-Master:
maybe same probem like the other Tico compat. I would like to see this changes/additions in code not in script.

rv-jgentile:
Cool, good call & no worries - I'll close this. Like I said just something I did to get it working locally and figured I might as well throw up a PR for.

Dream-Master:
@rv-jgentile whould be nice do a code pr to my core mod.

rv-jgentile:
@Dream-Master I promise I'll poke around when I can and see if I get lucky (I'm definitely curious) but no promises on a PR with a fix. Family and full time job to deal with, I was only able to whip this PR up during naptime. Typing this sentence from a phone with a 1mo old baby in my arm, lol. Kids definitely have an adverse effect on hobbies!

Prometheus0000:
https://github.com/GTNewHorizons/NewHorizonsCoreMod/blob/master/src/main/java/com/dreammaster/gthandler/GT_MachineRecipeLoader.java is the file where machine recipes go in that mod. Generally if it's a cross-mod interaction and not in a script, it's in the coremod. There's another in the same folder for normal 3x3 recipes too.

PR #7715, opened 2021-03-27
Fix typo in quest "Personally, I prefer Pepsi™"

al-tr:


Dream-Master (approved):

Prometheus0000:
Same as previous, why is this in the lang file instead of the questbook?

Issue #7725, opened 2021-03-29
Thaumic Bases Overchanting Table cannot be created due to incorrect specifications.


anabsolutesloth:
#### Which modpack version are you using?

#2.1.0.5
#### If in multiplayer; On which server does this happen? 

#Private server running 2.1.0.5
#### What did you try to do, and what did you expect to happen?

#Create the Overchanting Table from Thaumic Bases.
#### What happened instead? (Attach screenshots if needed)

#The craft started, but could not complete due to incorrect specifications
#### What do you suggest instead/what changes do you propose?

#in the Thaumic-Bases-01-Main script, Perditio and Praecantatio essentia are specified, but no amount is given.
"aer 32, aqua 32, terra 32, ignis 32, ordo 32, perditio, praecantatio, machina 32, cognito 32"
#### What is your GTNH Discord username?
Emperdog

PR #7726, opened 2021-03-29
Fix overchanter recipe

Prometheus0000:
See #7725


Dream-Master (approved):

Dream-Master:
Bad overseen. Thanks for fix @Prometheus0000

anabsolutesloth:
One thing, "cognito" should be "cognitio", as that is also incorrect.

Issue #7090, opened 2020-12-11
remove pss recipes


botn365:
the pss stores lots of energy very cheap wich wich allows people to mostly skip power gen in EV-LUV
sisne the there is now the LSC wich while stil has problems but is more balanced sisne it is way more expansif then the pss and equaly or more expansif then battery buffers
so the pss should be disabled by removing the abilety to make the controller and redox cells

alphaNOVAircraft:
The lsc on the other hand doesnt really feel rewarding at all to use early on, you gain zero capacity compared to usesing the orbs in a normal buffer, so basically all you do is accept some loss (its negligible, i know) and the ability to have infinit amp io. 
 
The pss in current state is bs, definitly agreeing with that one. With quite a lot of completly passive eu gen available you can simply chunkload it and you have nearly infinit energy to use while your playing, letting it charge up again when your doing something else. Your energyproduction can be less then an amp of your current tier but you can still dont really run out of energy while playing. Setting up an adequate energy production isnt required at all. 
If you need to remove it, not sure about that one, but since we cant change the storage capacity of it its maybe the only way. 

If you diable it i would like to see some small buffs to the lsc, making it rewarding to use it, maybe giving the energy capacity a muliplier.

alphaNOVAircraft:
Especially around zpm-uv batterys are even way cheaper then the equivalent orbs, so currently its fine in iv / luv, sucks at zpm / uv and is op af at uhv / uev once you got your really ultimate. 
So its not really a perfectly balanced thing too. But surley ensures you pay something for storage which a2 fails to do

Prometheus0000:
Make another issue to suggest changing the recipes of the lapo series and/or normal batteries and/or capacitors?

0lafe:
Don't make them easier, it's definitely fine atm. We did no LSC/pss and it's really fine. Good incentives for power gen. Removing it seems good. Making the LSC store more or have cheaper recipes seems bad. Hatch/dynamo based power storage is fucking amazing, giving it bonus storage too is just too much. 

Good idea bot, let's do it

Prometheus0000:
> We did no LSC/pss and it's really fine

Not everyone like the idea of removing all power storage 0lafe, 0lafeDusk when?

> Making the LSC store more or have cheaper recipes seems bad

[...]

PR #7802, opened 2021-04-11
Remove PSS controller recipe

Prometheus0000:
See #7090.

I tried removing the LuV+ redox cell recipes, but there's no way to remove assembler recipes with MT. 😢 

It should probably wait until after 2.1.0.6 at least.


Dream-Master:
There is. Still look into gt config. Set prcessing time of the recipe to 0

Prometheus0000:
I'm not sure what you mean? Recipes aren't defined in a config file? Not to mention this is a GT++ recipe, not GT?

Do you mean overriding a recipe with another one that's the same with 0 time in the script file?

Dream-Master (approved):

johnch18:
Based

Alexdoru:
Removed the quests related to the PSS ?

GTNH-Colen:
This is kinda ridiculous, a thread was made which was clearly in favour of NOT doing this so your response is an attempt at a stealth nerf? Very unprofessional.

GTNH-Colen:
@Dream-Master Please undo this PR until further discussion can be had on the matter, it is not resolved.

Prometheus0000:
BTW, I'm fine with it being reverted. I thought we'd be discussing it for _days_ at least.

0lafe:
This seems like a good change. LSC is more than enough, and we can't control the PSS

botn365:
gt++ is on its way out anyway so i dont see the problem by removing it now

iouter:
because player need a mod can be instead of gt++, not remove recipes only,

0lafe:
the LSC does that for power storage which is all that's being removed right now tho

iouter:
LSC is tooooooooooooooooooooooooooo expensive, so it is not suitable for IV-UV use.

0lafe:
it's more expensive than the PSS for sure but it's very cost balanced. The storage is costly but the throughput is cheap, which is very fair

GTNH-Colen:
The merits for or against the PSS are irrelevant, merging this without finishing out prior discussion in another thread is wrong and tantamount to a stealth nerf (esp when the vote was going heavily against doing it). The matter has been discussed internally by staff and agreed that it should be reversed for the time being.

iouter:
Make a Lapotronic Capacitor(LUV) need :

1x Multilayer Fiber-Reinforced Circuit Board
32x Europium Foil
4x LuV-tier Circuit
36x2x Engraved Lapotron Chip
64x High Power IC
32x Advanced SMD Diode
32x Advanced SMD Capacitor
32x Advanced SMD Resistor
32x Advanced SMD Transistor
64x Fine Platinum Wire
4x Osmiridum Frame Box
24x Osmiridium Screw
L1/720x Molten Soldering Alloy

i don't think this is more fair

core-exe:
It does not matter if it is cheap or not. LSC is not the substitude of PSS in IV stage. Many using PSS is because EV-LuV battery is way too expensive from it should be.

GTNH-Colen:
Please direct further comments about PSS/LSC balance to #7090 or I will lock the thread, thanks.

0lafe:
it's really not tho? as someone who didn't use a PSS or an LSC, you get very good storage from the batteries for the cost. you really don't need much storage at all in those stages, and the LSC multi really is just meant for centralization, which it excels in

Issue #7839, opened 2021-04-21
[Suggestion / RFC] Change ME Storage Components beeing extensive recursive crafting


alphaNOVAircraft:
#### What do you suggest instead/what changes do you propose?
Currently ME Storage compnents dont bring any benefit when beeing upgraded in terms of capacity (they even get more expensive but not more capacity). Furthermore the high tier cells take insane amounts of cheap compnents that just slow down crafting (1 16M drive is 64 000 ulv circuits e.g.)) 
Maybe add direct recipes for storage cells higher then 1k that get aviable later so that a alternative to the currently recursive crafting only exits
#
#### What is your GTNH Discord username?
alphaNOVA#9875

Prometheus0000:
There was already another issue on this. You couldn't go post there instead? We reopen stuff you know. Not that I could find it either, lol. I think it was colen's issue?

alphaNOVAircraft:
Mitch suggested i open an issue so i guessed there wassnt one, and reopening issues will get lost again, most ppl dont bother to check page 238 of issues..

GTNH-Colen:
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/5040 I think this was the other thread? Not of my creation though.

GTNH-Afx237v7:
I'm also all for reducing the nested crafting - the resource cost for larger AE cells are much higher than for a comparably sized super/quantum chest.

Glease:
The prohibitively high cost of ME storage cells is one of the several major reasons people use solid drawer cubes. Making ME storage cells easier, if not cheaper, to get would solve the lag from these stupid constructs.

Prometheus0000:
So what are the recipe suggestions? I don't see any? Which ones should have new recipes? 1 new one for 256k and that's it? All of them past 1k? You need to be clearer about _how_ you want them changed.

GTNH-Afx237v7:
So currently the base for all components is the 1K ME Storage component, consisting of 

- 4 ULV Circuits
- 4 Charged Certus Quartz Dust
- 1 Item Processor Tier 1

As components tier up, the required processors tier up, up and including UV-Tier circuits for the 16384K Storage component.

How about each tier of storage consist of

- 4 tiered circuits 
- 4 plates of appropriate tier
- 1 item processor

without requiring the previous tier, i.e. no nested crafting?

So a 1K storage component could remain as is, whereas e.g. the 4096K would be

[...]

PR #7847, opened 2021-04-22
Change AE2 storage component recipes

Prometheus0000:
Makes them not recursive. See #7839.

### DIFF
--- a/scripts/Extra-Cells.zs
+++ b/scripts/Extra-Cells.zs
@@ -57,7 +57,7 @@ val DataCircuit = <ore:circuitData>;
 val EliteCircuit = <ore:circuitElite>;
 val MasterCircuit = <ore:circuitMaster>;
 val UltimateCircuit = <ore:circuitUltimate>;
-val SuperconductorCircuits = <ore:circuitSuperconductor>;
+val SuperconductorCircuit = <ore:circuitSuperconductor>;
 val InfinityCircuits = <ore:circuitInfinite>;
 val EFlow = <gregtech:gt.metaitem.01:32706>;
 val DataOrb = <gregtech:gt.metaitem.01:32707>;
@@ -66,6 +66,7 @@ val QuartziteScrew = <ore:screwQuartzite>;
 val CertusScrew = <ore:screwCertusQuartz>;
 val NQuartzScrew = <ore:screwNetherQuartz>;
 val CertusPlate = <gregtech:gt.metaitem.01:17516>;
+val CCDust = <dreamcraft:item.ChargedCertusQuartzDust>;
 val AlPlate = <ore:plateAluminium>;
 val TitaniumPlate = <ore:plateTitanium>;
 val StainlessPlate = <ore:plateStainlessSteel>;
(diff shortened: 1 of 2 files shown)


Dream-Master:
They ass cheap now. Replace a component with a plate ?

Glease:
@Dream-Master they need to be cheaper than drawers. they are more tps friendly and is more pleasing to look at than a huge solid drawer cube. 

currently the only reason to use drawers at that scale is because the alternative is too expensive.

Dream-Master:
And if you change it please add a control circuit to the recipe

Prometheus0000:
See the discussion on the issue and in #github-discussion. And the previous discussion and PR.

Prometheus0000:
I knew I forgot something.

Dream-Master:
yeah but plates phew? Maybe we can think of a better recipe component for a storage then a plate ?

Prometheus0000:
Wouldn't the data be stored on the processor+chips? With the plates keeping it together and the crystal as some kind of medium?

Dream-Master:
compared to the old recipe we set recipe from 100% Material cost to 1%

Prometheus0000:
Is that bad? People weren't using them before.

Dream-Master:
Even vanilla recipe more expensive
https://ftbwiki.org/4k_ME_Storage_Component

GTNH-Afx237v7:
"Set recipe costs from 100% to 1%" ->Yes because the old recipe was insane. Compared to super/quantum chests the AE components were just ridicuoulusly expensive; and with the non-ticking TE handler that repolainen implemented, the AE disks became even less attractive

Prometheus0000:
It's still more expensive, as we have the intermediary item processor item. And the whole idea is to get away from recursive crafting.

Prometheus0000:
We also have the very expensive circuits in ours, whereas vanilla does not.

GTNH-Afx237v7:
then i suggest we take the discussion back to the RFC thread; I'll post what botn365 mentioned in the discord so we have a different perspective as well

Prometheus0000:
Why though? Just make the current tier chip more expensive?

Prometheus0000:
Also, what do you mean by 'chip', there's 3 different things it could be?

Dream-Master:
Why more expensive? if you se only one instead of 4 from previous tier? Or we can change the recipes of the ae ships. this recipes are very old and outdated too.

Prometheus0000:
What? You're the one that wanted to make the recipes more expensive? I'm saying make the item processor more expensive instead of using the previous tier of item processor. Which I already said on discord. Did you read the discussion?

Dream-Master:
You made the recipe very much more cheaper and i want to take care make the recipe more balanced. With the changes i will probably doing we not 99% cheaper but maybe 80%.

Prometheus0000:
And I'm asking why you want to add 
image
A previous tier of these in, when you can simply change their recipe?
image
Like add in some circuits or something.

Personally, I till think they're expensive enough, but I dunno.

Dream-Master:
This recipe is very old. We could add a solder and a control circuit. Or /and add some fine wire or some things make the recipe looks a bit nicer.I dont speak about changes you did to components i speak about minimal changes and adjustments :)

scripts/Extra-Cells.zs
```diff
@@ -619,16 +619,16 @@ Assembler.addRecipe(<extracells:craftingstorage:2>, <appliedenergistics2:tile.Bl
 Assembler.addRecipe(<extracells:craftingstorage:3>, <appliedenergistics2:tile.BlockCraftingUnit>, <extracells:storage.component:3>, 400, 1920);
 
 // --- Storage Cell Component - 256K
-Assembler.addRecipe(StorageComponent256K, [DataCircuit * 4, TitaniumPlate * 4, <dreamcraft:item.EngineeringProcessorItemEmeraldCore>, <appliedenergistics2:item.ItemMultiMaterial:8> * 4], null, 200, 1920);
+Assembler.addRecipe(StorageComponent256K, [DataCircuit * 4, TitaniumPlate * 4, <dreamcraft:item.EngineeringProcessorItemEmeraldCore>, <appliedenergistics2:item.ItemMultiMaterial:8> * 4, <gregtech:gt.integrated_circuit:9> * 0], null, 200, 1980);
```
Dream-Master, line 623: 1920 was ok

Dream-Master (changes requested):

Prometheus0000:
Forgot to update, they now look like https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7839#issuecomment-826104395

Dream-Master:
lets use a 3:1 ratio not ix it up. And yes plates are to cheap

PR #7902, opened 2021-05-03
A bunch of editing for (mostly) LV

Prometheus0000:
Also layout adjustment


Dream-Master (approved):
cant check it here will do it in game

Issue #8036, opened 2021-06-10
Duplicated Receipe for Clay dust causing Clay block to give 9 dust instead of 2


tommy307507:
#### Which modpack version are you using?
2.1.0.7a
#
#### If in multiplayer; On which server does this happen? 
Stone Legion
#
#### What did you try to do, and what did you expect to happen?
4 clay -> clay block -> 2 clay dust
matter is conserved and the universe is happy
#
#### What happened instead? (Attach screenshots if needed)
4 clay -> clay block -> 9 clay dust
![image](https://user-images.githubusercontent.com/49487505/121555336-805c0300-ca45-11eb-89ae-b3aa9b6e1893.png)
![image](https://user-images.githubusercontent.com/49487505/121555397-8eaa1f00-ca45-11eb-8f0f-4ffe43e3a66b.png)

matter is not conserved and the universe is not happy
#
#### What do you suggest instead/what changes do you propose?
remove the extra receipe... dunno how tf it got there 
#
#### What is your GTNH Discord username?
Hebi

tommy307507:
![image](https://user-images.githubusercontent.com/49487505/121557339-41c74800-ca47-11eb-8b2f-7197ba7c06c3.png)

Also this reverse receipe exist... someone told me its some oredict stuff messing up... not sure if this is been reported before but no one I know talked about it

Prometheus0000:
Looks like this is another autogenerated one that assumes block=9 items.

Prometheus0000:
Same problem as #7956

Prometheus0000:
The maceration part has to do with the reverse recipe line
```
RA.addPulveriserRecipe(aStack, new ItemStack[]{GT_OreDictUnificator.getDust(aData.mMaterial), GT_OreDictUnificator.getDust(aData.getByProduct(0)), GT_OreDictUnificator.getDust(aData.getByProduct(1)), GT_OreDictUnificator.getDust(aData.getByProduct(2))}, null, aData.mMaterial.mMaterial==Materials.Marble ? 1 : (int) Math.max(16, tAmount / M), 4, tHide);
```
in the GT_RecipeRegistrator class, but I couldn't figure out how to change it to fix it.

Issue #7956, opened 2021-05-21
Nether quartz block recipe issue


Quadrocake:
#### Which modpack version are you using?
2.1.0.7a
#
To create a qurtz block you need to compress 4 nether quartz:
![image](https://user-images.githubusercontent.com/46348077/119205667-893e6200-baa1-11eb-8089-01c86d1f0b73.png)
And then you can convert this block into 9 quartz:
![image](https://user-images.githubusercontent.com/46348077/119205692-a07d4f80-baa1-11eb-8e07-276f4afaef45.png)
In 2.1.0.7 it converts to 9 dusts, in 2.1.0.7a it converts to 9 quarts
#
#### What is your GTNH Discord username?
Quadrocake

Issue #4860, opened 2019-07-19
Iridium Blocks can be crafted directly into ingots


SampsaxD:
GTNH 2.0.7.5

Should use Arc Furnace to convert straight back to ingots like all other metal blocks.

![image](https://user-images.githubusercontent.com/42840657/61534758-2315cd80-aa39-11e9-84af-b9c23e5b55e4.png)

PR #8054, opened 2021-06-14
Change 4 lines to disable some recipes in recipes.cfg

Prometheus0000:
Disable iridium shapeless uncrafting https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/4860
Disable nether quartz shapeless uncrafting https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7956
Disable nether quartz maceration (another recipe is imported from IC2, so you can still macerate) https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7956
Disable clay block maceration (another recipe is imported from IC2, so you can still macerate) https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/8036

I did not remove the 9 dust to 1 block compression recipe so you can still get back to a block if you _really_ want to. Also, don't blame me for IC2's recipes losing half the mass when macerating clay blocks or balls.

Should prevent duping of clay and nether quartz. Also thanks to glease for help with the config file.

PS: I had to update the config file first, since there was a ton of stuff missing. There was also a bunch of stuff in my test instance not in the main pack for some reason, possibly added by obama or something, though they seemed like GT++ materials. So it probably should be updated again after obama is added and GT++ removed.


Dream-Master (changes requested):
use 0 instead of minus values

config/GregTech/Recipes.cfg
```diff
@@ -140857,8 +140857,8 @@ pulveriser {
     I:blockChromiumDioxide_756=756
     I:blockChromo-Alumino-Povondraite_400=400
     I:blockChromo-Alumino-Povondraite_882=882
-    I:blockClay_135=135
-    I:blockClay_400=400
+    I:blockClay_135=-135
```
Dream-Master, line 140860: I:blockClay_135=0

config/GregTech/Recipes.cfg
```diff
@@ -140857,8 +140857,8 @@ pulveriser {
...
+    I:blockClay_400=-400
```
Dream-Master, line 140861: I:blockClay_400=0


and so on

Dream-Master (changes requested):

config/GregTech/Recipes.cfg
Prometheus0000: > set it to any nonpositive value will disable the recipe

Is what glease said. 0 is positive, so I dunno if it works, but I'll try it

Dream-Master: i did all recipes i dont want with 0 and it works

Prometheus0000: It works, so I changed them. Personally, I think - is better since it's possible to search for, but whatever.

Dream-Master (approved):
cant see the change becaue of 30k changes. Will see ingame

PR #8057, opened 2021-06-15
Make centrifuging gravel for flint not terrible

Prometheus0000:


mitchej123 (approved):

Dream-Master:
No this recipe was there in for a reason at this Tier added by me and Namikon. We don't change that.

mitchej123:
> No this recipe was there in for a reason at this Tier added by me and Namikon. We don't change that.

What reason? It's silly

Glease:
what was the designed use case though? I can't quite understand the rationale behind it.

Prometheus0000:
> What reason? It's silly

Probably because you wouldn't think a low-power centrifuge would break apart a rock I guess? Though use-case-wise it makes no sense.

BTW, the reason I wanted to lower the tier of the recipe, besides the fact it seems far too high for it's use case (do you even use flint at EV?) is because getting full flint dusts is pretty terrible, and I was looking for a way to make it less terrible.

Maceration of regular flint gets you tiny dusts, and we're trying to move away from those. Is there a way to change the maceration amount? 2 flint->1 flint dust? Instead of 1 flint to 4 tiny+40% chance for another if you use an HV macerator.

You can use basalt dust, but that's a 16->8 ratio, taking 67s. I guess that's ok for people who do crops?

Dream-Master:
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/commit/7a0167cb50df815f4df017cb6fa1bf1148797644

Prometheus0000:
OK. I'll find another way to make flint dust without tiny piles then. I read the discussion on discord already, BTW.

PR #8070, opened 2021-06-16
Fix ReinforedGlass Plate

yukieiji:
Infusion recipe does not have Oredict enabled, so I'm specifying a specific item

### DIFF
--- a/scripts/Emt.zs
+++ b/scripts/Emt.zs
@@ -1080,7 +1080,7 @@ game.setLocalization("tc.research_text.ShieldFocus", "S.H.I.E.L.D.");
 mods.thaumcraft.Research.addPage("ShieldFocus", "tc.research_page.ShieldFocus");
 game.setLocalization("tc.research_page.ShieldFocus", "With some experimentation with the Portable Hole, you have found out more about the transportation of objects to alternative dimensions. You believe you can reverse these effects to summon a swath of blocks from a dimension full of matter to shield you from any oncoming disaster. Caution is to be taken when using this focus, as when bringing these blocks into our dimension, it will transpose any blocks in the way into the dimension it came from, never to return again. You will also be unable to move while this shield is in existence.");
 mods.thaumcraft.Infusion.addRecipe("ShieldFocus", <Thaumcraft:FocusPortableHole>,
-[<dreamcraft:item.ReinforcedGlassPlate>, <gregtech:gt.blockreinforced:3>, <IC2:itemPartAlloy>, <gregtech:gt.metaitem.01:17383>, <gregtech:gt.blockreinforced:3>, <dreamcraft:item.ReinforcedGlassPlate>, <gregtech:gt.blockreinforced:3>, <gregtech:gt.metaitem.01:17383>, <IC2:itemPartAlloy>, <gregtech:gt.blockreinforced:3>], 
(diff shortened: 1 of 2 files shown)


Prometheus0000:
Infusion auto-oredicts

Prometheus0000:
Also, from what I remember, making it ore:something causes it to not work. Did you test this ingame? Also, what problem are you trying to solve here?

Elisis:
Are there even any other entries for plateReinforcedGlass? It seems pretty specific...

yukieiji:
This request is a fix for an issue that can't craft of shield blocks because the ore dictionary isn't supported.
Also increase the accuracy of the NEI search by matching the items that are automatically converted.

yukieiji:
Seems to work fine in-game.

yukieiji:
@Elisis This should be the only ones as there were not many recipes that use Reinforced Glass Plate.

Prometheus0000:
Again, _infusion already auto-oredicts_. Did you check if it works already? Because I had no problem making the thaumic restorer. And as far as I know, there's nothing to oredict it _with_. Is there more than 1 reinforced glass plate?

yukieiji:
It looks like there are two plates.

Prometheus0000:
Are you saying that they're not oredicted against each other? Maybe that's on purpose? Are they made of the same thing?

yukieiji:
Recipes such as Unrouted Transport Pipe can be created with both "Reinforced Glass Plate" and "Molten Reinforced Glass Plate".
However, MT script recipes such as Wand Focus:Shield can only be created with "Reinforced Glass Plate" (NEI search has to be done with it). and "Reinforced Glass Plate" is converted to "Molten Reinforced Glass Plate".

Prometheus0000:
Someone has done something to mess with the reinforced glass plate that you normally make/use. The GT one did not exist before AFAIK.

Prometheus0000:
I guess it's because of https://github.com/GTNewHorizons/GT5-Unofficial/commit/9f8a8ac1bdcc360d3efd38670f20bbba6b8a28ca#diff-faf76538da89119ddb2b992386b619fd4152e300dd5cf01989b79d036d13690f somehow?

Prometheus0000:
It seems something other than GT or the coremod is doing this, but I'm not sure what. Using a stripped instance from 2.1.0.7 shows only 1 plate, even if I update those mods.

Prometheus0000:
OK, I have no idea where this is coming from. It's still there if I disable the line in the hot ingot commit, it disappears if I remove the material (but doesn't crash and tell me where it's creating the plate), it doesn't appear to be being created because of the config file (updating them all changed nothing). But in my stripped instance, it doesn't exist.

Dream-Master:
@Glease any idea ?

Glease:
That's just hot ingot. It had nothing to do with normal plates.

Dream-Master:
@yukieiji @Prometheus0000 @Glease @Elisis is this fix working or we need another solution ?

Prometheus0000:
It would be nice to find where this thing is coming from, but until someone does, the recipes need fixing since it unifies to the new one, you can no longer craft the old one, but none of the recipes (at least the ones here) accept the new plate. Yukieiji said it works, so...

Dream-Master (approved):

Issue #8089, opened 2021-06-22
Duplicate Nickel and Cobalt quest in Novice Thaum tab 


tommy307507:
#### Which modpack version are you using?
2.1.1.0
#
#### If in multiplayer; On which server does this happen? 

#
#### What did you try to do, and what did you expect to happen?
Do quests :D After mining some nickel
#
#### What happened instead? (Attach screenshots if needed)
![image](https://user-images.githubusercontent.com/49487505/122925151-23dbda80-d399-11eb-8b6d-cc0819a28cf2.png)
![image](https://user-images.githubusercontent.com/49487505/122925161-263e3480-d399-11eb-9adb-7c5006e96868.png)
Guess you get two for the price of one?
![image](https://user-images.githubusercontent.com/49487505/122925210-31916000-d399-11eb-8fb5-7faa7021327b.png)
![image](https://user-images.githubusercontent.com/49487505/122925232-36561400-d399-11eb-919c-49a17a1806d3.png)
Its highly similar and should be duplicate ... one doesnt have the colored text...
#
#### What do you suggest instead/what changes do you propose?
Remove one of them I guess? Why say the same thing twice? 
#
#### What is your GTNH Discord username?
Hebi

Issue #8083, opened 2021-06-19
Steam Grinder Quest requiring wrong number of blocks 


tommy307507:
#### Which modpack version are you using?
latest dev  2.1.1.0.pre2
#
#### If in multiplayer; On which server does this happen? 
Stone Legion
#
#### What did you try to do, and what did you expect to happen?
Finish the steam grinder quest 
#
#### What happened instead? (Attach screenshots if needed)
![image](https://user-images.githubusercontent.com/49487505/122633592-fd553f80-d10b-11eb-94af-a1f65a0ec548.png)
Steam grinder is a 3x3x3 hollowed multi 
i.e. it needs 26 blocks in total to function and the quests asks for 27 blocks in total
#
#### What do you suggest instead/what changes do you propose?
Update the quest to need 22 bronze plated bricks instead...

#
#### What is your GTNH Discord username?
Hebi

PR #8120, opened 2021-06-26
#8083, #8089. Add quest to explain all the covers. Add skip quests

Prometheus0000:
Skip quests will allow skipping the individual cell quests for things like osmium, if you hold a stack of it (because you already did it, and more than the quest asks for).

I also made them chain quests, that will show the chain when unlocked.

Also for rocket quests if you make titanium/TS without going to space.

If there was another one that would benefit from this, make a ticket. But I'm not doing it for every little thing, as it needs another quest to do it.


Dream-Master (approved):

Dream-Master:
@Prometheus0000 didnt test what "chain" in a quest did but i trust you on this.

Prometheus0000:
It just shows all quests in a chain at once as long as they're marked as chain.

Dream-Master:
ok will see ingame

Issue #7927, opened 2021-05-14
RFC: EC2 destiny


repo-alt:
Due to outstanding code quality, this mod gained universal respect among developers. To get to it's current state it was rewritten from scratch twice, once in a different language (there was also another attempt in yet another language, for the best of luck). It's maintenance is also made so much easier taking into account great fluid support in Forge. Unfortunately, there are some items that do not really fit the theme of the mod, like "Certus tanks", "Large storage cells" and "ME robot upgrade", these should be removed (to AE2 and OC respectively), to preserve perfection and consistency of the mod. If you think there are more items that do not really fit the theme of this great mod, please comment below.

Prometheus0000:
@repo-alt Is this still happening, or did you get busy? I recall you having second thoughts?

repo-alt:
Well I ported certus tank to AE2 to test the concept, and tested that id remapping works, but after that I cooled down a bit. I mean people who use it on their own servers or in SSP may not really care about bugs and dupes, and on officials we may just disable recipes if necessary. As such it is not worthy time investment.

Prometheus0000:
Then I think there needs to be a discussion about what to disable.

repo-alt:
I would disable fluid interface at least.

GTNH-Colen:
> I would disable fluid interface at least.

And the ore dict bus now that we have ore dict cards instead.

Prometheus0000:
Higher tier storage cells: Keep
Crafting storages: Keep
Quartz Tank: Keep
ME OC things: Keep

ME Block Container:Keep I guess
ME Portable Fluid Storage: Keep I guess
Walrus: Keep I guess

Fluid Vibration Chamber: Either way
ME Fluid Annihilation Plane: Either way
ME Fluid Formatting Plane: Either way

ME Drive Fixture: Is this even useful?
ME Energy Cell Fixture: Is this even useful?

Fluid storage cells: Remove
ME Fluid Terminal: Remove
ME Fluid Assembler: Remove
ME Fluid Interface: Remove
ME Fluid Auto Filler: Remove
ME Fluid Export Bus: Remove
ME Fluid Import Bus: Remove
ME Fluid Storage Bus: Remove
ME Fluid Level Emitter: Remove
ME Fluid Storage Monitor: Remove
ME Fluid Conversion Monitor: Remove
ME Fluid Pattern: Remove
Wireless Fluid ME Terminal: Remove
[...]

PR #8159, opened 2021-07-01
Remove/hide EC2 items, and other useless AE2 related items. Also change TE recipes

Prometheus0000:
Vibration chambers are useless, so removed. Grinder was removed already, so removed the related gear items.

Otherwise, see #7927.

Quests need to be adjusted after.

### DIFF
--- a/scripts/AE2.zs
+++ b/scripts/AE2.zs
@@ -678,12 +678,6 @@ recipes.addShaped(<appliedenergistics2:tile.BlockEnergyAcceptor>, [
 [FluixCrystal, GlowstonePlate, FluixCrystal],
 [TitaniumPlate, FluixCrystal, TitaniumPlate]]);
 
-// --- Vibration Chamber
-recipes.addShaped(<appliedenergistics2:tile.BlockVibrationChamber>, [
-[TitaniumPlate, <ore:screwTitanium>, TitaniumPlate],
-[<ore:craftingToolScrewdriver>, <gregtech:gt.blockmachines:264>, <ore:craftingToolWrench>],
-[TitaniumPlate, <appliedenergistics2:tile.BlockEnergyAcceptor>, TitaniumPlate]]);
-
 // --- ME Security Terminal
 recipes.addShaped(<appliedenergistics2:tile.BlockSecurity>, [
 [TitaniumPlate, <appliedenergistics2:tile.BlockChest>, TitaniumPlate],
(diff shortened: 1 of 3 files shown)


GTNH-Colen (commented):
Looks good.

Dream-Master:
I like to see a implementation of staff in ae2 first and remove ec after it. @repo-alt any future plans on this ?

repo-alt:
It is kind of stupid jub to move working stuff, since there are quite a bit remaining in EC2. That is, - I tried the moving and found it possible, but not quite trivial process, so I think this job has too few benefits to be worth the efforts.

Dream-Master:
@repo-alt so you vote for drop the mod without any alternative way store fluids in ae systems ?

repo-alt:
Nah, I vote to disable half of the mod and leave it be.

Dream-Master:
@Prometheus0000 @repo-alt ok but we need to do this inside the mod not with scripts  that it not loadedn all the unsed stuff.

repo-alt:
Sure, but I thought the idea was to make disabling recipes a first step, and then disable the blocks themselves (if that is even necessary, since, as I said, SSP users have all the right to kill their TPS). Main concern now is to disable recipes and not break anything else, like ThaumicEnergistics recipes.

Dream-Master:
Ok we can make a server script that disable ec2 stuff and let them enable in sp. Otherwise we decided for all players what they should do. @Prometheus0000 we should make a server branch and a client branch.

Prometheus0000:
I have no idea how to make a 'server script'. I've edited the scripts to remove the recipes and hide the items, my work here is done. 

boubou and colen were complaining about people using fluid ae and lagging the servers, so I decided to do this quickly. EC2 is not in java, so I don't even want to touch it, so I'm not disabling items/mechanics in the mod myself.

Additionally, my understanding was that we were removing this stuff from SP as well. Altering the quests in such a way as to have them valid for both having and not having EC2 fluid stuff would mean having two versions of the questbook, which I'm not doing.

Dream-Master:
Ok i can make new branches for server and client. And you can write in the book that this are disabled on server so not a big deal. Not need two versions of Quest book.

Prometheus0000:
OK

Prometheus0000:
Make sure to separate the 'this is useless so it's removed' removals from the 'this causes lag so it's removed' removals. Some things should still be removed in SP that aren't about fluid ae:

Vibration chamberx3 are not deisgned for 10x power use, walrus has some kind of rendering bus, blast resistant drive is crazy buggy, the part fixtures of power and drive had some problem I forget or were useless, the portable fluid thing I couldn't think of any use case for to leave it enabled, the 3 gear items have no use because the ae2 grinder is disabled.

Prometheus0000:
Did anyone ever do this?

PR #8194, opened 2021-07-05
Removed Tooltips on Adv Backpacks

### LINKED ISSUE #8189: Idiots text makes me feel worthless and want to stop playing
#### Which modpack version are you using?
2.1.1.0
#
#### What did you try to do, and what did you expect to happen?
I looked at the backpack and expected to see a helpful tooltip for it.
#
#### What happened instead? (Attach screenshots if needed)
I felt like I was being called an idiot and got depressed.
#
#### What do you suggest instead/what changes do you propose?
Remove the 'Idiots' part.
#

Ethryan:
Since this is supposedly fixed by #7137 and #7299.

Should fix #8189.

### DIFF
--- a/scripts/Advanced-Backpacks.zs
+++ b/scripts/Advanced-Backpacks.zs
@@ -288,9 +288,4 @@ mods.gregtech.Brewery.addRecipe(<liquid:melonjuice> * 750, <minecraft:melon>, <l
 
 
 
-// --- Tooltips ---
-
-
-
-
-<Backpack:backpack:*>.addTooltip(format.red("Idiots! Do not upgrade your backpack with important stuff inside - Dream 2017!"));
+// --- Tooltips --- 


leagris (changes requested):
Please add conservation of inventory when upgrading a backpack

Prometheus0000:
Doesn't it already do so now?

> Yes, I went in game and put the crafting recipes through a bench. All items that are in the input backpack persist though the output. I did this because I wanted people on my server to be able to swap out backpack styles as they pleased.

From https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/7137

Or do you mean 

> This can also be done for the backpack in "Backpack.zs" to prevent items being deleted when upgraded, but that will take more time on my part.

Which I don't think has been done and doesn't relate to the tooltip?

Ethryan:
#7292 - should include the backpacks.zs script fixes for backpack saving.

leagris (approved):

PR #8229, opened 2021-07-11
Better-Builders-Wands Tooltip

TrainerSnow:
Added Tooltips For BetterBuildersWands, informing about the /wandOops Command.


Prometheus0000:
Do they really need to have a tooltip like this on all of them? That's just annoying, it'd be better to put the command in the related quest instead.

TrainerSnow:
I would say yes, it can be really annoying to clean up the 2nd layer of a wall that should only be 1 block thick, because of a misclick. Having it in the quest is a good idea, but having it on the Wand directly would make it easier to detect. I could change it to only display on Shift-Clicking to be less obstrusive.

Dream-Master:
> Do they really need to have a tooltip like this on all of them? That's just annoying, it'd be better to put the command in the related quest instead.

If people dont do quests it better to have a tooltip?

Dream-Master:
@TrainerSnow @Prometheus0000 so what is the decission on this ? Quest or Tooltip ?

Prometheus0000:
I still say putting it in a quest is best, but you're the one that decides.

Dream-Master:
ok quest

PR #8386, opened 2021-08-10
Disables Witch of the Wind / Set Spawn Chance to 0

ScriptedPiky:
Disables them until they can be reworked. Otherwise, they are complete bullshit that has no legitimate counter-play.

### DIFF
--- a/config/SpecialMobs.cfg
+++ b/config/SpecialMobs.cfg
@@ -458,7 +458,7 @@ witch_rates {
     I:shadows=1
     I:undead=1
     I:wilds=1
-    I:wind=1
+    I:wind=0
 }
 
 


Prometheus0000:
I think you need to say _why_ they're BS. Do they TP follow you like jolts?

ScriptedPiky:
OH right. Indeed they do. They teleport after seemingly taking damage or the player noticing them on a short cooldown. This teleport will also try to follow the player but the Witches may tend to lose line of sight in wooded or condensed areas or if they teleport slightly far enough to break aggro. In this case, they will attempt to jug down a Healing potion (or other potions depending on the situation) that does not cancel or pause their teleportation before attacking the player once more. Additionally, if they have Bulwark and or Regen and or are applying Weakness, the only way to defeat these things is to run away or have enough damage that you just one-shot them regardless. Hitting them with a crossbow does almost nothing unless you hit them during the drinking potion animation. It also seems that maybe the original AI copies from Enderman AI in a way with how it interacts with projectiles.

Dream-Master:
And again why we dont try to fix the original code. I rember moron and whlm just fix the special and infernal mobs code. Maybe an overseen ?

Prometheus0000:
What specifically would fix it though? Not letting it teleport? Then it's just a regular witch I think? Maybe adding a 1-2m cooldown on TPing? Or removing the immunity to projectiles? I've never encountered them, so I dunno.

Dream-Master:
yes maybe a cooldon or a removing the immunity to projectiles there are mayn options

ScriptedPiky:
The teleporting could possibly be removed but the mob could also be made to move "faster" and have DR/damage reduction vs projectiles in exchange for a lack of teleporting. This could similarly be done with the original Jolt Blaze, but I'm not sure how the code was or is still like for comparing, or how to even apply the movement speed & the DR vs projectiles to the Witch of the Wind. Though if the DR implementation is too complicated then the Witch can just obtain the speed boost instead.

Dream-Master (approved):

PR #8428, opened 2021-08-19
Renovate issue templates

Glease:
We have too many different kind of issues. There is simply no one-size-fits-all template for an issue tracker of this scale. I have divided all our issues into 7 categories, each with their own template. Even better, with github's issue form functionality, the issue template are now a lot more clear as what need to be filled.

Play with the change live here: https://github.com/Glease/IssueTemplateExperiment/issues/new/choose Feel free to open issues there with dummy text to see how it looks like after submitted.


Dream-Master:
@Glease is this done? So we can merge it. Changes we can add later as well.  Did the tagging works?

Glease:
Yes it should work just fine.

Dream-Master (approved):

PR #8434, opened 2021-08-21
Fix lootbag typos

chavalitp:


TimeConqueror:
But LV is really a Tier 1, while ULV is like Tier 0...

chavalitp:
Yeah, But in the quest book Tier 0.5 is already the name of Stone Age, Calling Tier 0 ULV may cause confusion? I'm not sure if ULV is lower or higher than Tier 0.5 Stone Age.

Prometheus0000:
ULV/tier 0/whatever loot bag is not used and should be removed.

chavalitp:
When I remove ULV from LootBags.xml. It's transform to "LootBag (Error)". Not sure how to completely remove from the game.

Dream-Master:
Tier 1 was a placeholder. We cant remove it now anymore because if we shift the bag ids all bags are one tier better if you have bags ingame. I would say let's rename it to ulv or Tier 0 and all fine.

Dream-Master (approved):

PR #8455, opened 2021-08-26
add Tier 3-5 Barrel Upgrade Quests

Dream-Master:


Dream-Master:
@Prometheus0000

Prometheus0000:
I'm not sure how many people use barrels over drawers, considering how much more useful drawers are, but I don't see a specific problem with these quests or anything. Though I don't really see the point in making quests for more upgrades, we don't do that for the more useful drawer ones either, we assume people can figure that out. I assume they link off the previous ones?

Dream-Master:
They optional so not a big deal. If you dont do it nothing is locked behind.

Issue #8568, opened 2021-09-20
Illumar recipe consumes programmed circuit


MatthewCOwen:
### Your Pack Version

2.1.1.3

### Your Server

Zeta

### Your Expectation

I expected the circuit #8 to not get consumed.
![image](https://user-images.githubusercontent.com/16054364/134054393-4631eaf6-aabb-4bf6-a29b-7c4d504ed9cd.png)

### The Reality

![image](https://user-images.githubusercontent.com/16054364/134054490-20a16883-384c-4d69-85d0-0e0b02ddf707.png)

### Your Proposal

please check to make sure the illumar recipes don't consume the programmed circuit

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

Prometheus0000:
Similar to https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/8559

Dream-Master:
oh nice need to change

Issue #8559, opened 2021-09-19
Circuit is being consumed


Pxx500:
### Your GTNH Discord Username

Pxx500#6984

### Your Pack Version

2.1.1.2 with newer jenkins (before 2.1.1.3)

### Your Server

Private

### Type of Server

Vanilla Forge

### Your Expectation

Don't eat my circuit

### The Reality

My circuit is being eaten

![image](https://user-images.githubusercontent.com/81298696/133941960-da6db9df-561c-402e-af6d-2e95ed6b186f.png)

### Your Proposal

Pls fix

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [x] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

PR #8569, opened 2021-09-20
Fix recipes consuming circuits, and use oredict properly

Prometheus0000:
Should fix https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/8568 and https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/8559

Dream, you forgot to add the * 0 part. Also, why didn't you use oredict on the dyes, it's way less work?

### DIFF
--- a/scripts/Project_Red.zs
+++ b/scripts/Project_Red.zs
@@ -1621,175 +1621,52 @@ Mixer.addRecipe(<ProjRed|Core:projectred.core.part:58>, [<gregtech:gt.metaitem.0
 Mixer.addRecipe(<ProjRed|Core:projectred.core.part:57>, [<minecraft:iron_ingot>, <ProjRed|Core:projectred.core.part:56> * 8], 400, 30);
 
 // --- White Iluminator
-Mixer.addRecipe(<ProjRed|Core:projectred.core.part:19> * 8, [<minecraft:glowstone_dust> * 2, <minecraft:dye:15> * 2, <gregtech:gt.integrated_circuit:8>], 50, 8);
-// -
-Mixer.addRecipe(<ProjRed|Core:projectred.core.part:19> * 8, [<minecraft:glowstone_dust> * 2, <BiomesOPlenty:misc:8> * 2, <gregtech:gt.integrated_circuit:8>], 50, 8);
-// -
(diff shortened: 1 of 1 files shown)


leagris:
> Dream, you forgot to add the * 0 part. Also, why didn't you use oredict on the dyes, it's way less work?

Probably because Gregtech's MineTweaker interface does not support Ore Dictionary ingredient properly, It only take the first ingredient in the list. So if you want all white dyes you have to enumerate all individual item of white dye.

If doing the recipe internally to Gregtech, you'd be able to use OredictUnification

Prometheus0000:
All 6 show up
image
All of these worked
image

Prometheus0000:
It'd be nice if someone fixed this, but having it look better ingame is more important, I'll change it back.

Prometheus0000:
They show up as separate recipes even when I switched it back, so I don't see a point in doing so.

Dream-Master (commented):
you sure all the dyes are ore dicted ?

Dream-Master:
> > Dream, you forgot to add the * 0 part. Also, why didn't you use oredict on the dyes, it's way less work?
> 
> Probably because Gregtech's MineTweaker interface does not support Ore Dictionary ingredient properly, It only take the first ingredient in the list. So if you want all white dyes you have to enumerate all individual item of white dye.
> 
> If doing the recipe internally to Gregtech, you'd be able to use OredictUnification

Minetweaker support oredict since a few versions (for Assembler etc)

Dream-Master:
> Uh, I just tested it ingame and they all work?

if it is tested i am fine with it.

Dream-Master (approved):

leagris:
> Minetweaker support oredict since a few versions (for Assembler etc)

As I said earlier, the support is very limited since it just register multiple recipes for each oreDict item entry rather than the machine handling oredict name directly in its recipes map and rotating alternative ingredients from the oreDictionary within NEI

Dream-Master:
Or should I add recipes to code ?

leagris:
Recipes in code would allow use of the GT_Unificator using the orename

Prometheus0000:
Can you remove the original recipes if you put it in the coremod though? I have no idea how to do that (or if it's possible), or I'd suggest slowly migrating most things.

PR #8620, opened 2021-10-01
Delete NEICustomDiagram.cfg

D-Cysteine:
I was thinking that it would probably be better to just let the config file be auto-generated. This way, we don't have to keep updating a copy of the config every time new config options get added.


Prometheus0000:
I agree for now, but maybe add it back after there's less changes to it so often?

D-Cysteine:
Hmm, I was thinking that we don't need it, because it will be auto-generated with default values anyway. We should only need it if we want to have values other than the default, I guess. But then we could just change the defaults?

Prometheus0000:
I guess that's true.

Dream-Master (approved):

PR #8632, opened 2021-10-01
Clarify two quests to reduce confusion

bombcar:
Make it obvious that there are recipes with a percent change when starting AE, and that the conduit covers for ender io only work when painted.


Elisis (approved):

bombcar:
Added a second fix to a quest not asking for the produced amount.

PR #8643, opened 2021-10-05
Add nether portal to list to scan

bombcar:
Apparently needed to trigger research for mirrors and easy enough to scan - twilight portal may work too?


Prometheus0000:
The nether portal is not one of the items you need to scan to unlock all the aspects, which is what this quest is for. If it's necessary for research, it should be in a separate list at the bottom.

bombcar:
I can move it to a separate list; I only know of Nether Portal (ender pearl is one, but already listed)

bombcar:
Fixed minor typo

Dream-Master:
@Prometheus0000

Dream-Master:
can i merge this ?

PR #8677, opened 2021-10-12
Add note to Blastoff referencing steam compressor

bombcar:
It's a common discord question

Fix typos, clarify anvil


bombcar:
The quest for steel referencing Blastoff is visible before Blastoff. They get the steel quest and then can’t find the BBF quest. (Maybe the steel quest it has been updated and is now dependent on steam compressor?)

PR #8679, opened 2021-10-13
LP recipe fixes and rebalances

Glease:
As I have recently restarted, I think it'd be a good chance to rebalance LP recipes, since people are always saying stuff are too expensive.


Glease:
I have given this a bit more thought: an average GT recipe do 4 I/O, which roughly translate to 3 LP modules/pipes, i.e. 6~10 circuits. This is way more than what the machine itself cost. My plan is to make an average machine to require <2 circuits to fully automate. After all, EnderIO conduits don't even need circuits to begin with, and can achieve most of what LP can do.

Glease:
This is done for now and is ready for review.

Though do note that I left the autocrafting related pipes/components unchanged, as I did not use them in my own playthrough. They might still need another round of balancing, but I'll leave that to the others.

Dream-Master (approved):

Dream-Master:
lets test this on zeta

PR #8708, opened 2021-10-18
fix quest description and use SL Blueprint

Dream-Master:


Glease:
Quest 2856 (on Multiblock Goal tab) probably also needs editing. It mentions a list of mods that doesn't work with structurelib blueprints, which probably is no longer valid.

Prometheus0000:
> BTW that quest description looks quite different from what Dream (and I, but I'm not on 2114-dev-2 yet) have (see https://raw.githubusercontent.com/GTNewHorizons/GT-New-Horizons-Modpack/dae3e5d1c51257e1a0ed83055344b990044ac607/config/betterquesting/DefaultQuests.json) Did you change it locally but didn't push.

I just did a compare, and that quest wasn't touched, did you reload the QB after getting a new copy from the repo?

PR #8717, opened 2021-10-20
clarify that WA doesn't work with GT machines

bombcar:
someone in discord was sad their WA wasn't speeding up the ore washer 🥇


bombcar:
Aha! I now know how to add a reward to a quest (it's just a normal cake, was going to use the harvest craft cheesecake but that may not be craftable (or food))

Issue #8326, opened 2021-08-01
RFC: Mirrors are overnerfed


repo-alt:
#### Which modpack version are you using?
2.1.1.1
#
Magical mirror is currently gated behind enderioum, which means TS coils, so, kind of, late EV. At that stage it is easier to wait until trasceivers which are much superior. (Mirrors require ordo vis source nearby to have any throughput)
#
To make them less useless, perhaps replace enderium plates in the recipe with enderium base? It still requires some platinum, so not cheap at late HV. Or some early EV alloy? Maybe even tungstensteel, there is quite some gap between first TS ingot and TS coils

monroth:
Agree
It is the most useless transdimensional item transport already - without special setup with energised ordo node it has low speed and it voides items
There is no way you would use them unless there is literally no other way
No reason to not give this ability at HV

leagris:
Remember not to use them in Galacticraft dimension or expect memory leak and run-away mob spawns in GT Planets chunks where a mirror is placed.

repo-alt:
> Remember not to use them in Galacticraft dimension or expect memory leak and run-away mob spawns in GC Planets chunks where a mirror is placed.

Doesn't that also affect any chunkloader?

PR #8761, opened 2021-11-02
Magic Mirror Adjustment

ScriptedPiky:
Changed Enderium to Enderium Base; from EV to late HV tier. Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/8326

### DIFF
--- a/scripts/Thaumcraft-04-Artifice.zs
+++ b/scripts/Thaumcraft-04-Artifice.zs
@@ -623,7 +623,7 @@ mods.thaumcraft.Research.setComplexity("RUNICARMOR", 3);
 
 // --- Magic Mirror
 mods.thaumcraft.Infusion.addRecipe("MIRROR", <Thaumcraft:ItemResource:10>,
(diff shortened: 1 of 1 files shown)


Dream-Master:
not sure if enderium base is a better material. Sure the mirror is tiered a way to high. Will check ingame.

Ethryan:
Vibrant Alloy maybe? Or if that is too lategame maybe Energetic Alloy?

ScriptedPiky:
Enderium Base seems close to "magical" but is still in late HV. Vibrant Alloy also works as it's a late HV material. Maybe I could add Astral Silver (very early HV) and replace the Enderium with Vibrant Alloy or maybe keep & add both?

Dream-Master:
@ScriptedPiky can you change it to vibrant alloy ?

ScriptedPiky:
All ready now. Not sure if the recipe is lacking in specific materials or "difficulty" for that matter, other than the Vibrant Alloy gate, but this should do.

Dream-Master (approved):

Issue #8694, opened 2021-10-16
No Boom Tomorrow gives Fused Quartz


bombcar:
### Your GTNH Discord Username

bombcar#1042

### Your Pack Version

2.1.1.0

### Your Proposal

The quest "No Boom Tomorrow" tells you to not used Fused Quartz but then gives you ten as a reward.

Bronze Plated Reinforced Stone is cheap enough, I suggest we change the reward to that. Or Warded Glass. Or a "pick one" from a various list of blast-proof materials.

### Your Goal

Give a useful reward instead of a useless one.

### Your Vision

Have rewards be consistent with the quest.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

bombcar:
If a replacement is determined I can update the quest.

Elisis:
I thought this was supposed to be a joke, giving you an item that doesn't work anymore?

bombcar:
Could be - in which case I can revert. Not sure what the quest was before the fused quartz nerf

PR #8804, opened 2021-11-09
Update Time to Drill for Ore to not require MV hatches

bombcar:
By the time you build the MBM HV is easily obtainable, no need to force MV hatches on people if they don't want them.


bombcar:
Added a fix for #8694

bombcar:
Updated to clarify #8694

Dream-Master (approved):

PR #8811, opened 2021-11-11
Chinese translation of quest book for 2.1.1.4QF-2

Kiwi233:


Dream-Master:
@GlodBlock would you do a check about it. Thanks
and @Kiwi233 thanks for you work.

GlodBlock (approved):

Issue #8611, opened 2021-09-28
[RFC] Suggested new warp effects for Warp Theory


D-Cysteine:
### Your GTNH Discord Username

D-Cysteine

### Your Pack Version

2.1.1.3

### Your Proposal

Here's some ideas that I came up with; please give feedback and/or suggest new ones!

In roughly suggested order of increasing warp:
 * For a short period of time, flowers are spawned at the player's feet (if placeable). Maybe limit to grass blocks only; this shouldn't break anything in bases.
 * Player receives slowness, weakness, mining fatigue.
 * For a period of time, player randomly lays eggs.
 * For a period of time, player either cannot sleep, or takes twice as long to fall asleep.
 * Spawn a wisp near the player.
 * Player instantly loses half their current health (rounded down so this cannot kill the player).
 * Spawn a doppelgänger behind the player. Similar to the friendly creeper, the doppelgänger is a passive mob that follows the player. The doppelgänger looks just like the player and has the same max health, and any damage that it receives is also dealt to the player (maybe with a chat message each time this happens?).
 * Spawn some hostile endermen a moderate distance away. Maybe also give the player an ongoing effect that causes nearby endermen to become hostile and/or randomly spawns more endermen.
 * For a period of time, the player randomly gets pushed around.
 * Water is converted to ice in an area around the player. Maybe also place some snow.
 * Spawn a bunch of mushrooms in an area around the player (low chance of vishroom). Cows in range are converted to mooshrooms (need to figure out how this interacts with Thaumic Horizons infused cows).
 * _A voice whispers in your ear, "You look like you need this"._ Gives the player some litmus paper. This is a not-so-subtle hint that the player is about to reach (or maybe has reached) wither-level warp. Maybe give the player an effect that prevents receiving this warp effect again for a long time, to prevent getting free litmus paper?

I'm happy to work on implementing these myself.

For balancing, note that it looks like a random event is picked (out of those the player has enough warp for), so the more events that we add, the less likely any individual event is to be picked. See:
https://github.com/GTNewHorizons/WarpTheory/blob/master/src/main/java/shukaro/warptheory/handlers/WarpHandler.java#L228-L238

There is also a closed issue with more ideas: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/4270

### Your Goal

[...]

PR #8849, opened 2021-11-15
Add new warp effects to config

D-Cysteine:
Updating config to use warp levels suggested in:
 * https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/8611#issuecomment-958003423


TheElan:
Are values in quest already updated? If not I would suggest updating them here too, so it's all enters the pack in one PR

D-Cysteine:
Thanks! I've updated the quest.

yukieiji:
There must have been a sentence in Thaumonomicon that needed to be changed.

D-Cysteine:
Thanks! Updated that too, I think

Dream-Master (approved):

Issue #5956, opened 2020-04-21
[Sense] uncorrectly working


Motias-git:
#### Which modpack version are you using?
2.0.8.3
#
#### If in multiplayer; On which server does this happen? 
eta.gtnewhorizons.com
#
it continues  #4761 issue
#
- Sense excavating 5x5x5 i.g. dirt, gravel, stone ..
- Using right click harvest not IC2 crops, destroys blocks under it i.g. mods Natura, Pam's
#
https://youtu.be/pc89m3ovEgo

GTNH-Colen:
At least it doesn't break bedrock anymore? lmao

MotDT:
> 
> 
> At least it doesn't break bedrock anymore? lmao

Yes, only basic block i.g. dirt, gravel, stone, wood..

Cleista:
Still occurs on 2.0.9.0QF4

- [Try to harvest crops](https://cdn.discordapp.com/attachments/181078474394566657/728317113197920417/unknown.png)
- [Accidentally dismantle base instead](https://cdn.discordapp.com/attachments/181078474394566657/728317188267835401/unknown.png)

bombcar:
How can I get this to occur? It doesn't seem to be triggering in 2.1.1.5pre - does it need particular blocks?

bombcar:
I think https://github.com/GTNewHorizons/GT5-Unofficial/commit/42543040b0533d2cd714412efa29a51b75321bb6 fixed Most of this - it harvests bushes like the berry bushes instead of just picking the berries, but eh, that's not too bad.

PR #8885, opened 2021-11-20
Cleanup and clarify MMCC location

bombcar:
Also it appears that fullytoasted may be down and should be removed


Dream-Master:
yes toast servers not active anymore. I remeber i change the list a bit but maybe been not up to date.

bombcar:
> yes toast servers not active anymore. I remeber i change the list a bit but maybe been not up to date.

I'll remove

bombcar:
Added a quest text update because it seems #5956 has been mostly fixed.

Dream-Master (approved):

PR #8904, opened 2021-11-22
Tier 0.5 quests minor modifications

gabrielnoguerales:
- Modify dependancies of quests to reflect dependancy of Steel or Wrought Iron
 - Added some tips for the "Time to get some Steel" quest


config/betterquesting/DefaultQuests.json
```diff
@@ -335210,9 +335230,9 @@
       "home_image:8": "dreamcraft:textures/gui/gtnh_title.png",
       "party_enable:1": 1,
       "pack_version:3": 2110,
-      "editMode:1": 0,
+      "editMode:1": 1,
```
Dream-Master, line 335233: please disable edit mode or set this back to 0

config/betterquesting/DefaultQuests.json
```diff
@@ -4473,7 +4473,9 @@
     "47:10": {
       "questID:3": 47,
       "preRequisites:11": [
-        46
+        46,
+        76,
+        948
```
Dream-Master, line 4478: I am not ingame what quests are 76 and 948

Dream-Master: or add more text in your pr text field. Thanks

config/betterquesting/DefaultQuests.json
gabrielnoguerales: Done, thanks.

config/betterquesting/DefaultQuests.json
bombcar: 76 - Time to Get Some Steel
948 - Better Than Iron
46 - More Advanced Alloys

gabrielnoguerales: Thanks @bombcar

config/betterquesting/DefaultQuests.json
```diff
@@ -7666,7 +7683,7 @@
...
+          "desc:8": "You need steel? Uhh... fine. Let\u0027s get busy...\n\nHint: The bricked blast furnace can share walls with other ones, just like most GT multi blocks. How will this affect the air quality though...\n\n§3If you are short on Stone Dust, you can get more by macerating cobblestone.\n\nIf you spawn near a biome with Hardened Clay you can macerate it to get the clay dust you need. You can also use brick blocks from Villages to get brick dust in the macerator.\n\nIf you can, craft a WorkTable from Forestry it will make the crafting a lot easier as you can craft with all your buckets in one click\n\n§4You\u0027ll need 432 unfired clay brick, 432 brick, 108 clay and brick dust, 216 compressed fireclay, and 216 firebrick. Collect all the required materials for these first, or you\u0027ll have to do it again. Also, don\u0027t make the firebrick blocks until you collect all the gypsum you need unless you want to be forced to get more."
```
bombcar, line 7686: The comment text change is very good (though I'd word the sentence as:

Crafting the WorkTable from Forestry can make it easier as you can use all the buckets in one click.

bombcar (commented):
As for the other changes, while well intentioned, I think they make the quest page very "spiderweby" - perhaps instead some could be entirely hidden until the appropriate time?

image

vs

image

gabrielnoguerales:
@bombcar I agree it makes the quest page very "spiderweby", some quests need to be reorder to adjust to make them more clear about their dependancies. I can try to do it but it overwhelms me a little bit as I havent played before with betterquest.

@bombcar @Dream-Master I pushed more changes to the pr if you could take a look

config/betterquesting/DefaultQuests.json
gabrielnoguerales: Great

bombcar (approved):

bombcar:
>  What exactly was the goal with some of the changes? Make it clear you need to do another quest? Perhaps the solution instead of dependencies across the map would be to add some quest text - "If you can make ..."

gabrielnoguerales:
The purpose was to not have quests unlocked that you can't do before other ones. In other words so all unlocked quests are compleatable without needing to complete another one before. 
It is specially useful when deciding what quest to make as it deletes the path you currently can't take.

bombcar:
> The purpose was to not have quests unlocked that you can't do before other ones. In other words so all unlocked quests are compleatable without needing to complete another one before.
> It is specially useful when deciding what quest to make as it deletes the path you currently can't take.

The problem with that is very quickly you get many quest that depend on each other - GTNH isn't really a "railroad" pack where you have to follow one particular quest line - you're meant to learn how to use NEI and realize what you need to build (later on referencing other entire tabs).

I do agree it can be confusing early in steam, which is why I think quest text updates may help - and prevent the spiderweb of cross-connected quests.

gabrielnoguerales:
> > The purpose was to not have quests unlocked that you can't do before other ones. In other words so all unlocked quests are compleatable without needing to complete another one before.
> > It is specially useful when deciding what quest to make as it deletes the path you currently can't take.
> 
> The problem with that is very quickly you get many quest that depend on each other - GTNH isn't really a "railroad" pack where you have to follow one particular quest line - you're meant to learn how to use NEI and realize what you need to build (later on referencing other entire tabs).
> 
> I do agree it can be confusing early in steam, which is why I think quest text updates may help - and prevent the spiderweb of cross-connected quests.

I agree and its one of the things I am enjoying the most, being able to have multiple paths to take (Fe. I took a detour on Steam tier to do some bees). But when you arrive to Steam tier you unlock quests that you can only make after having materials that puts you near the end of the tier so I though it was better to have those quests unlocked when you have the materials needed to complete them.

bombcar:
I believe it is possible to hide a quest until one is completed without making it a dependency (so no line) - but not exactly sure how to do that.

Dream-Master:
@gabrielnoguerales sorry but this is a real mess. We cant let this like it is now. 
1. i make a reverse of the pr
2. you find a way to rearraged the icons no lines corssing
3. someone can add some code to BQ remove this lines (so you can choose if you want to show the dependencies or not)

bombcar:
BetterQuesting now supports IMPLICIT which is what you’d want to use for these (the long lines would only appear on hover over)

Dream-Master:
Are the implict works on single line or all lines are affected ?

bombcar:
> Are the implict works on single line or all lines are affected ?

I need to check but it looks like each one can choose - default normal.

bombcar:
let me try update - don't revert yet

bombcar:
All fixed in https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/8938 - requires latest BQ

PR #8930, opened 2021-11-27
Adjusting the quest "At the End of..."

Eldrinn-Elantey:
Fixing the required number of frostproof machine casings in the quest since 21 is not enough


boubou19:
3x3 cube hollow = 26 blocks
26 blocks - controller - energy hatch - maintenance hatch - input bus - output bus =26 - 5 = 21.

Eldrinn-Elantey:
> 3x3 cube hollow = 26 blocks 26 blocks - controller - energy hatch - maintenance hatch - input bus - output bus =26 - 5 = 21.

But what about another one that we spend on the controller? In other quests , this moment takes into account

boubou19:
ah yeah forgot about this one then in this case it's 22.

bombcar:
> ah yeah forgot about this one then in this case it's 22.

Are we consistent on that? Usually I remember building the controller, then building the casings (more) …

bombcar (approved):
Looks good. If someone overproduced by accident they can save or disassemble

PR #8959, opened 2021-11-29
Fix issue template

Glease:
Someone renamed `Type: suggestion` to `Type: Suggestion` and accidentally broke the issue templates. It does look better though, so let's keep it. This PR will fix the issue template.


boubou19 (approved):

Ethryan:
Sorry, 
Tried to standardize all issue tags since the format was all over the place.

Thought I avoided the template ones, but must have missed that suggestion was one of them.

PR #9090, opened 2021-12-13
Correct typo: upe to up

bombcar:


Glease:
Please use more descriptive names in PR so we can have a nice looking autogenerated changelog - you do not want a changelog entry like `#9090 up (by bombcar)` right?

bombcar:
True - is there a way to indicate that something is so minor it shouldn’t appear in the changelog (or vice versa indicate that something is important enough to appear in the highlights)

Glease:
To my knowledge, no.

Dream-Master (approved):

Issue #9123, opened 2021-12-15
Unable to load old save


4rchie20:
### Your GTNH Discord Username

4chie20

### Your Pack Version

2.1.2.0

### Your Server

SP

### Type of Server

Single Player

### Your Actions

Using MultiMC.
Trying to load my world on GTNH 2.1.2.0 and getting crash on startup. New world generating and starts with no problems.

### Crash Report

https://paste.ee/p/nA7hp

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the crash still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this crash consistently by following the exact steps I described above.
- [ ] I have asked other people and they confirm they also crash by following the exact steps I described above.

mitchej123:
This seems to be the root cause of the crash.  Not clear why it's throwing a CME.

```
04:26:35] [Client thread/INFO] [NEI Custom Diagram]: Generating diagram group [neicustomdiagram.diagramgroup.gregtech.materialtools]...
[04:26:35] [Client thread/ERROR] [FML]: Exception caught during firing event cpw.mods.fml.common.network.FMLNetworkEvent$ClientConnectedToServerEvent@d0c690c:
java.util.ConcurrentModificationException
	at java.util.ArrayList.forEach(ArrayList.java:1262) ~[?:1.8.0_312]
	at com.github.dcysteine.neicustomdiagram.generators.gregtech5.materialtools.RecipeHandler.initialize(RecipeHandler.java:158) ~[RecipeHandler.class:?]
	at com.github.dcysteine.neicustomdiagram.generators.gregtech5.materialtools.GregTechMaterialTools.generate(GregTechMaterialTools.java:72) ~[GregTechMaterialTools.class:?]
	at com.github.dcysteine.neicustomdiagram.main.Registry.generateDiagramGroups(Registry.java:223) ~[Registry.class:?]
	at com.github.dcysteine.neicustomdiagram.main.NeiCustomDiagram.onClientConnected(NeiCustomDiagram.java:94) ~[NeiCustomDiagram.class:?]
	at cpw.mods.fml.common.eventhandler.ASMEventHandler_1587_NeiCustomDiagram_onClientConnected_ClientConnectedToServerEvent.invoke(.dynamic) ~[?:?]

```

4rchie20:
[...]

PR #9197, opened 2021-12-19
Turn off Forestry worktable modifying recipe list

D-Cysteine:
This fixes the concurrent modification in:
 * https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/9123#issuecomment-994312296

If anyone has any opinions on this change, please let me know! On the one hand, it'd be nice to not have to worry about the crafting recipe list being randomly changed during gameplay. On the other hand, this code has been there for two years, and turning this off may also increase Forestry worktable lag (though we do still cache worktable recipes, for what that's worth).


bombcar:
This caught a warp commit? Maybe it will merge correctly nevertheless.

D-Cysteine:
That commit is already merged into upstream. Not sure why it's being shown as part of this PR (I probably messed something up), but it doesn't show up under changed files, so I think we're okay?

boubou19:
yeah, when you update your branch to sync with upstream, it does that sometimes. Don't worry

Dream-Master (approved):

PR #9281, opened 2021-12-26
Enabled adventure backpack tank tooltip

YannickMG:
As required to enable the tooltip fixed in https://github.com/GTNewHorizons/AdventureBackpack2/pull/4.

### DIFF
--- a/config/adventurebackpack.cfg
+++ b/config/adventurebackpack.cfg
@@ -46,8 +46,8 @@ graphics {
     # Enable rendering for tools in the backpack tool slots [default: true]
     B:"Enable Tools Render"=true
 
-    # Show hovering text on fluid tanks? [default: false]
-    B:"Hovering Text"=false
+    # Show hovering text on fluid tanks? [default: true]
+    B:"Hovering Text"=true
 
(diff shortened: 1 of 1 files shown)


config/adventurebackpack.cfg
```diff
@@ -46,8 +46,8 @@ graphics {
     # Enable rendering for tools in the backpack tool slots [default: true]
     B:"Enable Tools Render"=true
 
-    # Show hovering text on fluid tanks? [default: false]
-    B:"Hovering Text"=false
+    # Show hovering text on fluid tanks? [default: true]
```
bombcar, line 49: Was the default actually changed in the adv backpack code? Ie if you delete the config does it default tru?

bombcar (approved):

YannickMG:
I did enable it in the mod assuming it was disabled due to being a broken
feature. AFAIK when the code runs if no config exists, it's set to the
default.

config/adventurebackpack.cfg
YannickMG: I did enable it in the mod assuming it was disabled due to being a broken feature. AFAIK when the code runs if no config exists, it's set to the default.

PR #9295, opened 2021-12-29
Fix rubber tree rubber dupe

Glease:
No more exploit like this. Verified to work on my server.

scrn

### DIFF
--- a/config/betterbuilderswands.cfg
+++ b/config/betterbuilderswands.cfg
@@ -16,7 +16,7 @@ general {
 
     # Specify forced mappings for what to build from certain blocks.
     # (what you are looking at)=>(number required)*(item required)=>(block to build)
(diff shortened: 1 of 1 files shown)


GlodBlock:
why?

Glease:
Because it's an exploit?

GlodBlock:
i don't that is an exploit. i prefer to call it as a feature. it simply allows you to turn normal rubber tree log into the wet one that can produce resin.

mitchej123:
> i don't that is an exploit. i prefer to call it as a feature. it simply allows you to turn normal rubber tree log into the wet one that can produce resin.

That sounds a lot like an exploit of unintended mechanics

mitchej123 (approved):

PR #9332, opened 2022-01-02
flat bedrock on new chunks

bombcar:
We can have flat bedrock and it would be a nice QoL.

Retroactive flattening bedrock is not on because of the performance hit; that would better be done on a "once scan" perhaps.

### DIFF
--- a/config/cofh/core/common.cfg
+++ b/config/cofh/core/common.cfg
@@ -43,7 +43,7 @@ Version {
 
 World {
     # This will flatten the bedrock layer.
-    B:FlatBedrock=false
+    B:FlatBedrock=true
 
     # The number of layers of bedrock to flatten to. (Max: 8)
     I:FlatBedrockLayers=1
@@ -63,5 +63,4 @@ World {
     }
 
 }
-
 


Dream-Master:
https://discord.com/channels/181078474394566657/465207956745486336/927180082999607386

combusterf:
I've checked and it also affects the nether ceiling, and therefore you can't make a vanilla wither box anymore.
Have you also checked that it doesn't introduce https://github.com/micdoodle8/Galacticraft/issues/3237

Dream-Master:
> I've checked and it also affects the nether ceiling, and therefore you can't make a vanilla wither box anymore. Have you also checked that it doesn't introduce [micdoodle8/Galacticraft#3237](https://github.com/micdoodle8/Galacticraft/issues/3237)

In this case i vote for flat bedrock as well. Witherbox with Bedrock is an exploit.

bombcar:
> Have you also checked that it doesn't introduce [micdoodle8/Galacticraft#3237](https://github.com/micdoodle8/Galacticraft/issues/3237)

let’s test it but if it does looks like a relatively easy fix to backport

Dream-Master:
Screenshot 2022-01-02 22 04 30

Dream-Master (approved):

bombcar:
> > Have you also checked that it doesn't introduce [micdoodle8/Galacticraft#3237](https://github.com/micdoodle8/Galacticraft/issues/3237)
> 
> let’s test it but if it does looks like a relatively easy fix to backport

it doesn't flatten planetary bedrock as far as I can tell (tested by going 500,000 blocks out on a live server)

PR #9386, opened 2022-01-06
note tps lag

bombcar:
Add a note about potentials for lag.


bombcar:
> Perhaps a note saying super/quantum chests do not fall victim to this due to optimisations?

Good idea!

@boubou19 what should we encourage beyond drives and the chests above?

boubou19:
we observed that if you attach a storage to AE with a storage bus, but you do massive I/O externally (EIO/ExU pipes/...) then you force AE to rebuild its cache constantly, and it's an expensive operation, so you create easy lag.

Iirc, Repo did disable the external interactions to the GT super/quantum chests when it's attached to AE. Also, those chests are optimized for AE and act as a AE cell, so lagwise for bulk storage it's good.

repo-alt (approved):

PR #9427, opened 2022-01-09
soul brazier unnerf

ahmetsad:
https://github.com/GTNewHorizons/Thaumic_Exploration/pull/9 recipe should be reverted as it is fixed.


bombcar:
It's not ... completely fixed in multiplayer I think. More testing is needed.

Dream-Master:
@ahmetsad explain why ?

ahmetsad:
soul brazier useless with current recipe. ichorium can be used for pure tear which removes all warp. soul brazier doesn't remove warp but keeps it at the cost of essentia and vis. also taints biome in the process.

Dream-Master:
maybe replace ichorium with something else

ahmetsad:
this was the old recipe so i just removed ichorium. shouldn't be gated as its actual cost is not itself but the maintenance.

repo-alt (approved):
I agree, the recipe should not be expensive. I mean it probably can be made more interesting, but this is way better that the current one. Remember - brazier wards from warp only around itself, so if you have wither level warp, all the fun awaits you a few steps from the base. One can say that brazier is an FPS friendly version of Thaumic Sauna

Dream-Master:
@ahmetsad maybe add some thaumium or void ingots/plates or something more interested and replace with the ichorium one

PR #9509, opened 2022-01-19
Update SGCraft.zs

GTNH-Colen:


leagris:
Any description of this patch?

boubou19:
pushing the assline where it's supposed to belong: endgame

Dream-Master:
Not set computation to 1. it was 32 before

GTNH-Afx237v7:
Make Stargate difficult again!

Dream-Master (approved):

Sphyix:
> Any description of this patch?

Yes, "Shitty patch number 15"
Another nerf just for the sake of it, adding nothing to the gameplay other than more grind.

Stargate was moved to UEV, a completely empty tier that literally exists just for energy hatches, since there's 0 content there.
Instead of trying to add content, a casual nerf making the game more grindy and more time consuming was done, apparently that's how content is made, just nerf everything to oblivion.

Even worse, since ppl were obviously going to complain, this PR was completely nuked from discord.

So I'm here complaining. Recipe should use UHV components instead of UEV, since UEV is an empty tier.
At least there's a bit of content to UHV, so that's the real endgame now.
Making it UEV just makes it so grindy it's not even funny...

Sphyix:
I propose to revert the change and add circuits like biowwares starting from UV and UHV before nerfing this, like, actually adding content for endgame.
This change just makes no sense as it is right now.

GTNH-Colen:
Firstly changes were agreed to by Dream in DMs, boubou was also aware as I discussed it in vc with him. Secondly the VMs have made stargates significantly easier to craft and we can see the direct impact of that given no stargate was crafted in the 5 years preceding the first void miners. Then suddenly within a year or so we have 6ish people who have a stargate. It is a meme item, if this was balance surrounding a normal useful item like the draconic magnet I'd agree.

In short this is a vanity project that due to additional content has become easier, as such it was pushed back to be where it was meant to be: meme status.

GTNH-Colen:
Also side note: still in discord.

Screenshot 2022-01-19 at 23 19 14

PR #9513, opened 2022-01-19
fix stargate recipe

Sphyix:
Revert the change unless circuits for higher tiers are corrected.

Current circuit system goes 3 tiers above the starting circuit.
So example, with crystalprocessors, is:
IV -> LuV -> ZPM -> UV
UHV Crystalprocessors don't exist, next tier is wetware
Wetware goes LuV -> ZPM -> UV -> UHV

Then there's bioware
bioware goes:
ZPM -> UV -> UHV -> UEV -> UIV -> UMV
with UMV actually requiring 8 UIV circuits instead of 2 for some reasons... this is 256 ZPM circuits for 1 UMV

Game is clearly missing a circuit line starting from UV and another one starting from UHV
Adding those can make the UMV circuits actually possible. Stargate nerf should come after those circuits are made possible if they require those.

Other possibility is to make stargate require less power and only up to UEV circuits.

### DIFF
--- a/scripts/SGCraft.zs
+++ b/scripts/SGCraft.zs
@@ -22,10 +22,10 @@ val sgFoil = <dreamcraft:item.StargateShieldingFoil>;
 val sgChevron = <dreamcraft:item.StargateChevron>;
 val sgFrame = <dreamcraft:item.StargateFramePart>;
 val Nanocircuit = <dreamcraft:item.NanoCircuit>;
-val UEVFieldGen = <gregtech:gt.metaitem.01:32679>;
-val UEVPiston = <gregtech:gt.metaitem.01:32649>;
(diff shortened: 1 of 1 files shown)


GTNH-Colen:
So you want to make it even cheaper? Which goes completely against why the stargate exists? Bruh

bombcar:
There's clearly a disconnect between people who want to build a Stargate "to beat GTNH" and the Stargate being a meme.

Sphyix:
> So you want to make it even cheaper? Which goes completely against why the stargate exists? Bruh

Fix the circuits then :)
I accept a dumb line to make them, I don't accept the fact that I need 32 times the amount of circuits compared to any other line to make the quantums

GTNH-Colen:
Stargate was always a meme.

PR #9515, opened 2022-01-19
fix stargate recipe or change circuits

Sphyix:
I don't accept colen closing my PR without giving a valid reason.

Revert the change unless circuits for higher tiers are corrected.

Current circuit system goes 3 tiers above the starting circuit.
So example, with crystalprocessors, is:
IV -> LuV -> ZPM -> UV
UHV Crystalprocessors don't exist, next tier is wetware
Wetware goes LuV -> ZPM -> UV -> UHV

Then there's bioware
bioware goes:
ZPM -> UV -> UHV -> UEV -> UIV -> UMV
with UMV actually requiring 8 UIV circuits instead of 2 for some reasons... this is 256 ZPM circuits for 1 UMV
Just for reference this is 32 times more expensive than normal circuits, UV crystalprocessors only actually requiring 8 IV circuits

Game is clearly missing a circuit line starting from UV and another one starting from UHV
Adding those can make the UMV circuits actually possible. Stargate nerf should come after those circuits are made possible if they require those.

Other possibility is to make stargate require less power and only up to UEV circuits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # GT New Horizons Mod Pack
 
-Version 2.1.2.1 is out 21.12.2021
+Version 2.1.2.2 is out 20.01.2021
 
 [![Build Status](http://jenkins.usrv.eu:8080/buildStatus/icon?job=GTNewHorizons%20Configs)](http://jenkins.usrv.eu:8080/job/GTNewHorizons%20Configs/)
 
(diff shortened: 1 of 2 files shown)


GTNH-Colen:
I literally gave you a reason in discord, you just didn't like it. Dream accepted the PR quite literally minutes beforehand so spamming new PRs undoing work that he has clearly approved is not constructive.

Sphyix:
I'm giving a reason why the change doesn't make sense, game is clearly missing some circuits line before being able to use those in an actual recipe.

Dream-Master:
Again stargate is endgame if things are missing before we need to add stuff

Sphyix:
> Again stargate is endgame if things are missing before we need to add stuff

Yes, but nerf should happen after stuff is added, not before

boubou19:
the game ends with the deepdark portal recipe. Everything after is unfinished and are here as last material and power sinks for those who really want to beat the pack. But because it's in a wip zone, it's subject to see its recipe nerfed over and over, in fact each time we have new content to push back the endgame.

Now it is questionable that indeed we nerf the recipes before adding content, but in this particular case, the pico circuits and the uev components are craftable.

If you want to add x-ware circuit lines to cut down the amount of crafts by a lot (bioware start at zpm and we should at least have two more lines for the circuits) feel free to add it to the game via a PR. I'm sure once it'll be balanced, We will all happily change the sg recipes to accept the new content. But for now, no one wants to implement new content or no one has at least the inspiration to do so.

Sphyix:
So why nerf existing content just for the sake of it if you confirmed endgame content is not finished
Why not wait to nerf it after content is actually added

boubou19:
>Everything after is unfinished and are here as last material and power sinks for those who really want to beat the pack. 

that is why. it must be craftable only with the last tier items availiable. the recipes for sg were designed 3+ years ago, at a time where uhv wasn't correctly implemented. Now uev exists, even if it is empty, it is working 100%. I repeat myself, this is just a place holder for those who wish to beat the game to death. it might sound dumb, but it's like that.

Sphyix:
Then if endgame is uev, make it uev and not use quantum circuits

boubou19:
quantum circuit is the last circuit tier right? and stargate needs the last tier components. so it's coherent with what i said

Sphyix:
No, uev is the last tier and rest of circuits were added just to fix the energy hatches not having a crafting recipe.
I guess removing the recipe was not an accepted fix.

But that doesn't mean those circuits have decent recipes or that they should be used now for other stuff without getting fixed first

boubou19:
last tier stuff doesn't necessarily mean last tier of GT implemented. And the fact you can craft them at uev makes them uev no?

Sphyix:
> last tier stuff doesn't necessarily mean last tier of GT implemented. And the fact you can craft them at uev makes them uev no?

No?
Idk who added the recipe but it being 32 times more expensive than the rest of the circuits, being so recursive that it asks for 256 circuits of the starting tier i don't think is normal

It makes quantum circuits so grindy it's actually a pain to make them

boubou19:
Well you refuse to understand that stargate is a meme, that it must be a pain to do, that is it literally here just for the flex of beating the game to death. From that i think we can stop the discussion here.

(for all the flexers around here: nerfed stargate recipe means that there is a first place to take again)

Sphyix:
As commented on discord

> Changing the recipe of something because there’s going to be later a way to get said thing that’s easier is stupid

mitchej123:
Let's not just encourage more resource spam without more interesting things :)

Dream-Master:
stargate will stay endgame. This items are not needed before
Lets fill up the previous tiers with stuff.

PR #9585, opened 2022-01-26
Quest Description Updates

Steelux8:
- Changed the description of some quests that mentioned machines not being worth building, or not working, when they have been fixed;

There are other quests that are not updated, I'll work on them over time. One of them is the Large Boiler series of quests, since these multis have been changed considerably, but they might be too powerful or too weak so I'm not sure if I should rewrite the description right now.

Additionally, one of the GT++ multis whose quest is locked after the ABS quest, which is the Improved Coke Oven, can be fully built in EV without needing the ABS. I know this because I've done it myself, even if the ABS makes it easier to get all the ingots needed to craft the multi. That said, making this quest depend on another that isn't a requirement in any way is not the way to go, I believe, and this should be changed.


config/betterquesting/DefaultQuests.json
```diff
@@ -156272,7 +156272,7 @@
...
+          "desc:8": "Now that you have some tech, you\u0027ve discovered a way to prospect for ores. This tool only has a chance at finding the ores in a chunk. If you are clever, you can use it to locate ores.\n\nUnfortunately, the tech to make it repairable or rechargeable doesn\u0027t work, so you\u0027ll have to make a new one each time it breaks. It does work better at higher tiers and lasts longer with better materials.\n\nAnd since it doesn\u0027t make a convenient book like the seismic prospector, you\u0027ll have to keep track on your own where the ores are found.\n\nIt can be hard to choose a good material for the Basic Prospector. Here are some good choices for you in their respective tiers, the size searched, and the average amount of veins they can scan:\n\nLV Vanadiumsteel 3x3 - 50\nMV Vanadiumsteel 3x3 - 213\nHV Vibrant Alloy 5x5 - 299\nEV Shadow Metal 7x7 - 1080\nIV Oriharukon 7x7 - 2649\n\nEach chunk successfully scanned costs 100 durability. Each failed chunk costs 25. The average for MV is ~55 per chunk when scanning a large area. As you can see, the prospectors become better and better with each tier, so craft one whenever you think you need it.\n\nGood luck."
```
bombcar, line 156275: t make a convenient book like the seismic prospector, you\u0027ll have to keep track on your own where the ores are found

remove this - visual prospector solves everything imo

config/betterquesting/DefaultQuests.json
```diff
@@ -77350,7 +77350,7 @@
...
...put, depending on which cracking method it\u0027s doing (steam does 50%, hydro does 25% but consumes 20% less hydrogen). Cracked products will result in additional by-products in the distillation tower. It is a smart idea to get this multiblock since you are going to need TONS of industrial TNT from Toluene for advanced tier rockets. \n\nThe 16 coils form 2 vertical rings of 8 on either side of the controller. The casings between the rings can be replaced with a hatch for steam/hydrogen input. Energy hatch and maintenance hatch can replace any of the casings. The input product hatch and output product hatch must replace one of the casings outside of the coils.\n\nHint: Only GT/Railcraft steam can be used, not ic2 steam. You can use a fluid heater to make it instead of a boiler.\n\n§3The circuit goes in the controller.\n\n§3PS: I listed all the cracker recipes on the spreadsheet for you to look at in a convenient format. Now you can easily compare which oil path to take!\n\n§2Bart change the multi so it has locked I/O now, go badger him to fix the I/O on the sides or tell him you want to be able to input the steam/hydrogen from anywhere other than the top if you don\u0027t like it."
```
bombcar, line 77353: Bart change the  should be "Bart changed the"

That line may need rework as Bart is retired so ... ?

bombcar (changes requested):

config/betterquesting/DefaultQuests.json
Steelux8: I removed that portion of the description, since I tested the Oil Cracker when changing it and I was inputting Steam and Hydrogen from the back just fine.

config/betterquesting/DefaultQuests.json
Steelux8: Changed.

bombcar (approved):

PR #9602, opened 2022-01-28
Dedicated Oil Questline

Steelux8:
**### This is a work in progress.**

- First addition of the questline, with the quests placed in the right places (until further changes). The questline begins with the LV pump quest, and then spread out into 5 tiers, from LV to IV.

Working with oil is something that is recommended to basically every new player asking for advice. Despite being one of the most important processing chains in the entire pack for progression, the explanation in oil quests is spread all over the place, across several chapters, and a lot of it is referring to what *can* be done, not what *should* be done. A well-explained path to oil processing, in my eyes, is needed in the questbook to keep people from having to look at Discord or the spreadsheets so much.

The quest descriptions are empty, for now. I've already written almost all of them, I just have to format them into the file. There will be some overlapping quests, here and there, but I focused on letting the player known when they should do something, and not when they can. I'm not sure about what rewards to set, I'd like some feedback on that.

(I accidentally added the quests into a json that was not the most recent, so I manually changed the code. I think it's working properly, I tested it in the full pack before making this PR.)

Here are some screenshots, and an early HV quest example to push the player to use a DT to process oil. **The quest description is in a spoiler tag at the end.**

<details>
  <summary>Quest Description</summary>
  After getting all the oil fractions at once, what is a player supposed to do with them? They'll pile up in the tanks or output hatches, and the Distillation Tower will keep running even if it cannot output something; everything that cannot be output will just be voided instead, and it's in your best interest to avoid this waste. What do, then?

First off, it's important to note that almost everything you get from Oil can be processed into even more things. Whether or not the step is worthwhile is a different story altogether, and most of them aren't. That said, you're free to look through NEI or the spreadsheet to decide what you want to do, and how you want to do it. The charm of a processing chain like this, at least for me, is that it lets you choose your own path and gives you the information to know how to best utilize it.

Second off, Oil is a major source of byproducts, many of which will be needed to create other materials that are necessary, or simply helpful in progression. However, this isn't the only way to obtain most of them, and you might already found other sources, especially if you've dabbled in the Benzene or the Biomass chains. It's imperative that you automate most of them, but the way you do it is up to you.

Now, for the four fractions that you get from regular Oil. Up until now, the power generation needed both Light Fuel and Heavy Fuel processing, and Naphtha already proved to be a good source of Ethylene. What can these be used for, in case you can build more DTs to process the cracked fluids? Roughly speaking, these are the possibilities:

Light Fuel: You will most likely be using most, or all of this to make Diesel. If you do have leftovers, moderately hydrocracking it is a good source of Propene, and lightly hydrocracking is the best source for Octane. These two are important, but only in EV, for epoxy and powergen, specifically;

Heavy Fuel: If only working with regular Oil, a lot of this will also be consumed to make Diesel. However, it's important to get more of it automated, since it can be distilled directly to make three fantastic byproducts: Toluene, Phenol and Benzene. All of them are important for progression, so make sure you pay attention to this recipe. Toluene is especially useful in HV, as you will need many buckets of it to be able to get into EV, when making your first rocket;

Naphtha: You're probably already distilling this for Ethylene. You could switch this recipe into a Distillation Tower, but keep in mind that you need to build a max-height one to cover the entire recipe. However, if you do it, you will have most of the Oil byproducts automated, in more or less quantity, meaning that it's worth it to tackle this challenge when you get comfortable with making lots of Stainless Steel;

Refinery Gas: This is probably the least useful portion, even with DTs. More often than not, it's used as the power source to keep the entire chain working, although you need Gas Turbines instead of Diesel Generators to do this. Alternatively, you can distill it into Methane and/or Helium without cracking it, if you don't need the power.
</details>


Steelux8:
The quest icons are not fully done, amongst other things. I wanted to at least reserve the quest icons, so that they can easily be changed when I get back into this tomorrow.

Dream-Master:
I make it draft so long you working on it

Steelux8:
I thought the same, but I was told by other devs to get it merged so that the quest IDs are reserved.

Dream-Master:
ok . Are the empty Quest hidden ?

Steelux8:
This JSON is corrupted and beyond repair, due to manual changes I did to it. I re-did the questline and made another PR ( #9607).

PR #9607, opened 2022-01-28
New Oil Questline

Steelux8:
- First addition of the questline, with the quests placed in the right places (until further changes). The questline begins with the LV pump quest, and then spread out into 5 tiers, from LV to IV.

Working with oil is something that is recommended to basically every new player asking for advice. Despite being one of the most important processing chains in the entire pack for progression, the explanation in oil quests is spread all over the place, across several chapters, and a lot of it is referring to what can be done, not what should be done. A well-explained path to oil processing, in my eyes, is needed in the questbook to keep people from having to look at Discord or the spreadsheets so much.

Everything is done, except for quest rewards and the secret lategame quests at the end. They're unlocked once all the 5 questlines are finished, and they offer oil-related challenges for the tiers after IV.

Here is a screenshot of the questline:


Steelux8:
I had made a PR like this one yesterday, but I realized that I corrupted the JSON when I manually edited it, after realizing I had started from an old questbook version. I made the questline from scratch and it should be functional now.

Dream-Master:
cant review looking in the file will approve and check ingame

PR #9615, opened 2022-01-29
Remove useless TCon NEI clutter

GTNH-Colen:


mitchej123 (changes requested):
As discussed in discord, check into using INPureCore and expanding the Tcon config.
`It will leave a single of each type of culled items.` is the main benefit!

Dream-Master:
Yes I would suggest it as well. I remember I hide or  un hide there stuff long time ago as well

PR #9628, opened 2022-01-29
Remove plasma cell spam

GTNH-Colen:
Screenshot 2022-01-29 at 21 18 24
Bye bye useless items.

### DIFF
--- a/scripts/GraviSuite.zs
+++ b/scripts/GraviSuite.zs
@@ -1,6 +1,9 @@
 // ---Created by DreamMasterXXL ---
 
 
+// --- Importing things ---
+import mods.nei.NEI;
+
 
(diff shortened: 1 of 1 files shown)


Dream-Master:
not lets add more entry to scripts. Why we not using inpure on this items as well ?

GTNH-Colen:
> not lets add more entry to scripts. Why we not using inpure on this items as well ?

Because this item is totally useless, it's an unfinished feature from like 6 years ago that shouldn't even be here.

Dream-Master:
sure but lets use inpure core for hide items in the future. I want that we add all nei hide recipes there.

Dream-Master:
i am not against the hiding of this cells

PR #9676, opened 2022-02-04
Disable fluid AE components.

GTNH-Colen:


Dream-Master:
ik fluid handling in ec is buggy but isnt it nice to have fluids in ae? Not sure if remove is the best solution.

GTNH-Colen:
People have wanted this removed for ages, it's a blight on the modpack and only leads to people getting frustrated once they realise how bad it is.

bombcar:
Having it default not enabled for servers is the way I think, then server operators can turn on.

But it is really hard to make it work even remotely usably.

GTNH-Colen:
Not to mention the billion dupe bugs associated with it, just needs yeeting.

Nxer:
We should not kill a people because of just a  little cold he had .
If we follow this logic , we should remove all the mods because they all have Serious bugs .
There is just some bugfixes we need to do and not remove , right ?

Elisis:
> There is just some bugfixes we need to do and not remove , right ?

It's systemic with the mod. It's not that simple. Plus, people can just use P2P which is actually good

bombcar:
The reality is the mod is written in Scala and nobody has both the knowledge and the drive to fix it (if it even can be).

GlodBlock:
ec2 in nh fork has been rewritten with java, don't need much scala knowledge to read it. though it's coded brilliantly, it is still possible to fix and improve its fluid system. impact has dealed with ec2 bugs and performance problems iirc

bombcar:
Do you know where we could find the impact ec2 repo

GlodBlock:
https://github.com/GT-IMPACT/ExtraCells2

4gname:
> https://github.com/GT-IMPACT/ExtraCells2

I didn't do anything particularly interesting.

- Public fixes were implemented.
- My work with the Fluid Terminal GUI.
- Fluid Interface now save the names of liquids, not their IDs.
- Lots of shitty code (when I couldn't code yet).

Nxer:
> > There is just some bugfixes we need to do and not remove , right ?
> 
> It's systemic with the mod. It's not that simple. Plus, people can just use P2P which is actually good

So we need a new mod which can replace it or a bugfixed version , not just remove . 
And , trying to control players' play is not a clever idea .

Elisis:
> So we need a new mod which can replace it or a bugfixed version , not just remove .

It is called P2P. There are very few circumstances where EC2 trumps P2P anyhow. Why should we not remove a currently broken mod that already has alternatives?

Nxer:
> > So we need a new mod which can replace it or a bugfixed version , not just remove .
> 
> It is called P2P. There are very few circumstances where EC2 trumps P2P anyhow. Why should we not remove a currently broken mod that already has alternatives?

Nothing .
EC2 is crap , but fluid AE is necessary . Just please don't just disable fluid AE .

Nxer:
> > So we need a new mod which can replace it or a bugfixed version , not just remove .
> 
> It is called P2P. There are very few circumstances where EC2 trumps P2P anyhow. Why should we not remove a currently broken mod that already has alternatives?

Nothing .
EC2 is crap , but fluid AE is necessary . Just please don't just disable fluid AE .

joshaup1999:
There is already fluid support in base ae. Fluid p2p is superior. I would like to see the fluid gui work with p2p. That way there’s no complaints. Only thing is, it’ll break a lot of existing bases if deleted

yukieiji:
I think that even if we want to remove EC2, we should first provide players with an alternative.
Even if there is no alternative and we decide to remove it, if we don't provide a transition period, the systems of many player bases will break down due to update.

mitchej123:
Might I suggest one of the people complaining about the removal step up and fix the lag and bugs with the mod, otherwise kindly stop complaining?  Without a maintainer it's getting yeeted.

boubou19 (approved):

mitchej123:
> We should not kill a people because of just a little cold he had . If we follow this logic , we should remove all the mods because they all have Serious bugs . There is just some bugfixes we need to do and not remove , right ?

I agree.  Any mod with serious bugs and no maintainer should be removed.

Unless you're volunteering to step up and maintain it?

Dream-Master:
@mitchej123 you need to remove the code not only the recipes. And we need to look if we alternative ways to manageing fluids in ae

mitchej123:
@Dream-Master We're phasing it.  First step is removing recipes, and give people time to migrate away.  After that we'll start gutting the mod until it's just "Bigger Cells" (at which point, why not just roll those into ae2 themselves :thinking: )

Dream-Master:
and maybe add some fluid cells to ae2 that not duping fluids

repo-alt:
Recently there appeared some volunteers to fix/rewrite EC2, so there is still hope for it. For now we may just disable the recipes so that existing setups would not immediately break.

mitchej123:
That's the plan repo!

PR #9704, opened 2022-02-08
Add a bypass for smeltery crafting tasks

Glease:
Player can now choose to go with village raiding to bypass the crafting tasks, but the XOR quest logic would then forbid them completing the other quest, voiding their quest rewards.


Glease:
It's basically converting the old smeltery quest into 3 quests, 2 as prereq and 1 as final reminder. The two prereq quests are mutually exclusive with each other via XOR quest logic (similar to the personal dim vs loot choice). One uses crafting quests and has quest rewards, the other use retrieval quests and doesn't have quest rewards. Completing either of these would block the completion of the other and unlock the final reminder quest (which is a checkbox quest). This final reminder quest contains a bit of info on how smeltery work in our pack, doesn't have quest rewards and is the prereq for all other quests that used to depend on the old smeltery quest.

No other quests are modified. No sure why there is a huge amount of quest ID shifting tbh.

boubou19:
i'm not a fan of xor quests, as it prevents you from reaching the 100% completion tab (green tick on the tab)

mitchej123:
> i'm not a fan of xor quests, as it prevents you from reaching the 100% completion tab (green tick on the tab)

Sounds like a separate issue we should address.  Make it require 1 of the XOR'd quests?

Dream-Master:
> i'm not a fan of xor quests, as it prevents you from reaching the 100% completion tab (green tick on the tab)

i am ok with it if you reach the goal when you run the crafting way. But if the check box quest stay open maybe we need to look for another solution.

boubou19:
nvm: the xor quests are already excluded from progression in 2.1.2.0: 
image

Dream-Master (approved):

PR #9711, opened 2022-02-09
Optional tasks part1

miozune:


Steelux8:
Add screenshots of some of the quests as displayed in-game, if you can. It's much too difficult to know from the json file, and testing this on a dev environment still crashes the game, would have to load the full pack.

PR #9871, opened 2022-03-05
Disable useless EBF gas crap

GTNH-Colen:
- Cuts down on NEI bloat (like 200+ pages of recipes...)
- These overwrite existing recipes, so you can't set a recipe with radon to a specific time...

Need to re-add some argon recipes and remove some others, but neon, krypton, Og are all utterly useless for EBFing and afaik no one uses them. Radon left as is.

### DIFF
--- a/config/bartworks.cfg
+++ b/config/bartworks.cfg
@@ -418,7 +418,7 @@ system {
     B:"Disable BioLab"=false
 
     # This switch disables extra gas recipes for the EBF, i.e. Xenon instead of Nitrogen
-    B:"Disable Extra Gases for EBF"=false
+    B:"Disable Extra Gases for EBF"=true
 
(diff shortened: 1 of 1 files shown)


Pxx500:
With the new Planetary Gas Siphon there recipes may be useful. Maybe instead of deleting them they shoud be merged? One page shows all possible gasses with % time reduction for every gas?

Chaoschaot232:
Wouldn't it much better to have a zero value gas as placeholder and on pressing the NEI lookup key getting redirected to the EBF where you should be able to get a list of all recipies?

This could reduce the NEI pages a lot in general to as long as it's worth todo this.

Dream-Master:
disable the cfg make the recipe not using these fluids anymore. Hiding it in nei is a better option.

PR #9925, opened 2022-03-11
Enderium > Dark Steel for Basic EiO Stuff

ScriptedPiky:
These are incredibly basic machinery that is made around the same tier as you'd get at around LV-MV, though it so happened to apparently be gated at EV~


Dream-Master:
Will check later

Dream-Master (approved):

PR #9933, opened 2022-03-12
Add pyrolyse quest

Elisis:
image
image
image


Elisis:
I PRed it to new-quest to avoid any shuffling around of branches

Probably would be best to merge this before new-quest, if we do merge it

miozune:
Doesn't making both tasks optional lets you complete the quest immediately?

Elisis:
Couldn't think of a better way of including both machines. There're no rewards however so it's not like it benefits the player much

Glease:
Set taskLogic to OR and use normal retreival task.

Elisis:
Oh

Dream-Master (approved):

PR #10068, opened 2022-03-30
Even More Meteors & Adjustments

ScriptedPiky:
I'm going to add more meteors and adjust some; the results of me testing them and some others testing is very fruitful and promising so far!

Also going to add some "unique" meteors with a weight over 1000 for experimentation via Terrae and Orbis Terrae alchemy


ScriptedPiky:
I have tested every single of the ores to see if they worked or not. Also, this was tested on BM this PR: https://github.com/GTNewHorizons/BloodMagic/commit/595bd797fd961de2ba533d02646079d783ea3dd2
It is HIGHLY IMPORTANT that people briefly understand about this PR changing weights, as a lot of these other meteors wouldn't work without this specific Meteor Weight change. Otherwise, this'll work on the live versions just fine for the most part.

Dream-Master (approved):

PR #10142, opened 2022-04-13
Liquid Air prospecting planet buffs

ScriptedPiky:
Callisto: 200
Pluto: 300
Barnarde: 400

Instead of the awful original rates where it would be silly otherwise to go for this stuff, it's now a lot more worth while to go for


Dream-Master (approved):

Dream-Master:
@ScriptedPiky hope you know what you are doing and if this is balanced.

PR #10161, opened 2022-04-16
Increase max distance between traversable anchors

Elisis:
The 128 limitation actually presents issues with larger bases

### DIFF
--- a/config/enderio/EnderIO.cfg
+++ b/config/enderio/EnderIO.cfg
@@ -27,7 +27,7 @@
     I:travelAnchorCooldown=40
 
     # Maximum number of blocks that can be traveled from one travel anchor to another.
-    I:travelAnchorMaxDistance=128
+    I:travelAnchorMaxDistance=512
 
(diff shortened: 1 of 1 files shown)


Dream-Master:
128 to 512 is a huge step. Not sure about balance. What the others think?

Ethryan:
maybe 256 instead?

Dream-Master:
If 512 Is ok I am ok with this value as well. Just want to ask

boubou19:
To be honest i'd like it to be 512 blocks, as Elisis said, huge bases (like mine) are a pain in the ass with small distance of anchors. I even prefer to fly than use them because of that.

GTNHAfx237v7:
You playing on 32 render distance?

Elisis:
> You playing on 32 render distance?

For me, at least, it doesn't really matter what I set it to. The problem is still there

Dream-Master (approved):

Issue #7834, opened 2021-04-20
Fluid level emitters are stupendously laggy


RubiksImplosion:
#### Which modpack version are you using?
Zeta 2.1.0.5
#
#### If in multiplayer; On which server does this happen? 
Private 2-player server
#
#### What did you try to do, and what did you expect to happen?
We had a dozen or so fluid level emitters that  were controlling fluid levels
#
#### What happened instead? (Attach screenshots if needed)
They caused unplayable amounts of client-side TPS lag
#
#### What do you suggest instead/what changes do you propose?
Either fix them or disable them, since they will cause unplayable amounts of lag
#
#### What is your GTNH Discord username?
RubiksImplosion#7493

eigenraven:
Fixed in dev

Issue #7699, opened 2021-03-24
ExtraCells cant correct show fluid 


Pnc6lx:
in ae2 fluid terminal, cant show fluid over 2MegaB correctly. beacuse 2MegaB is 2G mb, whitch means 2^31-1, however you can storge much more fluid than 2MegaB in it.

Pnc6lx:
![WSLA7`N MR (Z1FF$}BZT7](https://user-images.githubusercontent.com/49830847/112266215-ab3c8380-8cae-11eb-945e-e6a6d39d70f9.JPG)
![5OP9MSD$K`$L}GG52E {032](https://user-images.githubusercontent.com/49830847/112266235-b1cafb00-8cae-11eb-8b02-38ff0a7eb4b7.JPG)
![TQ`JSTZBJZLKN}VLNC64MQ0](https://user-images.githubusercontent.com/49830847/112266243-b4c5eb80-8cae-11eb-8ddf-b245bf66c7c9.JPG)
![I(656CZ@OG~Z_2WF BUG_WC](https://user-images.githubusercontent.com/49830847/112266600-33228d80-8caf-11eb-81d7-ca527ff60b8d.JPG)
![Y3}{KFE)N APC}_Z_}BB PJ](https://user-images.githubusercontent.com/49830847/112266610-3584e780-8caf-11eb-981f-c3816dae43b1.JPG)
this fluids are much more than 2MegaB, and N2 is over 8 stucks of Bedrockium drum.

Issue #6330, opened 2020-06-25
bedrockium drums duping liuids if used with extracells interface


viljar290:
#### Which modpack version are you using?
2.0.9.0
#
#### If in multiplayer; On which server does this happen? 
ETA
#
#### What did you try to do, and what did you expect to happen?
Tried to take all the oxygen from my fluid storage into a bedrockium drum to merge the separate storages filled with them.
expected it to take all my oxygen and to put it in the drum.

#
#### What happened instead? (Attach screenshots if needed)
On putting bedrockium drum into the interface it fills itself to the amount that was in the storage. storage does not lose any so it doubles the amount of liquid.
#
#### What do you suggest instead/what changes do you propose?
if possible fix whatever code bit causes it to duplicate rather than fill as it does with other containers. if not possible disable bedrockium drum from working in ae2 fluid interface.

GTNH-Afx237v7:
Similar to [6090](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6090#issuecomment-626750470) 
Also belongs to [6099](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6099#issuecomment-645963720)

eigenraven:
Fixed in dev for a while

Issue #5915, opened 2020-04-14
Getting fluid from nothing


PredVesTNiK2142:
#### Which modpack version are you using?
2.0.8.8
#
#### What did you try to do, and what did you expect to happen?
To get any liquid in the game, we need Water Reservoir, ME Fluid Export Bus, ME Fluid Storage Bus and a sample of the liquid you need to get. We connect Storage to Reservoir. We connect Export to any storage and indicate the liquid that you want to receive. That's all :) The endless source of fluid is ready.

P.S. Works anywhere in the AE system
P.S.S. Logically, only water should be generated in this way.
#

draknyte1:
ExtraCells is broken as fuck. 🤦🏻‍♀️

quedral:
not surprised. extracells has been buggy/laggy like this for years and there are no fixes unless you code the whole damn thing from ground.

leagris:
I'd add that Extracells is wholly useful when used **responsibly**. But, as it is a dupes nest, even the most adult and mature players have to keep an eye not to accidentally dupe fluids though. If you are a server admin, you'd better review your whitelist applications so you can trust your players not to abuse the gazillion flaws in all the mods.

draknyte1:
Is there any real reason we even still use EC2?

I’m sure the handful of useful bits could just be rewritten and put into a new mod.

leagris:
EC2 is very convenient when autocrafting on-demand recipes with fluids interfaces and crafting patterns. Example sending 36L of soldering alloy fluid with items to the assembler or circuit assembler.

EC2 is a convenient mod when used responsibly. If you feel the gut to provide feature parity with AE2 integration ore rewrite/fix this whole mod, then go for it. Although I am used to consider all mods for educated mature players only. (the kind of player who will take mods for their good, will not try to abuse flaws, avoid overloading servers… everything that need maturity and education).

boubou19:
maybe i'm doing it wrong but can't get this to work in 2803

Prometheus0000:
If someone wants to rewrite it or break out the most useful parts (though it doesn't seem to have a whole lot to it to begin with, being an addon, so I dunno if that would be all that different) they have my full support. The problem is getting someone willing to do that of course. I do wonder if maybe some of these problems are fixed in a later version? And I believe AE2 itself added some kind of fluid support too later on. Maybe it would be easier to backport changes instead of rewriting the whole thing.

[...]

PR #10179, opened 2022-04-20
Ec2

Dream-Master:
Add back ec2 fluid support since a few fixes by @kubasz was added (only for zeta tesing)

### DIFF
--- a/scripts/Extra-Cells.zs
+++ b/scripts/Extra-Cells.zs
@@ -6,7 +6,6 @@
 
 
 import mods.gregtech.Assembler;
-import minetweaker.item.IItemStack;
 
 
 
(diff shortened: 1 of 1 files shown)


GTNH-Colen (changes requested):
I don't think this is a good idea, even with a few bugs fixed there are so many still left and not to mention the horrific performance issues of the mod. Best left disabled.

Dream-Master:
@GTNH-Colen make a List which stuff is broken so I can ask @kubasz to try fix it.

GTNH-Colen:
Excessive lag, not just from fluid level emitters but you can see it here https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7834

Bad handling of fluids > 2MB https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7699
Dupe: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/6330
Another dupe: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/5915

Some have been censored because obviously we do not want dupes in public. Everything provided by the mod is done by others, with better performance and scalability. I just do not see the advantage of enabling it again when people will inevitably just bitch more.

kubasz:
@GTNH-Colen 
> Excessive lag, not just from fluid level emitters but you can see it here #7834. Multiple censored dupe bugs, see private repos and everything provided by the mod is done by others, with better performance and scalability. I just do not see the advantage of enabling it again when people will inevitably just bitch more.

The fluid emitter lag was particularly bad as it sent a chunk update every time a fluid amount changed, this is now fixed to only send an update when the amount change triggers a redstone signal strength change just like the ae2 item level emitters. The import/export buses never set their tickrate to IDLE, this is now also fixed and they should behave exactly like ae2 item buses. I did a few tweaks to the fluid storage bus, but it has to trigger updates on its attached block just like the item storage bus, so that's probably not going to get better. Most dupe bugs were related to using the wrong cell api, I have switched it over to the proper one and tested it extensively with large GT cells and now xu2 drums, and couldn't find any way to dupe or void liquids so far. I think it can be a nice alternative to eio ender fluid conduit networks, with a terminal to view/access fluids, and storage monitors to display levels anywhere in a factory - both of which are not replicated by other mods afaik.

As for general performance, I removed some particularly bad code that was causing lots of allocations, other than that a savefile or instructions for a specific known-laggy setup would be necessary to profile the game and pinpoint what part of the code is causing issues, and I'd be happy to fix them if possible. (or even just a opis/visualvm profiled stacktrace could work)

Chaoschaot232:
@kubasz - I like how you fixed a lot of EC2 stuff :3 - I can't wait to play with my ME later on with all these fixes :)

Maybe the player mentioned in #7834 can also test what you have changed as they have reported it. Getting feedback from them with there network could be a good reference ;)

kubasz:
@GTNH-Colen Any chance of revisiting your opinion after the changes in <https://github.com/GTNewHorizons/ExtraCells2/pull/55>, <https://github.com/GTNewHorizons/ExtraCells2/pull/54>, <https://github.com/GTNewHorizons/ExtraCells2/pull/53>, <https://github.com/GTNewHorizons/ExtraCells2/pull/52>? If any more bugs or laggy code is found after all these changes, they can be fixed, and the only realistic way to do that is having players test EC2 in the dev pack.

Connor-Colenso (approved):
Worth a shot.

joshaup1999:
Is this in the zeta test pack atm?

PR #10263, opened 2022-05-03
Renewable Power Questline

Steelux8:
- Added quests and descriptions for a questline, covering LV to IV, focused on powergen and chemical byproducts obtained from renewable items such as wood, saplings and fish.

(Tasks and rewards are not in yet.)


Steelux8:
Questline is 100% finished now.

Dream-Master (approved):

PR #10288, opened 2022-05-06
Tiering Logistics Pipes to MV

Steelux8:
- Changed LP recipes to not need HV machines or components, except for specific upgraded items such as high-tier Logistics Chassis, which still need HV and sometimes above.

Logistics Pipes is an automation mod that was added to GTNH some months ago, but it has seen very little use, especially as a main automation mod. This is because it requires HV to set up proper automation with it, and in the next tier, EV, the player gets access to the faster, less laggy and vastly superior AE2. As such, there's very little room to use this setup before it gets replaced by another.

To combat this, I've changed a lot of recipes, so that full automation can be set up in MV, in a similar fashion to ProjectRed, a very similar mod to Logistics Pipes. Some upgrades to important pipes are still locked to higher tiers, but it should be possible to automate crafting, processing and others with LP in the MV tier, after this change. It is still an expensive setup to craft for the tier, so I think the challenge is well-balanced, considering that this automation will be slow, albeit powerful.

### DIFF
--- a/scripts/Logistics-Pipes.zs
+++ b/scripts/Logistics-Pipes.zs
@@ -143,9 +143,9 @@ Assembler.addRecipe(<LogisticsPipes:item.logisticsParts:5> * 2, [<BuildCraft|Sil
 Assembler.addRecipe(<LogisticsPipes:item.logisticsParts:5> * 2, [<BuildCraft|Silicon:redstoneChipset:2> * 2, <IC2:upgradeModule>, <ore:circuitGood>, <appliedenergistics2:item.ItemMultiMaterial:22>, <ore:boltBlackSteel> * 4, <gregtech:gt.integrated_circuit:18> * 0], <liquid:molten.solderingalloy> * 36, 900, 120);
 
 // --- Diamond Upgrade Chip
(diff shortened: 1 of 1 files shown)


glowredman (approved):

Glease (changes requested):
This unusual preference on diamond bolt is too terrible. With your recipe, 1 single diamond upgrade chip would need 8 diamond, and 2150.4 second of LV machine processing time. Of course, in theory you could recycle the leftover diamond dust and this will cost less raw diamond input, or you could use the hand crafting recipe to speed things up, which require flawless diamond as input. Whether these are practical depends on the state of your stockpile, but I guess people will avoid it to as much as possible.

I admit end steel is a terrible choice, as it gates this HV stuff to end EV, but diamond is just an equally terrible one.

Steelux8:
> This unusual preference on diamond bolt is too terrible. With your recipe, 1 single diamond upgrade chip would need 8 diamond, and 2150.4 second of LV machine processing time. Of course, in theory you could recycle the leftover diamond dust and this will cost less raw diamond input, or you could use the hand crafting recipe to speed things up, which require flawless diamond as input. Whether these are practical depends on the state of your stockpile, but I guess people will avoid it to as much as possible.
> 
> I admit end steel is a terrible choice, as it gates this HV stuff to end EV, but diamond is just an equally terrible one.

Changed the diamond bolts to red steel.

Glease (approved):

PR #10293, opened 2022-05-07
Make InGameInfo.xml more readable

glowredman:
- Make formatting codes more verbose (e.g. `$0` -> `{black}`)
- Remove inconsistencies (remove double whitespaces, capitalize some letters)

### DIFF
--- a/config/InGameInfo.xml
+++ b/config/InGameInfo.xml
@@ -2,8 +2,7 @@
 <config>
     <lines at="topleft">
         <line>
-            <str>{pingicon}$6{username}$f </str>
-            <str>FPS: </str>
+            <str>{pingicon}{gold}{username} {white}FPS: </str>
             <op>
                 <str>gt</str>
                 <var>fps</var>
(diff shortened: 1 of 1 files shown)


mitchej123:
Dream won't go for warp. We had it before but removed it

glowredman:
Alright, removed it again

mitchej123:
Looks fine -- do you have a screenshot of it so we can see it visually?

glowredman:
It should look like before:

mitchej123 (approved):

Chaoschaot232:
> Dream won't go for warp. We had it before but removed it

What is his problem with adding warp level here? I would also like to add the bloodmagic stuff but I am not sure how to add it properly. The only thing I added for was the real life time which could be added too ^^

chochem:
the reason is simple: there are ingame methods for warp and bm infos. you are supposed to use them. they are part of the game.

Chaoschaot232:
I see, make sense then :)

Connor-Colenso:
What does the -1 indicate before the users name? I always assumed this was an error. Maybe you know after looking into it @glowredman? Also the random ( after the light level?

glowredman:
The "-1" indicates the IGI can't get the player's ping (see [com.github.lunatrius.ingameinfo.tag.TagMisc$Ping](https://github.com/GTNewHorizons/InGame-Info-XML/blob/master/src/main/java/com/github/lunatrius/ingameinfo/tag/TagMisc.java#L141-L152)).

The "(" is supposed to be there, although the closing bracket is missing (which nobody noticed...)

PR #10322, opened 2022-05-11
change scripts and random things cfg to match with new version

Dream-Master:
If we decided to change to Random Things from GT Mega fork we need to merge this pr.


Chaoschaot232:
What is the difference between our RT and that one from GTM? Would be nice to know since I have no idea what is in general the Diff between our and there pack 🤔

Dream-Master:
If I understand sir fell right this pr is no longer necessary

PR #10345, opened 2022-05-15
Remove IV Repellator from IV lootbag

Elisis:


Dream-Master:
Add something back so the balance and chance of the loot bag not changed

leagris:
> _ No description provided. _

A few words, mention of discussion and reason for this change would give ground to this PR to stand on.

Elisis:
Ah, my bad. It's because it could be disassembled for the emitters which made it kinda OP to get

Dream-Master (approved):

leagris:
Or as already envisioned by just laying a persistent NBT tag like `gift` and having disassembler refuse to disassemble loot tagged stuffs, that would be enough.

Having looted magines and items with special properties not present in regularly crafted version, would be something akin to enhance gameplay by enabling many different balance adjustments levers like:

- Cannot disassemble.
- May break after some time.
- May have different efficiency, speed than its crafted counterpart.
- May have different durability than its crafted counterpart.
- May disassemble into specific listed parts (legit but limited re-use of the crafting ingredients NBTs). Makes it possible to create loots of "unknown technology" machines crates that can be turned to specific components or raw materials if disassembled or crushed.

Most of this re-use of previous implementation of crafting ingredients NBT but for looted items/machines from Gregtech would give more possibilities while not requiring additional developments. (maybe some work to enter those new loots with designated ingredients).

Other items or machines, since crafted without NBT ingredient, would continue to benefit optimised disassembly and reduced payload. Only specific loots would have extra data.

boubou19:
nbt tags for special ways of obtention sucks: you turn your stuff to be unstackable, which causes a lot of issues in automation.

Steelux8:
This was approved... IV lootbags are one of the few helping hands in the huge grind that is the IV tier, it keeps getting harder and longer, and lootbag nerfs contribute to that.

PR #10412, opened 2022-05-29
Extreme Industrial Greenhouse quest

kuba6000:
image
image
image
image


glowredman:
Also it might be a good idea to have this quest be displayed in the farming tab. Multiblock tab is fine but imo IV tab is not really fitting.

kuba6000:
Well the farming station quest is in HV tier tab, thats why i put it there as well

glowredman:
That I can understand because (if I remember correctly) it is grouped together with the more advanced EnderIO stuff and maybe even predates the farming tab. Although it too might be more fitting in the farming tab

chochem:
the enderio one is more versatile in some ways and does tree chopping for example. That is not exactly a farming/cooking thing.

kuba6000:
Ok, i moved it to farming tab

GTNHAfx237v7:
What does it mean „you can’t Farm two crop types at the Same Time?“ I remember you posting screenshots with up two four different types in there. Or should it say „you cannot harvest two or more crop types at the same time?“

kuba6000:
It means that you can't farm IC2 crop and normal one at the same time.

kuba6000:
I changed it to be more clear
image

GTNHAfx237v7:
Much better, thx

Dream-Master (approved):

PR #10586, opened 2022-06-29
Filling Out Information and Tasks on Nuclear Quests

Steelux8:
This is a work in progress. I made this PR before finishing the quest changes to make sure there weren't any strange line changes.

This only changes existing quests, so there should no problem with conflicting quest IDs.


Steelux8:
Changes to the nuclear questline are finalized. They are only description and quest task changes, so there should be no conflicts or other problems.

Dream-Master (approved):

PR #10643, opened 2022-07-06
Changed visability from quests with "SECRET" to "NORMAL"

Chaoschaot232:
Like discussed in Discord I sendout this PR which removes the "SECRET" setting of any quests so that now all quests are normaly visable when in view mode.

**Out of view mode they will be hidden until requirements are met!!!**


Connor-Colenso (approved):
Very much needed.

Dream-Master:
Nope I am against it

Chaoschaot232:
What is the problem with this change?

Connor-Colenso:
> Nope I am against it

Hiding information from the user serves no purpose. People just go into edit mode to bypass it. It needs to end, just like we did for the assline recipes.

Dream-Master (changes requested):
They hidden because they are a special line (legendary quests)

Dream-Master:
If this go through I go delete the quest line

Prometheus0000:
1) There's apparently 217 of them, but I only recall about 20-30 in the legendary tab, so it's not just this tab (unless a ton was added)
2) Many people would be happy to have it removed, so long as there's a way to get draconic cores

Chaoschaot232:
> 1. There's apparently 217 of them, but I only recall about 20-30 in the legendary tab, so it's not just this tab (unless a ton was added)
> 2. Many people would be happy to have it removed, so long as there's a way to get draconic cores

add arround 30 quests from Endgame Goals tab to this.

Dream-Master:
> this PR is not thought out. These are secret for various different reasons.
> 
> the bee ones can totally be made visible not sure why this is not the case.
> 
> I can also understand this for the endgame goal tab. So that people can be awed by what's there.
> 
> But making the legendary ones visible is silly. I am against that.
> 
> what are the others? wither x2, cinder, a few others like that? also should probably stay hidden

Exactly what I mean.

Chaoschaot232:
Since PR #10675 was merged this one is obsolet and will be closed.

PR #10645, opened 2022-07-06
Changed visability for all categories from NORMAL to UNLOCKED

Chaoschaot232:
Compared to #10643 this PR changes how the categories are handled within the questbook. This means that no quest was touched!

What I did, was to set each category from **NORMAL** to **UNLOCKED** excluding **Adn So, It Begins**.

Here is one pic of the old view:

image

Here is one pic of the new view, with a fresh started world and no view mode:

image

Here is a a few pics of partially solved quests (over cmd) to show that unlockings works and with view mode on:

image
image

Don't want to show to much but this change makes the book easier to read and focus on what quests are reveled. I have seen many BQ based packs which have choosen this path and looking at that 1.12.2 Modpack SevTech: Ages it is the same way only without the ability to hide NEI stuff and thrwo out items which you can't handel currently (configurable in the correct mod config from what I know).

Even if I wouldn't like that unknown items are getting thrown out of your inventory, smth. like Progression would be interesting for this pack ... as a side node.


Chaoschaot232:
> As I pointed out on Discord, I'm against this change. Yes this is a good change for newbies. Providing limited information will hide unrelated stuff and players can focus on what to do now. But this is bad for skilled players - those who are used to expert packs, have played with any version of GT or GT-like mods, and second or more runner for GTNH. They tend to plan for future expandability or something, by using NEI or quests. And quests can provide grouped information, which is better than NEI showing all tier recipes. So what I suggest instead is:
> 
> 1. If view mode is disabled, UNLOCKED and still locked questlines are not visible, like you changed.
> 2. If view mode is enabled, all questlines are visible, as if they were set to NORMAL.
> 
> This will satisfy both requirements I listed. It might not be able to be achieved with current mod, but we can change it in code.

I fully understand you and yes it is hard to make it suitable for both groups speaking of group A) which wants to have al quests visable all the time - group B) which wants a solution like this one -- group C) which wants to discover every pice by themself with less hints as possible, which means no "view mode" and no "edit mode" usage even if they are aware of this.

The main problem is after all the category "Legendary Tasks/Quests" where I can also fully understand all groups. Finally we must find a suitable solution for every group even if this needs some code change within BQ itself to be achived.

PR #10675, opened 2022-07-09
Changed visability for all categories from NORMAL to UNLOCKED

Chaoschaot232:
Re-Created due to deletion of dev2 branche - Reference: #10645 

Compared to #10643 this PR changes how the categories are handled within the questbook. This means that no quest was touched!

What I did, was to set each category from **NORMAL** to **UNLOCKED** excluding **Adn So, It Begins**.

Here is one pic of the old view:

image

Here is one pic of the new view, with a fresh started world and no view mode:

image

Here is a a few pics of partially solved quests (over cmd) to show that unlockings works and with view mode on:

image
image

Don't want to show to much but this change makes the book easier to read and focus on what quests are reveled. I have seen many BQ based packs which have choosen this path and looking at that 1.12.2 Modpack SevTech: Ages it is the same way only without the ability to hide NEI stuff and thrwo out items which you can't handel currently (configurable in the correct mod config from what I know).

Even if I wouldn't like that unknown items are getting thrown out of your inventory, smth. like Progression would be interesting for this pack ... as a side node.


Connor-Colenso:
After significant thought on this knowing it may annoy a few endgame users who "know" everything already I think personally this is a much cleaner setup for new players and prevents information overload.

Ethryan:
This seems to have been merged into the dev branch, it isn't there in the master branch and thus isn't implemented in the pack.

@Dream-Master

PR #10682, opened 2022-07-10
16 x 16 Sprite Revisions for GT:NH

FDAapproved:
In this pull request I am proposing several non 16 x 16 textures be replaced with more vanilla-esque versions (natura, sleeping bag). If these texture files are in the wrong directory, feel free to move them to somewhere more appropriate before merging.


eigenraven (changes requested):
These should go into the respective mod repos instead of into the pack as resource overrides: <https://github.com/GTNewHorizons/Natura/> and <https://github.com/GTNewHorizons/SleepingBags>

FDAapproved:
Alright; in that case I'll re-consolidate as separate merges for those repositories and get back to you.

Issue #9394, opened 2022-01-07
Cable Anchor recipe is weird


bombcar:
### Your GTNH Discord Username

bombcar#1042

### Your Pack Version

2.1.2.2

### Your Proposal

![image](https://user-images.githubusercontent.com/533931/148593087-3ec116a3-736d-4525-b2e9-b570a110beda.png)

this is just a variation of the original, replacing certus quartz blade with the knife

### Your Goal

it's thematically weird to be building the advanced alien technology and then having to bust out a knife to whittle ingots.

### Your Vision

Either a shapeless recipe to convert bolts to anchors (buffs the number of anchors per ingot, anyone cares?) - or a simple assembler recipe for two bolts to an anchor.

I don't know if anchors/facades are used before AE in anything.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

Dream-Master:
wiremil maybe better and a hand recipe with wirecutter and plates maybe but with a different recipe shape

bombcar:
There are 13,000 facade recipes also hmmm

GTNH-Colen:
Just add a shapeless iron bolt -> anchor. This is effectively all the item is, just a bolt placed in the world.

GTNH-Afx237v7:
Yeah don’t overcomplicate it. Iron Bolt to anchor is fine. Don’t think we need three different recipes for the same thing.

chochem:
no nothing yet

PR #10688, opened 2022-07-12
Change Cable Anchor to use bolts

Ethryan:
- #9394

Thoughts?


GTNHAfx237v7:
Sure why not. Better materials should give more cable anchors.
Looks good to me!

miozune (approved):

PR #10713, opened 2022-07-15
Halve all Molecular Transformer recipe times

xSkewer:
Originally the MT had a 2x speed boost that was unintended. Even with this boost, the multi wasn't considered very good, but it got fixed since it was still unintended. This effectively brings back that speed boost as a buff.

Sunnarium will be changed separately.

### DIFF
--- a/config/AdvancedSolarPanel_MTRecipes.cfg
+++ b/config/AdvancedSolarPanel_MTRecipes.cfg
@@ -11,23 +11,23 @@ version=1.0
 ##################################################################################################
 
 #redstone -> ruby
-minecraft:redstone:1; gregtech:gt.metaitem.01-2502:1; 5000000
+minecraft:redstone:1; gregtech:gt.metaitem.01-2502:1; 2500000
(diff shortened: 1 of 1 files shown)


Dream-Master:
We removed the single block

xSkewer:
The multiblock gets its recipes from this file.

Dream-Master:
Ah ok let me check that first

chochem:
so this is just reversing the nerf from a few weeks ago in a more complicated way? (https://github.com/GTNewHorizons/GTplusplus/pull/220)

Still dont understand why it got nerfed, so I am all for it. But you should at least talk to @miozune or this just goes back and forth.

xSkewer:
The MT wasn't meant to have a speed boost. The other multis with the "Speed: x% | EU: x% | Parallel: x" tooltips use the speed listed, not speed + 100% like the others. 

This buff makes it clearer in NEI too since before the recipes shown were effectively twice as long as they actually were.

Issue #10782, opened 2022-07-20
Compressed Meteoric Steel has wrong name


chochem:
### Your GTNH Discord Username

chochom#0271

### Your Pack Version

2.1.2.4-18-pre

### Your Server

SP

### Your Expectation

iron is iron and steel is steel.

### The Reality

its made from triple meteoric iron plates:
![image](https://user-images.githubusercontent.com/40274384/180090981-5a91947d-dc88-481c-aa9f-a116053857ef.png)

it makes dual compressed meteoric iron plates
![image](https://user-images.githubusercontent.com/40274384/180091080-846d5565-94d5-4f5a-9ffa-4cc18e58cdd8.png)

it pulverizes to meteoric iron dust, same with arc furnace, and fluid extractor. It also links to the meteoric iron gregtech NEI diagram pages.

The oredict is compressedmeteoriciron

![image](https://user-images.githubusercontent.com/40274384/180091739-c8cb7dde-4dc1-42c5-ab38-ab403ba3f7df.png)

And even betterquesting shows this very item as compressed meteoric iron
![image](https://user-images.githubusercontent.com/40274384/180091516-4b20edf9-8155-4a04-b312-7afc3954d7ad.png)

...but somehow its named compressed meteoric STEEL.

![image](https://user-images.githubusercontent.com/40274384/180091793-2a48a64d-d174-46e0-ad4f-74752b3bdc66.png)

### Your Proposal

Iron this out please. :)

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [X] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

PR #10797, opened 2022-07-22
GalactiCraft NEI override meteoric iron typo

vlaetansky:
GTNewHorizons/GT-New-Horizons-Modpack#10782


scripts/Minecraft.zs
```diff
@@ -3283,7 +3283,7 @@ furnace.setFuel(<minecraft:wooden_pressure_plate>, 75);
 
 
 // --- Compressed Meteoric Steel Plates
```
miozune, line 3285: change here as well to avoid future confusion

miozune (approved):

Issue #10741, opened 2022-07-17
Backport fix for PR: Transportation Crafting Extension Cards


OrderedSet86:
### Your GTNH Discord Username

Order#0001

### Your Pack Version

2.1.2.4-5

### Your Server

Private server

### Type of Server

Vanilla Forge

### Your Expectation

Crafting extension cards from Project Red: Transportation don't work at all. They are supposed to allow for output to be on a different block than input (for example, EBFs). Instead PRT eternally waits for the output to be on the input block:
![2022-07-17_13 10 58](https://user-images.githubusercontent.com/88752681/179423154-c17c3601-4d34-4802-a2cb-33d75bdc4a6d.png)
This is a simple setup - going from the left: request pipe, broadcaster on chest with 64 steel ingots, crafting card for steel ingot -> steel plate in bending machine, and crafting extension card (the bending machine outputs into the rightmost chest). But PRT waits forever for return on the bending machine instead of properly on the rightmost chest.

**Apparently this is a known issue and was fixed in PR 1.12: https://github.com/MrTJP/ProjectRed/issues/1416**

The fix appears to only be two lines long.

### The Reality

See above.

### Your Proposal

Backport the fix so crafting extension cards work properly.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [X] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

OrderedSet86:
Hi @Shiroi -

I looked into it, and it appears to work if the input item is in the top left of the crafting chip, like so:
![image](https://user-images.githubusercontent.com/88752681/207474727-a1a8038e-d6e4-437d-aa71-95599afde2ca.png)

But if you move the input to the bottom left, it breaks and sends the input to the extension chip instead:
![image](https://user-images.githubusercontent.com/88752681/207474776-681e0a7b-ab24-449e-99bf-f8a5cbd867e2.png)

No idea what the cause is...

Zrashin:
Hi, I cannot reproduce this issue and can use the extension perfectly fine in testing.
Testing with a Blast Furnace setup for Aluminium:
[...]

PR #10802, opened 2022-07-22
Project Red: Transportation Buffs

OrderedSet86:
(I wrote the https://gtnh.miraheze.org/wiki/Project_Red_Transportation_-_MV_Storage_and_Autocrafting article)

Project Red: Transportation is the earliest option for request crafting, but it's barely worth it.

The primary complaints I get from people after setting up PRT are:
1. This costs way too many circuits / is generally cost prohibitive for the value I get
2. Help, I used active chips (which you said not to use) and now I'm confused / it's buggy (this needs logic patches and questline adjustment, which will be done later)
3. Not having auto crafting table options is pain
4. Because of idiosyncracies with how item distribution works, you cannot reuse machines during a multi-step crafting procedure. This leads to additional logistics costs that may not be obvious if you haven't built a setup yet. (Later a more thorough fix for this would be good, but that requires logic instead of simple recipe changes.)

In b6e65e7, I modified the recipes to make them a bit easier to make:
1. Routed interface pipe: 4 gold plate, 576 molten redstone -> 1 gold plate, 144 molten redstone
2. Crafting chip, stock keeper chip, item broadcaster chip: 1 MV circuit -> 1 LV circuit
3. Stock keeper chip: 1 diamond plate -> 1 aluminium plate
4. Routed junction pipe: 1 diamond plate -> 4 aluminium bolt
5. Null-logic chip: 6 gold wire -> 4 gold wire
6. Disabled: Item crafting extension chip (broken https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10741), pressure/pressure resistance tube (no documentation on wiki), network latency pipe (slows items down - why?)
7. Item responder, dynamic item responder, item overflow, item extractor: 1 LV circuit -> 1 MV circuit (these are buggy and cause people to be confused - they should be discouraged)
8. Misc formatting / typo fixes

After b6e65e7, crafting chips cost (including 1/4th of routed interface pipe cost):
0.25 diamond plate -> 1 aluminium bolt (this keeps similar processing time - 384 / 9 / 4 = 10.6s LV vs 80 / 8 = 10s EBF + MV, but moves it to a more common material / workflow)
1 MV circuit + 0.25 LV circuit -> 1.25 LV circuit
1.75 -> 0.75 gold ingot
10 -> 7 redstone
2 stone plate
1.25 copper
1 glowstone dust
2 lime illumar
0.5 raw silicon
2 iron ingot
1 glass pane

In a102b30, I added GT assembler options for auto crafting table methods (worktable, auto workbench, enderio crafter, enderio machine chassis). These are needed to provide robust options for crafting table autocrafting for PRT.

I am added a NHCoreMod PR for faster / consistent voltage motor/pump/etc. assembler recipes. https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/347

(Sorry about the messy aluminium bolt commit - I reverted and cherry picked so it made three commits, but I interactive rebased and they are all combined into 1 commit now.)

### DIFF
--- a/scripts/Build-Craft.zs
+++ b/scripts/Build-Craft.zs
@@ -691,6 +691,8 @@ Assembler.addRecipe(<BuildCraft|Transport:item.buildcraftPipe.pipeitemsemerald> 
 // -
 Assembler.addRecipe(<BuildCraft|Transport:item.buildcraftPipe.pipeitemsemerald> * 8, <ProjRed|Transportation:projectred.transportation.pipe> * 8, <gregtech:gt.metaitem.03:32082>, 400, 480);
 
+// --- Auto Workbench
+Assembler.addRecipe(AutoWorkB, <Forestry:factory2:2>, <ore:gearGtSmallAnyIron>, <ore:itemCasingAnyIron> * 2, <gregtech:gt.metaitem.01:32600>, 200, 30);
 
 
 
--- a/scripts/Ender-IO.zs
+++ b/scripts/Ender-IO.zs
@@ -693,7 +693,7 @@ recipes.addShaped(<EnderIO:blockCrafter>, [
 [Circuit, MCasing, Circuit],
 [SteelCasing, <EnderIO:itemFrankenSkull:1>, SteelCasing]]);
 
-// --- Machine Casing
+// --- Machine Chassis
 recipes.addShaped(MCasing, [
 [SteelCasing, ESteelPlate, SteelCasing],
 [ESteelPlate, Capacitor, ESteelPlate],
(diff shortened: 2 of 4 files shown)


Magentanide:
Some of these changes make a lot of sense, like replacing done diamond with aluminium and lowering the cost of null chips to make more sense.

However, this doesn't really address the pull of PRT: storage, and some simple automated tasks before conduits become available - in the same tier. Anyone trying to use it outside of that shouldn't be encouraged by the low cost of fundamentals like routed junction pipes.

Instead, encourage migration away from them and towards conduits. After that, for more complex or specific tasks, you should look at Logistic Pipes.

While I do find LP far more prohibitively expensive, it sits on HV and that would give a nice progression from PRT --> EIO --> LP.

Nobody should be building large systems in PRT, which is what these changes encourage.

OrderedSet86:
@Magentanide I don't consider conduits to perform the same functionality as PRT. While you can approximate it using active cards like item responder, I think using PRT for this is laggy, buggy, and would never be cheap enough to make sense with the pack (hence why I actually nerfed these by increasing the cost).

Where PRT excels is request crafting automation, which EIO conduits, even with inventory panel, cannot do.

Dream-Master (approved):

Issue #10616, opened 2022-07-03
Give personal dimension at HV [RFC]


maniifestantes:
### Your GTNH Discord Username

maniifestantes

### Your Pack Version

2.1.2.4-14-pre

### Your Proposal

The personal dimension being tiered so high is weird, because by the time you get it, you already have your base planned, which makes moving it really annoying.

### Your Goal

Changing it to an earlier time would be better, i'd say to give it once you have done your t1 rocket, as that's when you get titanium, and when some people start their ae2 time, which makes it a perfect time to move.

### Your Vision

.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

Devtron3D:
Ill be honest this makes sense.
Many when they get it have AE and a LOT of stuff automated, we could go the hard-luck way and say that's life. 
But we also need to consider what issues come from this.
When people start packing up their base in IV, they have a LOT of stuff, barrels, super tanks, ME storage cells, drawers and so forth.
Most when moving at that point want to do it is as few trips as possible, so they trow a lot of stuff in ME cells, or compressed chests, like dollys with drawers and barrels. 
I know from personal experience how easy it is to cause errors to your player file when doing this, or corrupting a chunk.

Sovereignty89:
If we take the benefits of a personal dimension and place them in a list, we can better understand the possible balance implications.
Personally for me I see 3 benefits;
1. No monsters
2. Always day (is this still the case with Personal Space?)
3. FPS friendly and most headroom for expansion due to practically 'empty' world

1) Monsters
There comes a point in the pack, personally for me it's when I get a Nano Suit, that monsters aren't an issue anymore. Even with their crazy Infernal Mobs buffs, especially if you couple a Nano Suit with a good crossbow. The more you progress in this pack, the less you leave your base. You still need to go to specific dimensions to progress so you are still subjected to monsters. But long before HV mobs shouldn't be an issue in/around your base.

2) Always day
[...]

PR #10869, opened 2022-07-30
Moved the personal dimension quest to be available after going to the moon

chochem:
closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10616

Moves the personal dimension from the IV tab to the space race tab. More precisely, after the moon dungeon quest. This allows the player to build a new base with AE in the personal dimension in EV.

This has been discussed/proposed a few times. most recently in https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10616.

As far as I can tell this PR is in line with what dream wrote in that ticket. The dim should be available only after the player visited the moon and when he starts AE.


Connor-Colenso (approved):
Very much in favour of this. Will let Dream give his opinion before merging.

PR #10891, opened 2022-08-01
Removing MT Script

Runakai1:
removed MT script for removing fluid manipulator recipe in order for the new YOTTANK recipes to work

Yottank PR: https://github.com/GTNewHorizons/GoodGenerator/pull/61

New Recipe:
image


boubou19:
why not use another item instead?

Dream-Master:
did the fluid manipulator have no changed recipes? If no please add some

Dream-Master (approved):

PR #10916, opened 2022-08-03
Endgame goals rework

Connor-Colenso:
- Quick readme adjustment
- Endgame goals

image


chochem:
that looks a bit weird without edit mode or stuff unlocked. Shouldnt those probably all be secret except for infinity?
image

chochem:
actually 2 more things:

1. you can also change the name on the NaN quest back to something else. That just made sense as the gate to the stargate quests (its a reference to like the first episode/movie)
2. there are coild quests in the multiblock tab already. we should probably use the same an not make another one for each imo.

Dream-Master (approved):

PR #10923, opened 2022-08-04
Blacklist Gendustry Industrial Apiary from acceleration

kuba6000:
Blacklist Gendustry Industrial Apiary from acceleration to force people to use the new version
Also why there was **gregtech** apiary ? Removed it.

### DIFF
--- a/config/GTNewHorizons/dreamcraft.cfg
+++ b/config/GTNewHorizons/dreamcraft.cfg
@@ -278,7 +278,6 @@ modules {
             com.supsolpans.tiles.TilePhotonicSolarPanel
             gtPlusPlus.core.tileentities.general.TileEntityFishTrap
             gtPlusPlus.core.tileentities.general.TileEntityDecayablesChest
-            gregtech.common.tileentities.machines.basic.GT_MetaTileEntity_IndustrialApiary
          >
     }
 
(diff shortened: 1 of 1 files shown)


Connor-Colenso (changes requested):
I would say blacklist the regular apiary as well while you are there.

kuba6000:
> I would say blacklist the regular apiary as well while you are there.

Ok, done

Connor-Colenso (approved):

boubou19:
sad day for the WA enjoyers

Ethryan:
I thought the gregtech apiary was blacklisted since you could insert upgrades that does the same thing.

If you can use a normal WA on them they get insanely broken.

kuba6000:
> I thought the gregtech apiary was blacklisted since you could insert upgrades that does the same thing.
> 
> If you can use a normal WA on them they get insanely broken.

Not needed. All gregtech machines are blacklisted by default: https://github.com/GTNewHorizons/NewHorizonsCoreMod/blob/master/src/main/java/com/dreammaster/gthandler/accelerator/GT_MetaTileEntity_WorldAccelerator.java#L434

DianeXD:
MagicBees also has an apiary

PR #10934, opened 2022-08-05
Updated industrial apiary

Runakai1:
moved to HV

image
This but just imagine hv arms


Dream-Master (changes requested):
please move the recipe to code

PR #10953, opened 2022-08-06
Stop exempting Icebox

mitchej123:
It's never going to happen anyway... it's a graveyard polluting the active issues


YannickMG:
How about this instead: Anything currently closed in the icebox milestone gets removed from the milestone, then close all icebox issues. When new issues are added to the icebox, close them.

If someone wants, they can still tell appart "open" icebox issues from the rest and pick one to work on.

mitchej123:
I'm tired of 2 year old issues nobody looks at because of 47 pages.  I want it trimmed down to stuff people will actually work on.  I don't want to do more work around this

YannickMG:
Hmm well it'll still be tagged as 'Stale' right? Then I suppose there's no need to do anything else, that tells us they're actually (potentially) unsolved issues.

Dream-Master (approved):

Issue #10422, opened 2022-05-31
Add recipe to turn gt++ algae into biomes o'plenty algae


chochem:
### Your GTNH Discord Username

chochom#0271

### Your Pack Version

2.1.2.4 5-dev-pre-test

### Your Proposal

Add a simple recipe to turn the gt++ 'Algae Biomass' into the 'Algae' from biomes o'plenty. This could just be a basic 1:1 unshaped crafting grid recipe.

### Your Goal

The goal is to make the Algae farm actually produce algae while at the same time giving the player more options to produce sunnarium if they dont want to do ic2 crops or ic2 reactors.

At the same time, this change would not be too powerful either, as the algae farm is relatively slow and the autoclaving step for sunnarium would still be needed anyway.

### Your Vision

Make the algae farm usable for a quite useful type of algae. :)

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this feature please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I believe there is nothing similar in the pack already, or the existing solution isn't good enough.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members thinking it's inappropriate.
- [X] I believe this feature would make the pack better.

PR #11072, opened 2022-08-19
Add alternative algae recipe for glow coral

chochem:
Adds an additional recipe for glow coral that uses gt++ algae instead of bop.

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10422.


Dream-Master (approved):

Dream-Master:
lets add new recipes in the future in code

Issue #11106, opened 2022-08-22
Add AE2FC quests


AlbertoPirillo:
### Your GTNH Discord Username

AlbyNexus#4160

### Your Pack Version

2.1.2.3qf

### Your Proposal

~As we all know some EC2 components are deprecated and can be no longer crafted: thus their corresponding quests should be removed from the Applied Energistics tab.~
AE2FC has no quests to guide the player through this "new" mod. 

I suggest that we have at least a quest for every section of AE2FC's wiki: https://github.com/GlodBlock/AE2FluidCraft-Rework/wiki.
I haven't played with the mod yet so I cannot make any further recommendations. 

### Your Goal

~Replace old quests with new ones.~ New quests.

### Your Vision

This could provide a quickstart to AE2FC for every type of player, all inside of the game.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this feature please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I believe there is nothing similar in the pack already, or the existing solution isn't good enough.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members thinking it's inappropriate.
- [X] I believe this feature would make the pack better.

eigenraven:
EC2 has been un-deprecated and is supported now, the major issues have been fixed so the quests for those are still relevant.

chochem:
yea we are not removing them

AlbertoPirillo:
> EC2 has been un-deprecated and is supported now, the major issues have been fixed so the quests for those are still relevant.

Cool I'm editing out that part then

PR #11107, opened 2022-08-22
Added AE2FC quest

Runakai1:
Well read it duh
image
image

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11106


chochem (approved):
thank you!

PR #11125, opened 2022-08-24
Removed Apiary from World acceleration

Runakai1:
Normal Apiary isn't meant for production but rather as a pre alveary way of mutating bees. Adding it to the blacklist removes another useful feature from the apiary and making bee mutation way worse. 

Oblivion Frames can be found, which is annoying
Mutagenic frames need plutonium 241, which is annoying.
The only remaining usefull bee mutation device is the mutator and alveary

Since Industrial Apiary is HV, there is no need to blacklist the normal apiary

### DIFF
--- a/config/GTNewHorizons/dreamcraft.cfg
+++ b/config/GTNewHorizons/dreamcraft.cfg
@@ -279,7 +279,6 @@ modules {
             gtPlusPlus.core.tileentities.general.TileEntityFishTrap
             gtPlusPlus.core.tileentities.general.TileEntityDecayablesChest
             net.bdew.gendustry.machines.apiary.TileApiary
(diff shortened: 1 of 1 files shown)


chochem:
should probably be discussed with @Connor-Colenso @kuba6000 as they just blacklisted it.

Connor-Colenso (changes requested):
I would rather add lesser mutagenic frames and easier oblivion frames than this.

Runakai1:
> I would rather add lesser mutagenic frames and easier oblivion frames than this.

I could make mutagenic frames easier to obtain then if thats okay with yall.

mitchej123:
WA is a lag inducing workaround and should be avoided/minimized and removed as we provide alternatives

Issue #11226, opened 2022-09-06
[Iron Chests/Avaritiaddons] Add assembler recipes for easier Chest upgrading


BritishCynic:
### Your GTNH Discord Username

BritishCynic#9193

### Your Pack Version

2.1.2.4-dev29

### Your Proposal

To add the following Assembler recipes:

1 x Copper Chest + 1 x Double Steel Plate = 1 x Steel Chest
1 x Copper Chest + 1 x Double Iron Plate = 1 x Iron Chest
1 x Iron Chest + 1 x Double Gold Plate = 1 x Gold Chest
1 x Steel Chest + 1 x Double Gold Plate = 1 x Gold Chest
1 x Gold Chest + 2 x Diamond Plate = 1 x Diamond Chest
1 x Diamond Chest + 1 x Dense Obsidian Plate = 1 x Obsidian Chest
1 x Diamond Chest + 2 x Glass Pane = 1 x Crystal Chest
2 x Diamond Chest + 1 x HV Piston + 6 x Dense Obsidian Plate = 1 x Compressed Chest

### Your Goal

This makes upgrading chests via the Assembler possible while adding incentive for doing so by making the conversion slightly cheaper, and this has the added bonus of being automateable.

### Your Vision

To make upgrading chests less of a PITA.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

BritishCynic:
This was only partially implemented.

PR #11283, opened 2022-09-11
Add assembler recipes for lower tier metal chests upgrade

Quarri6343:
related to #11226 
I'll do Compressed chests on coremod side


Dream-Master:
@Quarri6343 please do ma a favor and add all recipes to coden. Even if you correct it. Thanks
I slowly do the same with recipes i fixed.

Issue #11290, opened 2022-09-12
Dark Deep Portal infusion recipe currently broken, likely from its recipe's instability value being over the limit. 


ChromicRedBrick:
### Your GTNH Discord Username

ChromicRedBrick#6260

### Your Pack Version

2.2.0.0-RC-2

### Your Server

SP

### Type of Server

Single Player

### Your Actions

I was testing the new infusion interceptor, everything went smoothly at first. But when I start the infusion of Portal to Deep Dark, the game crashes. I tried several times including performing a clean install of 2.2.0.0-RC-2, it still crashed when the infusion started, giving the same error. 

I also tried many other infusion recipes, but only DD made the game crash.

### Edit: Note from Alastor:
Not an issue with the intercepter, an issue with the portal's recipe, not related to the intercepter whatsoever, we gotta stop putting infusion instability values over 25 my guys.

### Crash Report

First crash report (RC1): https://pastebin.ubuntu.com/p/QjT2r5VKyx/
After updated to RC2: https://pastebin.ubuntu.com/p/nZ9cVxhTYX/
After a clean install of RC2:  https://pastebin.ubuntu.com/p/bvZnBz54Sz/

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the crash still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this crash consistently by following the exact steps I described above.
- [x] I have asked other people and they confirm they also crash by following the exact steps I described above.

Alastors:
@ChromicRedBrick 

A few questions:

1. Were you WAing the intercepter?
2. How were you supplying essentia?
3. Could you provide a recipe for the DD portal? (I could get it myself in game, but in general it would help to have it here to know if it's anything specific with the recipe itself?

Alastors:
Oh one other thing, were you WAing the infusion matrix?

Alastors:
I'm wondering if it's a bug with the matrix itself, the intercepter, the WA, or if it's a bug with recipe tweaker as a whole (I doubt it's this one, but one never knows)

ChromicRedBrick:
I reproduced this crash in the video: https://youtu.be/9u3fcMvMm8I

Crash report is almost the same: https://pastebin.ubuntu.com/p/Zd4fTTJmWh/

1. I'm using the infusion interceptor only. 
2. I supply essentia with thaumic energistics.
[...]

PR #11305, opened 2022-09-13
Reduce instability for deep dark

Ethryan:
Since it seems to crash with values above 32
#11290


Alastors (approved):

Alastors:
I'm going to merge this, since it's a small enough change, and should fix the issue itself but if you wouldn't mind, please test the new recipe and see if it does actually fix it.

PR #11597, opened 2022-10-09
decrease LP item price by 33% to 50%

Glease:
because apparently people still complain about this being too expensive

also added another tier for the Unrouted Logistics Pipe. Mass boro glass usage just doesn't belong to a tier before LuV. unrouted pipe is even worse than tin item pipe so this seems pretty justified.


Dream-Master:
1 min to later for be in 2.2.4 :P

Dream-Master:
but yea will look into it

Dream-Master (approved):

PR #11623, opened 2022-10-12
Edited olive saplings quest to allow all Harvestcraft saplings as choice rewards.

hallucinogender:
Until the market is fixed, this change allows players to access whatever Harvestcraft saplings they want without having to spend hours searching with little chance of success at finding any particular one.


chochem:
This might have been better as a ticket for discussion. not just a PR. See discusssion in #quest-dev.

In short, it removes some exploration for the purpose of (even) faster food progression (and thus hearts). Which might be a negative or positive depending where you stand on this.

PR #11643, opened 2022-10-14
more lp changes

Glease:
mostly bugfixes. will push more changes as I progress


Steelux8:
The commit I made is to set the Request Logistics Pipe at MV, where it's supposed to be. The Item Circuit Tier I input is an HV Forming Press recipe, whereas the cheaper Iron Chipset is an MV Forming Press recipe. Suggested and tested by @TerraMatter-dev

Dream-Master (approved):

PR #11683, opened 2022-10-19
Change fluid manipulator to assembler recipe

greesyB:
AE is very slow at simulating shaped crafting recipes for anything that has an input with a containerItem (like canisters), making it impractical to craft even a couple dozen or so yottank fluid cells that require many fluid manipulators due to being unable to simulate them in reasonable time.

image


greesyB:
Most items that are affected by this slow simulation that I can think of have alternative recipes in assemblers. I guess I can move most or all of the GC minetweaker recipes to code, will look into it later.

PR #11732, opened 2022-10-25
Mainframes, Your First IV Circuit

Eldrinn-Elantey:


Dream-Master:
@Eldrinn-Elantey what you add to the quest? a bit more info would be nice

chochem:
apparantly it adds 24 smd capacitors. 

The why is not clear though. This is not the fix this quest needed. There actually is outdated info in the description but this does not fix that, actually it makes it worse.

chochem:
I am not just closing though, I will make a better version. + maybe some other fixes

Eldrinn-Elantey:
> I am not just closing though, I will make a better version. + maybe some other fixes

Thanks

PR #11777, opened 2022-11-01
Cheaper tiered solars

Tyrantgtnh:
Tiered solars are more reasonable to craft, for example for uv solars requiring 6 uiv circuits as well as 2 of the previous tier is quite excessive so it was changed to be craftable in tier+2. Also the change from iv->luv solars becoming vastly more expensive with them requiring more circuits, the circuits being tier+3 rather than tier+2, and requiring 2 of the previous tier over 1. While these solars still won't compete with fusion spam (as it should be) they aren't completely irrelevant either and actually have a niche. 
image
image
image


Connor-Colenso (approved):
Discussed in discord vc.

Dream-Master (approved):

Dream-Master:
The solar need a rework some day to get rid of the recursive crafting. I want it more put in machines make high tier foilds and stuff to get petter solar modules but for now we can merge this.

Issue #10147, opened 2022-04-13
Remove Bedrockium TC nugget dupe


LewisSaber:
### Your GTNH Discord Username

Lewissaber#7231

### Your Pack Version

 2.1.2.3QF

### Your Proposal

Remove thaumcraft way to dupe bedrockium nuggets

### Your Goal

Make it less op

### Your Vision

It makes bedrockium too easy for how good it is

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

Steelux8:
Bedrockium is very easy to make in the first place, for the high tiers that it (for some reason) is used in. Even more, the setups used for the cobble are often laggy, so it's a good thing to avoid the lag on cheap recipes by having options like this.

Also, if it was overpowered, everyone would be doing it, which I'm sure is not the case given how many people simply don't use TC. I don't know how any of the duping recipes can be considered OP by that standard.

boubou19:
this is the recipe in question:
![image](https://user-images.githubusercontent.com/12850933/163279972-1b7ae8f7-0f74-4e77-bbfa-10ab9f96852c.png)

The bedrockium creation in it's normal way is totally automatable passively, and is boring. Also, players wanting to abuse bedrockium with cobble gen will most of the time destroy the TPS. Also, post 2.1.2.0, GT++ adds a multi to make cobble, making this even easier.

I'll add that as @Steelux8 said, almost no one will use thaumcraft, because it's thaumcraft.

Now, if we look at what is actually useful the bedrockium for, well you'll see that it's used in UHV component, in bedrockium drums (which are gated by moon anyways due to the normal drum in the recipe, so EV), some tools, a wire(which is useless as player can use better cables because they aren't lossless) and pipes (there too players can make pipes in better material to be able to make them in bulk). In any case, except UHV components, you don't need much bedrockium to achieve that. So even if the recipe gets removed, all the """overpoweredness""" of the bedrockium is still availiable. So i don't see the point except nerfing thaumcraft for the sake of it.

Leaving some advantages to use thaumcraft is a great way to get people's interest in it.

PR #11802, opened 2022-11-08
blacklist some space ore and lanth nuggets dupe recipe

GlodBlock:


boubou19 (changes requested):
some of them are unoffensive and bedrockium dupe is harmless too

BlueWeabo:
Bedrockium isn't too op. You need to use tons of essentia to get the bedrockium needed. And only ways to get this kind of essentia is through platinum or wisps.

boubou19:
for bedrockium, see my comment here: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/10147#issuecomment-1098544639

after tantalum, except bedrockium, they are all kinda useless (except duralumin that is used a bit maybe)

Also Mythril is kinda useless too

GlodBlock:
some are lanth materials that may going to be used in future.
the space ore part can be reverted if you really think they are useless, tho they have very good turbine values.

boubou19 (approved):

BlueWeabo (changes requested):
I don't see any point in blacklisting stuff like aluminium, chrome, ect. Which are materials easily accessible and allows people, which go into TC early to get some benefits for a higher chance to destroy their base with the current warp effects.

Not to mention that to first dupe a material you need to have gotten at least one of the materials as an ingot.

I can understand about lanth lines, but those materials don't have a meaningful production until DD portal. So the dupling is an alternative after you have gotten some of the material. So you will not be skipping the lanth line.

The essentia costs to upkeep these resource productions are very hard to meet. One cannot easily make the needed resources without giving up some other material and a lot of TPS.

GlodBlock:
the lanth line material still should be blacklisted imo. currently some lanth material are only used in very small amounts, so if it is dupable, people will only manually make some and turn to dupe recipe

BlueWeabo:
People will manually make some if they know about the duping recipe. I am not sure if the recipe is shown after the nei update, which allowed us to see non-unlocked recipes.

Many people would still prefer to do the lanth line than set up duping as getting the essentia can seem annoying.

I can let lanth line materials to stay blacklisted, but some other not so much.

Why are Gallium, Tantalum and Neodymium still black listed? Sure they aren't used a lot, but they are much easier to get later in game so it's a good way for those, which can go through TC get some benefit from it. I must remind that TC starts at MV where one can start these dupes, which won't benefit those at LV where Galium is used first.

Neodymium you get tons when you enter EV cus you have been to the moon, so it doesn't really matter if there is a dupe.

Tantalum is mainly used in EnderIO, which is already a bit of a pain to get. It does give an alternate smd capacitor recipes which isnt even worth using.

boubou19:
i'll add that xp buckets is also a thing if really you want to get stuff

BlueWeabo (approved):

PR #11821, opened 2022-11-12
Thorium as Early Power

Eldrinn-Elantey:
I think it is wrong to indicate the creation of a nuclear reactor and its chambers as an optional part of the quest


Steelux8:
These quests have minimal, if any rewards. There is basically no benefit to completing them, and locking their completion only serves to block progression to the following quests.

I made the quests this way because the nuclear setups are grouped in that page, but most people will not build them in order: they'll simply decide to build a setup at some point, and it might very well be one that's not at the beginning of the questline.

PR #11943, opened 2022-12-10
Add assembler recipes for Apiary, Scented Paneling, all Alveary upgrades

Alrightsc:
This is my attempt to add 15 new recipes to GT assembling machine, which all are tiered at MV. All the recipes here take 60 seconds at MV tier, to avoid making Carpenter immediately obsolete as it's still cheaper and faster in terms of time/EU. All alveary upgrades' honey cost also are halved (open to feedback). At higher tiers becomes painful to automate carpenters as you need to spam them and automate each, and so on which I think takes a lot more effort, so those recipes are more or less useful for later in the game.

image
image
image

### DIFF
--- a/scripts/Extra-Bees.zs
+++ b/scripts/Extra-Bees.zs
@@ -4,6 +4,7 @@
 
 // --- Imports ---
 
+import mods.gregtech.Assembler;
 import mods.forestry.Squeezer;
 import mods.gregtech.Pulverizer;
 import mods.gregtech.Extractor;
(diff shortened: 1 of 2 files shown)


Connor-Colenso (changes requested):
I like this but this needs to be added in code, we are not permitting new scripts unless under exceptional circumstances. I would recommend putting this in core mod.

Alrightsc:
See https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/452

PR #11944, opened 2022-12-10
Update bartworks.cfg

Impos913:
Add Turbo Supercharging Combustion Generator to EV generators list for ross128b ruins.


Impos913:
Added EV gas generator too and removed 16A EV cable diode from generators list.

Dream-Master (approved):

PR #11948, opened 2022-12-10
Fixed the "Mundane Larch" quest

Eldrinn-Elantey:
The mandatory item and the optional item were mixed up in places


Dream-Master:
@Eldrinn-Elantey why you need 100 of commits to just change two lines of questbook ?

Eldrinn-Elantey:
> @Eldrinn-Elantey why you need 100 of commits to just change two lines of questbook ?

I started my acquaintance with git when I started making changes to your modpack. I honestly don't know why it happened. I tried to figure it out, but the only thing that happened was to make sure that only these two lines were sent. The remaining commits are either leftovers from other edits or their cancellation. I still don't understand why they are listed here even though the repositories are synchronized

chochem:
the reason is that you are doing a PR from your master which has a different history even if the files are identical. To avoid that make a branch based on upstream/master.

Eldrinn-Elantey:
> the reason is that you are doing a PR from your master which has a different history even if the files are identical. To avoid that make a fresh branch based on upstream/master.

@chochem Is it done here?
image

Issue #12080, opened 2022-12-24
Impure tear recipe is bugged


IceyPhantom:
### Your GTNH Discord Username

Icephantom#0300

### Your Pack Version

2.2.3

### Your Server

SP

### Type of Server

None

### Your Expectation

Providing all recipe ingredients for infusion crafting of impure tear gives pure tear

### The Reality

recipe does not start with all correct ingredients provided as per NEI

### Your Proposal

Shibva came up with a script workaround by adding the gt fluid cell to the recipe, which let the infusion start.  Something about the universal fluid cells are not being recognised properly in the recipe

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [ ] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

PR #12081, opened 2022-12-24
Update Warp-Theory.zs impure tear infusion not starting

Shibva:
Changed **<IC2:itemFluidCell>.withTag({Fluid: {FluidName: "lifeessence", Amount: 1000}})** to **<gregtech:gt.metaitem.01:30694>**; using the GT cell version of life essence instead of IC2. fixes the issue some people were having with the infusion not starting. Because you can only fill GT cells up to 1000mB this should be the same portions in the recipe as before

-unsure problems: unsure if initial method returned empty IC2 cell or not; including this here to make it known and to point it out if it did or not


Shibva:
should fix #12080

Dream-Master (approved):

Issue #9331, opened 2022-01-02
Change Plutonium 238 vein height so it can spawn properly


fourk:
### Your GTNH Discord Username

fourk#5436

### Your Pack Version

2.1.2.1

### Your Server

sp

### Type of Server

Single Player

### Your Expectation

![image](https://user-images.githubusercontent.com/101964/147866743-9575bcd9-62e2-4227-a0aa-c396f3ea0033.png)
![image](https://user-images.githubusercontent.com/101964/147866775-e842ccf8-1f29-4622-97c0-8b5fa6a17d04.png)

When visual prospector shows a vein as 'Plutonium' it should contain ores. 

### The Reality

I have found 4 of these so far on Callisto, all of them have no ores, not even the small ores that should be associated with the vein type.

### Your Proposal

Either don't show plutonium as the ore type in visual prospector when those ores don't exist in the marked ore chunk, or spawn the appropriate ores for that vein type.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [X] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

GTNH-Colen:
Annoyingly I don't think this is actually a bug, the vein would've spawned there but the map wasn't high enough so it couldn't do so. Plutonium spawns at 55-120, so you can see how this easily occurs.

The vein should be moved down but I'm not sure how to do that.

Dream-Master:
sure i can do that. Which high you suggest? It can interfear with other mixes.

bombcar:
@SinTh0r4s could visual prospector “realize” a vein failed and not show it?

Dream-Master:
yes if the vein is to small it not shown or if you digg out most if it it vanished /not shown.

fourk:
> but the map wasn't high enough so it couldn't do so.

[...]

Issue #7252, opened 2021-01-05
Uranium 238 Vein/Plutonium 341


braeven:
#### Which modpack version are you using?
2.1.0.0
#

#### What did you try to do, and what did you expect to happen?
I would like to propose to change something on the Ore Generation:
Currently the Uranium 238 Vein is the only source for people to get Plutonium 241(out of the Plutonium 239) if they are not using Reactors, but it's really rare to find on T3 Planets:
it only has a weight of 5 and generates above 65(height 65-120)
it currently spawns on Callisto, Ceres, Ganymede and Io(T4).

I just searched on Epsilon for 2 veins on Callisto, I prospected around 1100 Veins to find 2 partly not generated veins.
it should be around a 1% chance for a Vein to be generated on Callisto(80 Chalco + 120 Limo + 160 Magni + 10 Scheelite + 16 Ilmenite + 30 Bastnasite + 10 Platinum + 40 Callisto Ice + 5 Uranium -> 471, Uranium itself has a weight 5 -> (5/471)*100 ->  1.06%), so mathematically I should have found  around10 Veins(I know GT-Generation works a little different).

The T3 planets tend to have a rough Terain, with some high Mountains, deep Holes and a "sealevel" around 60 and not much area for this vein to be generated.

#
#### What do you suggest instead/what changes do you propose?
I would like to propose to lower the minimum height from 65 to 55. This might not be much, but I think it will be enough to let the vein generate a little more often and give the players more access to plutonium related stuff, that are not using Reactors

johnch18:
In my playthrough I very quickly had more 241 than 239, so I don't think this is the biggest problem

Glease:
Voiding half the vein is quite frustrating though.

repo-alt:
I think it is just an oversight in the first place.

repo-alt:
Where are those veins defined?

Glease:
Both in [source code](https://github.com/GTNewHorizons/GT5-Unofficial/blob/c99a6e14495c82f6022e8676b412a37775af85f6/src/main/java/gregtech/loaders/postload/GT_Worldgenloader.java#L202) and [config](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/blob/422b5b7762d68c20e3a9afacb8c3143c946786e7/config/GregTech/WorldGeneration.cfg#L773-L778). Probably we could just add a line in the config to override the height.

johnch18:
[...]

PR #12287, opened 2023-01-15
Update WorldGeneration.cfg

thecodingchicken:
Increase weight of richnuclear vein, as well as change height to make it actually spawn
Noticed this today as all of the plutonium veins that I found didn't exist, given that ground level on those planets is around 70 blocks.  
Weight was increased by 5 to make it slightly easier to find, however it is still very rare.  
Closes issues #9331 and #7252

### DIFF
--- a/config/GregTech/WorldGeneration.cfg
+++ b/config/GregTech/WorldGeneration.cfg
@@ -781,7 +781,7 @@ worldgen {
                 B:GalaxySpace_Io_false=true
                 I:MaxHeight_120=40
                 I:MinHeight_55=5
-                I:RandomWeight_5=10
+                I:RandomWeight_5=5
             }
 			rutile {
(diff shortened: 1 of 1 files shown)


chochem:
they do absolutely spawn, just maybe like 1 in 3 or so at the moment. That would make this a 6x buff I guess. sounds fine.

thecodingchicken:
I think it would be fine to change the weight back to 5 if that would be better, but I also feel like while they are super rare(and should be given the rewards from them), they are a bit _too_ rare.
My thoughts behind it are that if journeymap says that if an ore exists in a given chunk, I should be able to find it.

Dream-Master:
@thecodingchicken change back to 5 .

thecodingchicken:
Alright

Dream-Master (approved):

PR #12293, opened 2023-01-16
Food Balance: HungerOverhaul and SpiceOfLife

Minepolz320:
This food-formula allows you to eat food even if its nutritional value has greatly decreased, but only if you are very hungry and maximum restoring hunger capped by 3 bars 

I think this solution is more logical and not so annoying 
and also food can be eaten more times

### DIFF
--- a/config/HungerOverhaul/HungerOverhaul.cfg
+++ b/config/HungerOverhaul/HungerOverhaul.cfg
@@ -299,7 +299,7 @@ hunger {
     B:disableHealingHungerDrain=true
 
     # Speed up or slow down the rate that hunger drops (set to 0 to disable hunger loss) [vanilla: 100.0] [range: 0.0 ~ 3.4028235E38, default: 133.33334]
-    S:hungerLossRatePercentage=100
+    S:hungerLossRatePercentage=50
 
     # The amount difficulty modifies the hunger value set after respawning  ('difficultyScaling' and 'difficultyScalingRespawnHunger' must be true) [vanilla: 4] [range: 0 ~ 2147483647, default: 4]
     I:respawnHungerDifficultyModifier=4
--- a/config/SpiceOfLife.cfg
+++ b/config/SpiceOfLife.cfg
@@ -86,13 +86,13 @@ server {
     # If set to greater than zero, food eating speed will be affected by nutritional value
     # (meaning the lower the nutrtional value, the longer it will take to eat it)
     # Eating duration is calcualted using the formula (eating_duration / (nutritional_value^eating_speed_modifier))
-    D:food.eating.speed.modifier=0.5
+    D:food.eating.speed.modifier=1.0
 
     # The maximum amount of eaten foods stored in the history at a time
-    I:food.history.length=64
+    I:food.history.length=20
 
     # If true, food history will not get reset after every death
-    B:food.history.persists.through.death=true
+    B:food.history.persists.through.death=false
 
     # Rounding mode used on the hunger value of foods
     # Valid options: 'round', 'floor', 'ceiling'
@@ -127,7 +127,7 @@ server {
     B:give.food.journal.on.dimishing.returns.start=false
 
     # The number of times a new player (by World) needs to eat before this mod has any effect
-    I:new.player.food.eaten.threshold=1
+    I:new.player.food.eaten.threshold=10
 
     # If true, any foods not in a food group will be excluded from diminishing returns
     B:use.food.groups.as.whitelists=false
--- a/config/SpiceOfLife.cfg
+++ b/config/SpiceOfLife.cfg
@@ -117,7 +117,7 @@ server {
     # 	total_food_groups : The total number of enabled food groups
     # 	exact_count : The number of times the food (ignoring food groups) has been eaten within the food history
     # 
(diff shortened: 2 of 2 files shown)


Elisis:
You've changed a lot more than what's stated in the description. Did you copy this config from somewhere?

Steelux8 (changes requested):
You have to clarify all the changes made here, there are nerfs to the beginning of the pack due to more foods being stored in memory, the penalties kicking in after only 1 food instead of 10, etc.

iouter:
then it's not just changing the formula

KaneHart:
100% denied. What you suggested was not bad but you butchered the config with so many changes like food.history.length=64, persistent through death, etc.

Minepolz320:
I:food.history.length=64 
So now food lose nutrition more slowly this means we need a bit longer food history 
32 also just fine 

Here i edited round mode for saturation modifier if needed i can include some ROUND(x) to formula insted of using this 
S:food.hunger.rounding.mode=round
 
 
Why i set this to 1 i feel this thing is useless its resets only if you create new player profile and you can eat any food if you have 3 or less hunger bars
I:new.player.food.eaten.threshold=1


Some ppl just abuse this thing by dying 
B:food.history.persists.through.death=true

Some food in this pack restore ALOT of hunger
and eating this food feels like eternity
D:food.eating.speed.modifier=0.5

Minepolz320:
@KaneHart Why is that so scary?

BlueWeabo:
> Some ppl just abuse this thing by dying B:food.history.persists.through.death=true
And that is a problem? Let people play how they like.

Minepolz320:
@BlueWeabo maybe... but then what's the point of having this mod at all

KaneHart:
> 

I think the biggest shocker is all the hidden stuff you did not mention then trying to make food harder in some ways. People are lazy their rushing for healing axes, and such... I like to see more life given to food and your idea originally was not bad, but then the other changes are what made people turn on you. One they were not spoken about and 2 they just don't agree with. If you like to see the conversation about it you can go into the #mod-dev discord channel.

boubou19:
@Minepolz320 Revert any hidden change and we will accept the PR. By hidden change i mean changes not on the description of the PR.

Minepolz320:
@boubou19 ok

Steelux8:
Please explain the changes you made in the formula.

Minepolz320:
I need more time for testing
given the changes you suggested
I'm sorry for being kind of sus 😅

joshaup1999:
Thats just gonna make beef welingtons at the beginning of the game even more annoying... Unnecessary nerf for early game imo

Minepolz320:
@joshaup1999 Quite the opposite

Minepolz320:
Initial idea for implementation
image

Dream-Master:
@Minepolz320 i looked the video but i cant get what is changed? MAybe make 2 vides so we can compare old settings and your new settings.

Minepolz320:
https://drive.google.com/file/d/1rmWBdR9MbF-Wqn6-GWuonlVt95Fk3mz2/view?usp=sharing

Minepolz320:
Even if it makes the "food-gameplay" worse, which I doubt.... it's just a config, just revert it back

capSAR273:
> Even if it makes the "food-gameplay" worse, which I doubt.... it's just a config, just revert it back
> 
> This is why I'm so persistent 😅
> 
> _Food is a critical thing early game and you can't keep eating the same things because of Spice of Life. Better watch out for some gardens from Pam's Harvestcraft to get some variety into your nutrition!_

Telling people to "just revert it back", means the change should not be made in the first place. The point of a PR is to introduce changes that the majority of players would agree with, and I don't think that would be the case with what you want to change about the food system.

Minepolz320:
@capSAR273 Why do you think these changes are worse than the current state of the food mechanics?

AlexisGoblot:
Yeah let's just spam medium food more instead of OP food. Seriously, OP food should stay OP because it gives an incentive to make them. Given it's already bad as it is to make food, we should not push away the players.

> it's just a config, just revert it back
>
> This is why I'm so persistent 😅

Nice so following this logic we can just not accept this PR 🤦. Seriously the goal of making a PR is to establish default configs for the pack, which mean default for everyone. Being persistent to make such change to force _*your*_ default, despite being very impopular, nerfing the early game for the sake of it, and not following our vision of the pack is a good indicator that we should not accept this PR.

Minepolz320:
@AlexisGoblot  it seems ppl just immediately think that this is just senseless food nerf and after that very concerned about it

boubou19:
> @AlexisGoblot it seems ppl just immediately think that this is just senseless food nerf and after that very concerned about it

from your own message:

> TLDR:
>
> 1. OP food got nerfed
> 2. Medium food buffed
> 3. If you have any food, you will not die starving

point 1 and 2 are worsening the food balance, making useless the top food of the game, because why would you make complex and time consuming food when you can just make basic food?

@Dream-Master the 3. is up to you tho: do you agree to allow player eating every food to avoid death by starving? right now if you are starving, you can avoid death if and only if you have food you can eat, ie, food that have been eaten less than 7 times over the 28 last food.

boubou19:
After dicussing the PR with Dream, it is clear that it's a change we don't want in the pack.

Minepolz320:
@boubou19 Ok thanks for discussion, I hope there will be another solution in the future :)

Dream-Master:
Sure if you @Minepolz320 have a better idea maybe open up a ticket first and add your solution for discussion there or just start it in discord before the pr.

PR #12294, opened 2023-01-16
change item IDs from shown to hidden

Alexdoru:
the players don't need to see non-sense numbers appened at the end of item tooltips

### DIFF
--- a/config/NEI/client.cfg
+++ b/config/NEI/client.cfg
@@ -12,7 +12,7 @@ inventory.disableMouseScrollTransfer=false
 inventory.gamemodes=creative, creative+, adventure
 inventory.hidden=false
 inventory.invertMouseScrollTransfer=false
-inventory.itemIDs=1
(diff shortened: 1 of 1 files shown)


Dream-Master:
if you hit H+F3 this will be enabled again ?

Alexdoru:
> if you hit H+F3 this will be enabled again ?

my PR just changes the default setting to not show item ids added by NEI : 
Currently it's like this : 

image

My PR changes the default to this : 
image

image

You can always turn it back on in the NEI settings in the inventory and f3+H still works no problem : 
image

Alexdoru:
> mhm not a fan of this, because with the ids you can sometimes spot why your recipe doesn't start by looking at the meta ids

you can always re-enable it, my goal is to make the game look cleaner and more appealing

Zereff06:
it would be nice if all unnecessary information was shown only in debug mode (f3 + h)

PR #12295, opened 2023-01-16
change default position of waila hud and remove background

Alexdoru:
The goal is to make the game look nicer and more appealing, I changed the default position to be centered at the top but below the boss bar and removed the background because it takes too much space on the screen

### DIFF
--- a/config/Waila.cfg
+++ b/config/Waila.cfg
@@ -1,7 +1,7 @@
 # Configuration file
 
 general {
-    I:waila.cfg.alpha=62
+    I:waila.cfg.alpha=0
     I:waila.cfg.bgcolor=1048592
     I:waila.cfg.fontcolor=10526880
     I:waila.cfg.gradient1=5243135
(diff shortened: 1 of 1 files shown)


glowredman (commented):
If it uses "too much space" why did increase the scale? I personally prefer bottom-right so all in all I don't support this change.

glowredman:
That doesn't answer my question in any way.

Alexdoru:
> That doesn't answer my question in any way.

using values different than 1 makes the text "compressed" it looks very bad 
image

here in scale 1 : 
image

Zereff06:
this is a bad idea by default, but you can add this to the config and anyone can use it at will

glowredman:
With this change, something like this will block the main part of the screen (where the crosshair is). I find this problematic.

Alexdoru:
> With this change, something like this will block the main part of the screen (where the crosshair is). I find this problematic.

do you spend your whole gameplay time looking at ender io conduits filled with 40 different fluids...

Alexdoru:
> uses "too much space" why did increase the scale? I personally prefer bottom-right so all in all I don't support this change.

I increased the scale so we can read better but by removing the background we can now see through and it frees lots of space on the screen to see the actual gameplay

POPlol333:
It's my personal opinion, and it only involve me but I feel like even if it is transparent it's too big, kinda like a visual magnet

embeddedt:
IMHO the background should not be removed. It significantly reduces the contrast of the WAILA information which is unfriendly to new players. I'd say it's better to leave the default and players can customize it to their own personal preference.

leagris (changes requested):
Change rejected

Everyone taste is different. No reason to change the default. Anyone can change these settings to fit needs if it differs from default.
Users are satisfied with current default by a good margin.

Alexdoru:
good

PR #12636, opened 2023-02-18
fix readme

chochem:
Dream reverted stuff by accident. this is just putting things back in.


boubou19:
template needs to be updated in DAXXL or it will be erased again

Dream-Master (approved):

PR #12852, opened 2023-03-10
EV Quest Changes to Reflect Tier Rebalance

Steelux8:
- Added and tweaked quest descriptions and prerequisites to better reflect the EV changes that were made since the last stable;
- Other minor description tweaks.


chochem (approved):
thanks!

PR #12901, opened 2023-03-14
Make ritual stones slightly less CBT to make en masse

BestMod:
Title. This change wasn't very well-thought-out, but if the general concept of making these cheaper is fine and the change itself is not, I can put some more thought into it.


Dream-Master (changes requested):
7.5 is no value that work make it 8 or 10

BestMod:
ok give me a minute

Dream-Master (approved):

PR #12955, opened 2023-03-19
Small LV solar recipe adjustment

chochem:
Change:
LV Solars no longer need Sunnarium. 

MV+ Solars still do need Sunnarium as before. And LV solars still need a Direcrafting table, MV SCs, Phosphor doped wafers, PTFE, etc.

Impact:
LV Solars will be available in late HV instead of mid EV, but will still be very expensive.

Reasoning:
As also explained in meta-dev, solars should be very expensive and hard to get as when you actually have them, they are easy free power. The interesting infrastructure and expenses are solely up front. However, they should still be in the realm of doable and at most delayed to tier+2 (tier+1 later), to not be completely irrelevant. Indeed the LV solar is currently only available in tier+3=EV and indeed completely irrelevant.
Indeed for a long time sunnarium was HV, but it was moved to EV.

### DIFF
--- a/scripts/CoreMod.zs
+++ b/scripts/CoreMod.zs
@@ -676,7 +676,7 @@ Assembler.addRecipe(<dreamcraft:item.EngineeringProcessorItemAdvEmeraldCore>, <a
 Assembler.addRecipe(<dreamcraft:item.LeadNickelPlate> * 2, <GalaxySpace:item.CompressedPlates:3>, <GalaxySpace:item.CompressedPlates:6>, <liquid:ic2coolant> * 2000, 600, 1024);
 
 // --- Irradiant Reinforced Aluminium Plate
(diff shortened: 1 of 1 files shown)


Dream-Master:
ok but glowstone is not adequate exchangeable item

chochem:
ah we still have the compressed ones from gt++ right? double compressed good?

Dream-Master:
no the gt one i removed it look snot realy good but can we do a complet overall of solar not change single recipe. I guess they need to redone and the late tier solar can be in pcb or so. I want to remove avaritia recipes for solar as well and the low tech ones have to be in machines no crafting recipes if you ask me.

chochem:
There is no double compressed glowstone in gt. only gt++. I don't know what you are thinking of then as a replacement here.

This particular recipe is in a gt machine so should work well with any future larger overhaul. I can start thinking about that a bit.

chochem:
Its tricky to find a HV replacement for a small piece of sunnarium if nothing glowstone related is ok.
How about a vibrant crystal? certainly adds some extra complexity.

Dream-Master:
everything than glowstone is more complex and better

chochem:
alright. changed to vibrant crystal

Dream-Master:
but the overhaul need to be not forgotten :)

PR #13001, opened 2023-03-22
from tiny to dust

Quetz4l:
All the tiny piles and nuggets have changed into dust and ingots. The odds of getting a single dust or ingot are almost the same.  It will get rid of unnecessary tiny piles of dust


Runakai1 (changes requested):
Can you change the chance for the full naquadah dust to 0.4%

Dream-Master:
other values are ok @Runakai1

Runakai1 (approved):

Quetz4l:
> Can you change the chance for the full naquadah dust to 0.4%

in 2.3.1 these are the odds, and platinum and naquadah are 2 tiny piles of dust (don's ask me why)
image

after changing the odds, we get 2 platinum dust and 1 naquadah dust
image

So I increased the odds for naqahdah

PR #13038, opened 2023-03-25
Unlock world generator settings

eigenraven:
They all seem to work fine now, tested in 2.3.0


Dream-Master:
Not sure if this is a good idea? Maybe people complaining not finding this or those. Sure we locked it because vanilla biomes have big mushroom biomes. That might be fixed but not sure in which issues we run when we open it.

eigenraven:
Seems the mushroom biome issues is still there, nevermind. I thought it was only generating them, not just bigger ones

PR #13080, opened 2023-03-29
Convert quests to latest format

chochem:
specifically BQ [3.3.11-GTNH](https://github.com/GTNewHorizons/BetterQuesting/releases/tag/3.3.11-GTNH)


D-Cysteine:
Thanks for doing this! There's actually one more format change that I wanted to do, but wasn't sure of a good time to do it. I'll see if I can sneak it in after this one =P

I wanted to move the quest lines ordering to a separate file, so that we can have multiple people adding / removing / re-ordering quest lines without getting nasty merge conflicts. Right now, we'd need to manually fix the order number in each affected quest line file.

D-Cysteine (approved):

PR #13114, opened 2023-04-02
Update questbook i18n workflow for new format

MuXiu1997:
The `en_US.lang` generated through this script is slightly different from the one generated using commands in the game (in the order of quests). Please refer to the actual generated file 
- generated file: [en_US.txt](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/files/11132492/en_US.txt)

- ref: GTNewHorizons/BetterQuesting/pull/89

### DIFF
--- a/.github/scripts/questbook-i18n.py
+++ b/.github/scripts/questbook-i18n.py
@@ -1,60 +1,119 @@
 import json
+import logging
+from abc import ABCMeta, abstractmethod
+from pathlib import Path
 from typing import TextIO
-from abc import ABCMeta
 
-DEFAULT_QUESTS_PATH = 'config/betterquesting/DefaultQuests.json'
-EN_US_LANG_PATH = 'config/txloader/load/minecraft/lang/en_US.lang'
-DEFAULT_QUESTS_US_PATH = 'config/betterquesting/DefaultQuests-us.json'
+DEFAULT_QUESTS_BASE_DIR_PATH = Path('config/betterquesting/DefaultQuests')
+QUEST_LINES_ORDER_PATH = DEFAULT_QUESTS_BASE_DIR_PATH / 'QuestLinesOrder.txt'
+QUEST_LINES_DIR_PATH = DEFAULT_QUESTS_BASE_DIR_PATH / 'QuestLines'
+QUESTS_DIR_PATH = DEFAULT_QUESTS_BASE_DIR_PATH / 'Quests'
+EN_US_LANG_PATH = DEFAULT_QUESTS_BASE_DIR_PATH / 'en_US.lang'
 
+ID_LENGTH = 24
 
-class Thing(metaclass=ABCMeta):
-    def __init__(self, obj: dict, id_key: str, title: str, infix: str):
-        self.obj: dict = obj
-        self.id: str = obj[id_key]
-        self.name: str = obj['properties:10']['betterquesting:10']['name:8'].replace('\n', '%n')
-        self.desc: str = obj['properties:10']['betterquesting:10']['desc:8'].replace('\n', '%n')
-        self.title = title
-        self.name_key = f'gtnh.{infix}{self.id}.name'
-        self.desc_key = f'gtnh.{infix}{self.id}.desc'
-
-    def update_properties(self):
-        self.obj['properties:10']['betterquesting:10']['name:8'] = self.name_key
-        self.obj['properties:10']['betterquesting:10']['desc:8'] = self.desc_key
-
-    def write_to_lang_file(self, lang_file: TextIO):
-        lang_file.writelines([
-            '\n',
-            f'# {self.title}.{self.id} - {self.name}\n',
-            f'{self.name_key}={self.name}\n',
-            f'{self.desc_key}={self.desc}\n',
-        ])
+DIR_NAME_NO_QUEST_LINE = 'NoQuestLine'
+DIR_NAME_MULTIPLE_QUEST_LINE = 'MultipleQuestLine'
 
 
-class Quest(Thing):
-    def __init__(self, obj: dict):
-        super().__init__(obj, 'questID:3', 'Quest', 'quest')
+class BQJson(metaclass=ABCMeta):
+    def __init__(self, path: Path):
+        with open(path) as fp:
+            self.obj = json.load(fp)
+        self.path = path
+
+    @property
+    def name(self) -> str:
+        return self.obj['properties:10']['betterquesting:10']['name:8']
+
+    @property
+    def desc(self) -> str:
+        return self.obj['properties:10']['betterquesting:10']['desc:8']
+
+    @property
+    def desc_without_break_line(self) -> str:
+        return self.desc.replace('\n', '%n')
+
+    @property
+    def id(self) -> str:
+        return self.path.name.removesuffix('.json')[-ID_LENGTH:]
+
+    @property
+    def short_id(self) -> str:
+        return self.id.rstrip('=')
+
+    @abstractmethod
+    def write_to_lang_file(self, fp: TextIO):
+        pass
 
 
-class QuestLine(Thing):
-    def __init__(self, obj: dict):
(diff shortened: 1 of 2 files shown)


.github/scripts/questbook-i18n.py
```diff
@@ -1,60 +1,119 @@
...
-class Quest(Thing):
-    def __init__(self, obj: dict):
-        super().__init__(obj, 'questID:3', 'Quest', 'quest')
+    @property
+    def desc(self) -> str:
+        return self.obj['properties:10']['betterquesting:10']['desc:8']
 
+    @property
+    def desc_without_break_line(self) -> str:
+        return self.desc.replace('\n', '%n')
```
D-Cysteine, line 35: I believe that it is also necessary to replace '%' with '%%' (and do this before replacing '\n')

.github/scripts/questbook-i18n.py
```diff
@@ -1,60 +1,119 @@
...
+class QuestLine(BQJson):
+    @property
+    def quests(self):
+        return self.obj['quests:9']
+
+    def write_to_lang_file(self, fp: TextIO):
+        fp.write(
+            f'\n\n'
+            f'## Quest Line: {quest_line.name}\n'
+            f'betterquesting.questline.{quest_line.short_id}.name={quest_line.name}\n'
```
D-Cysteine, line 59: It might be a good idea to apply the replacement logic here as well, just in case some quest names contain '%'

.github/scripts/questbook-i18n.py
```diff
@@ -1,60 +1,119 @@
...
-from abc import ABCMeta
 
-DEFAULT_QUESTS_PATH = 'config/betterquesting/DefaultQuests.json'
-EN_US_LANG_PATH = 'config/txloader/load/minecraft/lang/en_US.lang'
-DEFAULT_QUESTS_US_PATH = 'config/betterquesting/DefaultQuests-us.json'
+DEFAULT_QUESTS_BASE_DIR_PATH = Path('config/betterquesting/DefaultQuests')
+QUEST_LINES_ORDER_PATH = DEFAULT_QUESTS_BASE_DIR_PATH / 'QuestLinesOrder.txt'
+QUEST_LINES_DIR_PATH = DEFAULT_QUESTS_BASE_DIR_PATH / 'QuestLines'
+QUESTS_DIR_PATH = DEFAULT_QUESTS_BASE_DIR_PATH / 'Quests'
+EN_US_LANG_PATH = DEFAULT_QUESTS_BASE_DIR_PATH / 'en_US.lang'
```
D-Cysteine, line 11: If the intention is to fully replace the `en_US.lang` generated by BQ, then I would suggest putting `en_US.lang` somewhere that is not in the `DefaultQuests/` directory. Possible ideas:

1. `config/betterquesting`
2. `config/txloader/load/betterquesting/lang`

The reason for this is the `DefaultQuests/` directory gets deleted every time `/bq_admin default save` is run, and if this file is no longer being generated by BQ, that will cause PRs to needlessly delete the file repeatedly only to have it get re-created by the workflow.

.github/scripts/questbook-i18n.py
```diff
@@ -1,60 +1,119 @@
...
+    @property
+    def quest_id_high(self):
+        return self.obj['questIDHigh:4']
+
+    def write_to_lang_file(self, fp: TextIO):
+        fp.write(
+            f'\n'
+            f'# Quest: {self.name}\n'
+            f'betterquesting.quest.{self.short_id}.name={self.name}\n'
+            f'betterquesting.quest.{self.short_id}.desc={self.desc_without_break_line}\n'
```
D-Cysteine, line 81: Same comments re: formatting, for the quest text

.github/scripts/questbook-i18n.py
```diff
@@ -1,60 +1,119 @@
...
+    def __init__(self, path: Path):
+        super().__init__(path)
+
+    @property
+    def quest_id_low(self):
+        return self.obj['questIDLow:4']
+
+    @property
+    def quest_id_high(self):
+        return self.obj['questIDHigh:4']
```
D-Cysteine, line 74: You could probably just use `self.short_id` rather than a tuple of these values. The `short_id` is guaranteed to be unique. Though it is true that this will break if we ever remove the ID from the file names

Up to you if you want to change this. It should be fine either way, just maybe simpler if we don't need a tuple

D-Cysteine (commented):
This is great! Thank you for updating this workflow!

Is the intention to replace BQ's generation of `en_US.lang` with this? I.e., we can turn off BQ generating `en_US.lang`, and just rely on this workflow to do it? If so, I would suggest moving `en_US.lang` out of `DefaultQuests/` (see my comment) to reduce change churn. I do note that the order of quests will be different with this workflow vs. the current BQ code, so we probably shouldn't keep both as it would result in a lot of noise in PRs.

Also, one potential idea is, once we have translated lang files, we could update this script to automatically update those lang files as well, keeping the same ordering as in `en_US.lang` and automatically moving existing translation entries and marking new translation entries. I would be happy to take a look at implementing this, when I have time (and once a translation file other than `en_US.lang` exists).

MuXiu1997:
> Is the intention to replace BQ's generation of en_US.lang with this? I.e., we can turn off BQ generating en_US.lang, and just rely on this workflow to do it? If so, I would suggest moving en_US.lang out of DefaultQuests/ (see my comment) to reduce change churn. I do note that the order of quests will be different with this workflow vs. the current BQ code, so we probably shouldn't keep both as it would result in a lot of noise in PRs.

Yes, using this workflow to generate en_US.lang can prevent situations where json files are submitted without en_US.lang. We can turn off BQ's generation of en_US.lang, and you can decide where to place en_US.lang.

> Also, one potential idea is, once we have translated lang files, we could update this script to automatically update those lang files as well, keeping the same ordering as in en_US.lang and automatically moving existing translation entries and marking new translation entries. I would be happy to take a look at implementing this, when I have time (and once a translation file other than en_US.lang exists).

We also have some automation for our Chinese language files, so they will automatically maintain the same order as en_US.lang.

.github/scripts/questbook-i18n.py
MuXiu1997: Changing the directory where en_US.lang is located would require modifying the BQ code, so it's up to you to choose a path.

.github/scripts/questbook-i18n.py
MuXiu1997: done

.github/scripts/questbook-i18n.py
MuXiu1997: done

.github/scripts/questbook-i18n.py
MuXiu1997: done

.github/scripts/questbook-i18n.py
MuXiu1997: I haven't read the new code for BQ, but if questIDLow + questIDHigh works, I would prefer to keep this method. If someone mistakenly submits a JSON file, questIDLow + questIDHigh can serve as a simple check. It's like using QuestLinesOrder.txt to also check if the JSON files in the QuestLines directory can be correctly matched.

.github/scripts/questbook-i18n.py
D-Cysteine: Once BQ code is modified to no longer generate `en_US.lang`, wouldn't it be determined by line 11 here?

MuXiu1997: > 5. Just drop the translated lang file into config/txloader/load/betterquesting/lang to apply the translations. In previous versions of BetterQuesting, it was necessary to load a custom version of the quest book data for translations to work, but this is no longer needed.

Oh, I forgot that all the translated files are in the `config/txloader/load/betterquesting/lang`. Alright, then let's put `en_US.lang` in there as well.

MuXiu1997: done

D-Cysteine (approved):

PR #13121, opened 2023-04-03
Updates needed for new `lang` file handling

D-Cysteine:
* Updated workflow to export to `template.lang` and have a consistent format with the BQ code
 * Moved `en_US.lang` to `template.lang` and regenerated its contents
 * Updated translation instructions

I renamed the `lang` file to `template.lang` instead of `en_US.lang` so that it won't get automatically applied to anyone editing the questbook in English, as this would mask changes made to the quest text and potentially be very confusing.

This PR is also needed to make the `lang` file contents consistent:
 * https://github.com/GTNewHorizons/BetterQuesting/pull/103


Dream-Master (approved):

D-Cysteine:
Do you mean in `template.lang`? If an ID appears more than once, then yes, that would indicate a bug somewhere. However, I only found one occurrence of the ID that you mentioned:
 * https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/blob/c67683089eeb0c2e12054864e89e833f345a6e72/config/txloader/load/betterquesting/lang/template.lang#L11179

Do you see another occurrence somewhere?

MuXiu1997:
Please disregard what I said earlier. I was checking issues with the automation of the Chinese translation file and did not realize that I had turned off "case sensitive matching".

PR #13147, opened 2023-04-06
Update villageBlacksmith.xml

Rika-TH:
adds the "gt" the the item name for the "gregtech:gt.blockmachines"


Dream-Master:
@Rika-TH did you test if the name is right via /mt hand  ?

Rika-TH:
I couldn't find a machine in the blacksmiths and machines ran with mt hand are called gregtech:gt.blockmachines instead of the original gregtech:blockmachines (it would seem most gregtech things have the prefix gt.

Dream-Master:
ok

Dream-Master (approved):

Issue #13036, opened 2023-03-25
Monthly Ticket for Quest Changes


chochem:
### Your GTNH Discord Username

chochom#0271

### Your Pack Version

2.3.1-2.3.2

### Overview

A ticket for any large or small problems with any quests. Just post below. Of course, you can also make a dedicated ticket. In particular for more indepth suggestion or problems.

The most up-to-date quest files can always be found here: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/tree/master/config/betterquesting/DefaultQuests
Depending on your version, you might also need to update BQ to use them.

Below are the remaining things from last month (https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/12495):

### Outstanding Quest Changes

- [ ] https://github.com/GTNewHorizons/GT5-Unofficial/pull/1146 reworked turbine materials, so quests related to turbine materials need to be checked to confirm if any changes are needed.
- [ ] **Quest.3165 UV Power Per Turbine**
  > Large Ichorium 170% 342,720 EU/t (very durable)
  Ichorium turbine has been buffed. So this data is outdated.
- [ ] **Quest.3166 Powering-up Fluid Nukes**
  > Oriharukon and MAR-Ce-M200 are your best choices.
  MAR-Ce-M200 is LuV (need ZPM tier EBF) now. See https://github.com/GTNewHorizons/GoodGenerator/blob/33681ff5f240800c70b70c1e314d73ae262caf7c/src/main/java/goodgenerator/loader/RecipeLoader_02.java#L1145
- [ ] **Quest.2568 Infinity Helmet** and **Quest.2569 Infinity Breastplate**
   Infinity Helmet can't remove all negative debuffs, it's the ability of Infinity Breastplate.
- [x] Quest.1510 Biomes Everywhere, But Not the One You Want? Overhaul based on https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11385#issuecomment-1257110174
- [x] there are 2 vanadiumsteel quests (LV and MV). remove one. the remaining one should be in LV. 
- [ ] LV cobalt from cobaltite quest: make it about both cobalt and arsenic, put it before the GaAr cystal (though optional). also make first task optional.
- [x] quest 'power up fluid nukes' has a broken task
- [x] "Walk Through Fire" quest completes with normal Pig Adventure Backpack
- [x] remove OSS quest. its super buggy and not needed for anything.
[...]

PR #13323, opened 2023-04-25
Quest Improvements - 25 April

chochem:
- add new quest for algae farm
- add new quest for coal
- add new quest for aal
- move hssg quest to ev and reduce amount to what is needed there
- fix tungstate step 1 quest. needs sodium not sodium hydroxide.

all from https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13036


chochem:
well I will need to merge that properly because we actually moved around quests in the same tab. this one goes first: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/13313

PR #13351, opened 2023-04-29
Fix README typo

chill-was-taken:


chill-was-taken:
As I was reminded, making PRs to this Readme is incorrect because it's assembled automatically from https://github.com/chill-was-taken/DreamAssemblerXXL

PR #13357, opened 2023-04-29
docs(quest): Rubber Sheets: mention Piston Boots

chill-was-taken:
Rephrase the Rubber Sheets quest to mention Piston Boots.

Also clarify the main usage of Rubber Sheets at this stage.

Remove the mention of other things that are crafted not as often.


chochem (changes requested):
cheaper? you just mean steam or what? pretty sure its the same ratio.

instead of 'look up the quest' better name the questline they should check out!

chill-was-taken:
Yes, I meant steam.
Tin Cable in Alloy Smelter is 700 EU ULV in 5 sec.
Rubber Sheet in Alloy Smelter is 60 EU ULV in 0.5 sec. Rubber Sheet is then manually crafted with Tin Wire at no cost.
Both recipes spend 2 rubber bars.

chochem (approved):
thanks

chill-was-taken:
Added "in EU" in the last force push to clarify your first point.

chochem:
maybe 'power'? since this is still steam age

chill-was-taken:
True, one second.

PR #13361, opened 2023-04-29
docs(quests): BetterQuesting Readme: fix typos and improve wording

chill-was-taken:
Explaining lines 90-91:
Swap and reword the items 6 and 7 to change the intention from 
"a quest can be in multiple tabs but only if necessary" 
to
"a quest really should be in one tab, but can be in many if really necessary"

The rest is minor rewording and typo fixes.


chochem (approved):
sure

PR #13362, opened 2023-04-29
docs(quest): Biomes Everywhere: add more info

chill-was-taken:
Add more information about how the Nature's Compass works.

Done by following the BQ Readme and then spellchecked in JSON.

from https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13036


chochem (approved):
thanks. I changed the link in the PR text to the active ticket and checked it off there.

PR #13363, opened 2023-04-30
docs(quest): CuttingaTreelogb: improve wording and formatting

chill-was-taken:
* Decolor orange because it's not a good match with the default background color.
* Decolor dark_aqua because it's just another tip.
* Explain how to make netherrack parts because they are not shown in NEI.
* Move the parts about upgraded materials and quest detection closer together to the end.


chochem (approved):
ok

PR #13364, opened 2023-04-30
docs(quest): RubberSheets: update and reword

chill-was-taken:
* Update the reference to the Getting Around questline because it was split in two.
* Simplify wording.


chochem (approved):
sure whatever

PR #13367, opened 2023-04-30
docs(quest): MushroomBiomes: explain sunflower and reword

chill-was-taken:
* Decolor and move the code snippet for consistency.
* Explain that RWG is different from the vanilla worldgen. In particular Sunflower Plains are not generated.


chochem:
I dont believe that belongs there. nor that it is particularly relevant.

PR #13370, opened 2023-04-30
Quest: BiomesEverywhere: add that compass shows not-generated biomes

chill-was-taken:
The commit adds a single line in the quest.

In its default state, the Nature's Compass shows a full list of biomes that contains even the ones not generated by Realistic World Generator that GTNH uses.

A prominent example of that is Sunflower Plains. A user that wants to find Sunflowers might find vanilla info that they are located only in Sunflower Plains, although they are scattered across other biomes because the Sunflower Plains are not generated by RWG.

That might result in long journeys for nothing, which is frustrating. Let's prevent that frustration by adding the info about them in this quest.


chochem (approved):
ok

PR #13375, opened 2023-04-30
Quest: JustLikeMomUsedt: mention kitchen multiblock

chill-was-taken:
Mentioning only Forestry Worktable as a way to store food recipes is questionable because the kitchen multiblock is overwhelmingly better for food in this regard.


chochem (approved):
sure whatever

PR #13378, opened 2023-04-30
BQ Readme: add how to order quests

chill-was-taken:
Also add a warning about advanced values.
An error in editing advanced values can lead to a crash.


chochem:
ideally you should make a ticket about the crash, not a warning to avoid it

chill-was-taken:
Will do. As it is now, a warning is the least I can do.

chill-was-taken:
The ticket about the crash is created.
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/13379

Because we don't know how long it will take to address that, I am still in favor of the warning the proposed commit.

chochem (approved):

PR #13385, opened 2023-04-30
Quest: Glue: add glue option

chill-was-taken:
The quest mentions refined glue as an option to make Good Circuits. The usual glue can also be used to make Good Circuits.

What is more, manually farming glue from Slimy Trees is faster and easier than manually farming Sticky Resin from Rubber Trees.

If I understood the quest intention correctly, it is made to prepare a player for the creation of Good Circuits. If this is so, overlooking glue farming from Slimy Trees is an unaffordable luxury.

Therefore, I added glue as an option:
* Make Requirements disjunctive (OR) and add Fluid Extractor as a requirement that is visible on hover.
* Add Congealed Green Slime as an Optional Retrieval. The number to retrieve is 11 to give 6kL Glue, which is equivalent to 3kL Refined Glue if using Distillery.
11 Congealed > (Extractor) 44 Smlimeballs > (Fluid Extractor) 6336L of Glue.

All retrieval parts of the quest were tested.


chochem (approved):
why not

PR #13393, opened 2023-05-01
Quest: DivineRodoftheGo: update and reword

chill-was-taken:
* Specify that small ores cannot be found with the usual ore.
* Decolor paragraphs and integrate them into the text.
* Rephrase and reformat the text with the aim to be concise.


chochem (approved):
looks fine

PR #13467, opened 2023-05-09
Template crash markdown

Ethryan:
If this works, it should markdown the crash log tab in the template for Crash Reporting.

<details>
  <summary>Log</summary>
  
I grew tired of walls of text in reported crash issues.
  
</details>

Needs testing.

### DIFF
--- a/.github/ISSUE_TEMPLATE/000-report-crash.yml
+++ b/.github/ISSUE_TEMPLATE/000-report-crash.yml
@@ -72,9 +72,9 @@ body:
   id: report
   attributes:
     label: Crash Report
-    label: <details><summary>Log</summary>
+    value: <details><summary>Log</summary>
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/000-report-crash.yml
```diff
@@ -72,7 +72,9 @@ body:
   id: report
   attributes:
     label: Crash Report
+    value: <details><summary>Log</summary>
     description: "If no crash report is created, attach `logs/fml-client-latest.log` or `logs/fml-server-latest.log`, depending on if it's a dedicated server crash or a client crash. You can also type in additional info that you consider important."
+    value: </details>
```
wlhlm, line 77: You are not allowed to have duplicate keys in YAML, unfortunately.

wlhlm (changes requested):

.github/ISSUE_TEMPLATE/000-report-crash.yml
Ethryan: So this won’t work then…

PR #13625, opened 2023-06-06
mob repellator quest

Gofer2x:
image
image


Gofer2x:
ok i just realized the repellator V is missing oops lol

PR #13641, opened 2023-06-09
Update PreciseAssembler-AAAAAAAAAAAAAAAAAAAMEg==.json

Reflex18:
Changed requirement for Precise Assembler from iv energy hatch to Luv energy hatch.

This is due to the material of MAR-Ce-M200 Steel requiring a ZPM tier EBF which needs two LUV tier energy hatches.

image

image


chochem (approved):
technically you can make it an mega, but yea this is good

PR #13704, opened 2023-06-14
Small clarifications to ME p2p transport options

greesyB:
Removes text stating that you must use a different network to transport ME p2p as it is both false and confusing to players new to AE, added a little more info on how to transport them instead.

Before and after for ease of viewing changes (first paragraph changed, second expanded upon and made into third):
image
After
image


chochem (approved):
yep thats better

Issue #11814, opened 2022-11-11
Extreme Extermination Chamber won't spawn Wisp and Hellfire Wisp 


fluffle:
### Your Pack Version

2.2.3

### Your Server

Local MP

### Type of Server

Vanilla Forge

### Your Expectation

I wanted to have a renewable source of Thaumcraft Essentia, so I created an EIO powered spawner and combined it with a Hellfire Wisp spawner I liberated from the nether. I expected this to produce wisp drops when placed in the Extreme Extermination Chamber (EEC) slot.

### The Reality

Placing this spawner in the EEC's slot does nothing. The only thing that drops essence from the EEC is the Eldritch Guardian, according to NEI.

### Your Proposal

Add Wisp and Hellfire Wisp to the list of mobs that work with the EEC.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

Alastors:
Intentional, they have some issues that prevent this, closing

kuba6000:
> Intentional, they have some issues that prevent this, closing

It's not.

Alastors:
> > Intentional, they have some issues that prevent this, closing
> 
> 
> 
> It's not.

What do you mean it's not, we've been over this before and last time you said it was too much overhead for the wisps because of their unique generation mechanics that cause the incompatibility with the EEC since they have to generate->generate aspects->die to drop their actual ethereal essence, which doesn't work in the EEC

PR #13717, opened 2023-06-15
Add Wisp drops to NEI info

kuba6000:
Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/11814
Generated using https://github.com/GTNewHorizons/KubaTech/commit/6aa44640ef243bdafdffa61915dab0e6813f03fb
image
image
image
image


Alastors (approved):
I'm going to approve this, but I really wish you would change the primals to be less than 15%, maybe you could put them at 5%? 

I know it's what thaum does in the base mod, but the base mod also didn't anticipate 67 total aspects with 6 more to come, and even just reducing them to 5% each would bring the rest up to just around 1%, as it should be quite frankly.

Alastors (changes requested):
At further thought, no this certainly needs to have the primals switched to 5% each with the remaining 70% split about the remaining 61 aspect, otherwise players are going to continue to use the already intensely laggy wisps spawning methods that the condenser is left in the pack to combat.

chochem:
> At further thought, no this certainly needs to have the primals switched to 5% each with the remaining 70% split about the remaining 61 aspect, otherwise players are going to continue to use the already intensely laggy wisps spawning methods that the condenser is left in the pack to combat.

as said already if you want a buff you should provide numbers. just adding a random buff over base numbers is weird. I would say we can dismiss this. can always comeback to it later if someone actually does a comparison to a normal wisp farm.

Alastors:
> > At further thought, no this certainly needs to have the primals switched to 5% each with the remaining 70% split about the remaining 61 aspect, otherwise players are going to continue to use the already intensely laggy wisps spawning methods that the condenser is left in the pack to combat.
> 
> as said already if you want a buff you should provide numbers. just adding a random buff over base numbers is weird. I would say we can dismiss this. can always comeback to it later if someone actually does a comparison to a normal wisp farm.

okay well let's look at the numbers, 15*6 is 90, that leaves 10% for the remaining 61 aspects which is 0.16% per aspect, wisps only drop 1 aspect and are not affected by looting.

Kuba has directly stated the EEC spawns 48 wisps a second at UHV which means 43 wisps a second would be primal, which means 5 wisps a second that are non-primal spread across 61 aspects.

The planar vortex can be made in MV with enough spite, and spawns 2-8 wisps of the aspects found within the vortex per 1 salis mundis.

You do the math to what players are going to use

kuba6000:
> At further thought, no this certainly needs to have the primals switched to 5% each with the remaining 70% split about the remaining 61 aspect, otherwise players are going to continue to use the already intensely laggy wisps spawning methods that the condenser is left in the pack to combat.

This is not going to happen. Chances are shown in NEI page and not only for EEC, its for general information about the mob so it should not be lying. Also if you really want to buff the EEC then make the Wisp not always infernal (8x buff to EEC) or nerf other methods (maybe they are too powerfull)

Alastors:
> > At further thought, no this certainly needs to have the primals switched to 5% each with the remaining 70% split about the remaining 61 aspect, otherwise players are going to continue to use the already intensely laggy wisps spawning methods that the condenser is left in the pack to combat.
> 
> This is not going to happen. Chances are shown in NEI page and not only for EEC, its for general information about the mob so it should not be lying. Also if you really want to buff the EEC then make the Wisp not always infernal (8x buff to EEC) or nerf other methods (maybe they are too powerfull)

Kuba, what do infernals drop that would make you say that? Is it the ender pearls? the blaze powder? the chainmail? perhaps the pistons?

chochem:
he means it would be faster ;)

Alastors:
> he means it would be faster ;)

So 40 non-primals a second at UHV compared to 2-8 of the exact type you want per salis at MV, you do realize what's wrong with that right?

Alastors:
I should make a correction here, Kuba has confirmed it actually only makes 15 wisps a second at UHV also known as 1-2 non primals a second

if infernals were removed it would only be 124 wisps a second at UHV, also known as 15 non primals a second 


At **_UHV_**

This fixes nothing and shouldn't get through until it does

boubou19:
Do we really want to tweak spawn ratios for a single mob type tho? Because i have understood that Kuba copied the spawn chances of the wisps. Imo relying on mob drops for industrial quantities is a game design mistake. A multi smelter like multiblock but for decomposing items into aspects would work better.

fluffle:
I appreciate this being worked on but if it is causing disagreements it may be better to drop my feature request in favour of pushing people towards the Euclidaisy from Forbidden Magic, now that Botania is a thing in GTNH.

kuba6000:
> I should make a correction here, Kuba has confirmed it actually only makes 15 wisps a second at UHV also known as 1-2 non primals a second
> 
> if infernals were removed it would only be 124 wisps a second at UHV, also known as 15 non primals a second
> 
> At **_UHV_**
> 
> This fixes nothing and shouldn't get through until it does


> I appreciate this being worked on but if it is causing disagreements it may be better to drop my feature request in favour of pushing people towards the Euclidaisy from Forbidden Magic, now that Botania is a thing in GTNH.

Regardless, this pr still fixes the empty pages on NEI and informs a player what could they get from a mob. It's not just about EEC.

Alastors:
> Do we really want to tweak spawn ratios for a single mob type tho? Because i have understood that Kuba copied the spawn chances of the wisps. Imo relying on mob drops for industrial quantities is a game design mistake. A multi smelter like multiblock but for decomposing items into aspects would work better.

We already have this and it's absolute shite

Alastors:
> > Do we really want to tweak spawn ratios for a single mob type tho? Because i have understood that Kuba copied the spawn chances of the wisps. Imo relying on mob drops for industrial quantities is a game design mistake. A multi smelter like multiblock but for decomposing items into aspects would work better.
> 
> We already have this and it's absolute shite 

No one uses it because it sucks

Alastors:
I'm going to point this out again, people are already relying on a simple cheap and easy source of essentia through wisps at MV or earlier that certain variations can even give you the exact ethereal essence you need.

We've made a lot of mistakes in the magic development over 8 years that have lead to the condenser being the only non laggy source for most people, but we're trying to fix this, and this doesn't even get close

Alastors:
> I still don't see why mob farms have to be the best way of generating essentia in the entire pack. 

... steelux they already _are_

> Why not add aspects to items that don't have them yet, and improve essentia automation from items that way? You won't need to change multis that are unrelated to magic like the EEC.

You know on the other side of the argument you could just tell people to stop adding aspects they think are "hard" because all essentia has the baseline ease of mana beans, wisps, and etc

Dream-Master (approved):

combusterf (approved):
Is it an improvement over the status quo and is it worth including? yes.

Does it solve the larger problem? maybe not as much as we'd like.

combusterf:
Honestly I think the talking points here are quite valid, but I think the discussion as a whole feels a bit like bikeshedding at the cost of the feature itself. 

The numbers are not static, let's head back to #magic-dev and decide as a team if we want to improve the odds, and if/how we want to give players an opportunity to be smart (many options here. $.02: why not retain the wisp's aspect in the process of you jarring one and getting it in a spawner?)

Rakdarian:
Its really easy to build a Salis farm, and you can dump stacks at a time into a planar vortex.
where you get 2-8 wisps per salis. Cant you spawn thousands of wisps at a time if you wanted to.

boubou19 (approved):

Dream-Master:
@Alastors as i wrote you majority wins.
If you have additional ideas make a new pr

Alastors:
> @Alastors as i wrote you majority wins.
> 
> If you have additional ideas make a new pr

Yeah that's fine lad

Alastors:
It'll be grand

PR #13768, opened 2023-06-21
change standart key not longer using w

Dream-Master:


Dream-Master:
change refresh rate to 2. it is the default value

minecraft7771 (approved):

PR #13782, opened 2023-06-23
Fix pollution config values

chochem:
The pollution mixins where moved from modmixins to hodgepodge. However our config values got lost in the process.

This PR puts the values back to exactly what we had before the refactor. (the old values can be found here https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/blob/master/config/ModMixinsPlugin.cfg)

edit: also removes the outdated modmixins config.


wlhlm:
Can you also remove the pollution config from modmixins config file?

chochem:
that mod is removed. is that config used in any way?

wlhlm:
Yeah, that can be yeeted.

wlhlm (approved):

PR #13816, opened 2023-06-27
New quest to showcase Tesla tower

Reflex18:
Added in a new quest to storing and transforming to showcase the Tesla tower. The tower is a powerfull wireless energy transfer option that scales from ev all the way to ZPM.


chochem (approved):
cool, thanks!

PR #13887, opened 2023-07-05
Hide Hardness and Blast Resistance values in Waila

DmitryScaletta:
I think this information is useless in 99.9% of cases but it still takes space and distracts from really important stuff.

<details><summary>Before/After</summary>

image

image
</details>

To Enable it back:
[Waila] Config screen (NUMPAD0 by default) -> Modules -> Wawla-General -> Set "Show Block Hardness?" and "Show Blast Resistance?" to Yes.

### DIFF
--- a/config/Waila.cfg
+++ b/config/Waila.cfg
@@ -126,8 +126,8 @@ modules {
     B:wawla.harvest.showTool=true
     B:wawla.horse.showJump=true
     B:wawla.horse.showSpeed=true
-    B:wawla.info.showHardness=true
-    B:wawla.info.showResistance=true
+    B:wawla.info.showHardness=false
+    B:wawla.info.showResistance=false
     B:wawla.light.lightLevel=false
     B:wawla.light.monsterSpawn=true
     B:wawla.light.showDay=true


boubou19:
and when you want to know the hardness in game how do you do?

Alastors:
Why wasn't this a localized change? I feel like there are more things that spam info than 2 lines total?

DotJason:
While not particularly useful, personally I don't think that blast resistance and hardness should count as spam. It's good that the main WAILA panel got uncluttered, but maybe this should still be displayed when looking at blocks while holding Shift? If WAILA has that functionality (pretty sure it does) and it's not too hard to put this info there, I think that's reasonable.

And IMO leaving this to the config makes little sense, I don't know if anyone would be bothered by this so much that they would enable it just to look at some block's blast resistance/hardness. Putting a feature in config and then disabling it by default practically makes sure that the vast majority of players aren't going to use it (of course, cue the messages in the "help" channel on Discord asking about some block's blast resistance, inviting the same "enable it in the WAILA config" response).

boubou19:
i predict a wave of people asking how to turn it back. Between the players against any aesthetical change, the players who will feel weird about the new look, the ones who will have to turn it back on after each update because they use it... @Dream-Master we should have done a vote first, because forcing such a change on the whole community without feedback will just cause a wave of hate. Especially since this can be toggled on/off at the user's preference, if those two lines are bothersome for a few people, they should be the ones who disable them in their instance.

Dream-Master:
<https://discord.com/channels/181078474394566657/465207956745486336/1126547064898195456>

ZuOverture:
Can't it be modified to please both parties? For example show H/BR if shift is pressed, don't otherwise. Or add two more unexplained numbers to the topmost line, it has a couple already.

minecraft7771:
Both parties should already be pleased by there being an option to set it in your client which takes less than a minute.

Alastors:
> Can't it be modified to please both parties? For example show H/BR if shift is pressed, don't otherwise. Or add two more unexplained numbers to the topmost line, it has a couple already.

Honestly personally I like this proposal myself

704009905:
I personally give less info on my server, to the extent that WAILA only provides names to help players know new blocks. But I will vote for keeping it unchanged(up).
This is always an *option* in GTNH, so players always have this choice on their own (and their server admin). What the pack provides should confine to be a default, which from what I'm seeing the old settings had a more player base.

Shift-toggling is a good feature addition though, WAILA already has shift-toggling for harvestability-related things.

PR #13893, opened 2023-07-07
Three new quests for some Hatches and buses

Reflex18:
One quest in Ev tab for the Quad input fluid hatch.

Two quests in the IV tab with one for the linked input bus and the other for the Super buses.


chochem (approved):
thanks!

PR #13964, opened 2023-07-16
Add NEI-Integration config

DotJason:
Another way to do what [this](https://github.com/GTNewHorizons/NEI-Integration/pull/7) was supposed to do, as per request from @glowredman.


glowredman (approved):

AnrDaemon:
Just a hint. You could add only important options to template configuration, and leave everything else to the magic of config parser.

PR #13971, opened 2023-07-16
Specify Config Properties which shouldn't be reset when GTNH is updated

glowredman:
Only useful if we decide to use the [Default Configs](https://github.com/GTNewHorizons/Default-Configs) mod.

Also removes some config files of mods we no longer use.


glowredman:
I haven't fully understood it myself, here is what I know:

At the end of the client startup routine (`net.minecraft.client.main.Main#main(String[])`), `net.minecraft.client.Minecraft#run` is called, which immediately calls `et.minecraft.client.Minecraft#startup`. This method usually first sets up the game setting but DC [injected its own method](https://github.com/GTNewHorizons/Default-Configs/blob/master/src/main/java/net/blay09/mods/defaultkeys/coremod/DefaultOptionsClassTransformer.java#L22-L51) at the beginning of `startup()`, which does the following:
1. If no `./options.txt` file exists, [create a default one](https://github.com/GTNewHorizons/Default-Configs/blob/master/src/main/java/net/blay09/mods/defaultkeys/EventHandler.java#L216-L235) using `./config/defaultoptions.txt`. *Not using this*
2. If no `./optionsof.txt` file exists, [create a default one](https://github.com/GTNewHorizons/Default-Configs/blob/master/src/main/java/net/blay09/mods/defaultkeys/EventHandler.java#L237-L253) using `./config/defaultoptionsof.txt`. *Not using this*
3. `./config/localconfig.txt` is [copied from the JAR, if it doesn't yet exist](https://github.com/GTNewHorizons/Default-Configs/blob/master/src/main/java/net/blay09/mods/defaultkeys/EventHandler.java#L57-L67).
4. `./config/localconfig-example.txt` is [copied from the JAR, if it doesn't yet exist](https://github.com/GTNewHorizons/Default-Configs/blob/master/src/main/java/net/blay09/mods/defaultkeys/EventHandler.java#L68-L78).
5. If `./config/modpack-update` exists, the [local configs are restored](https://github.com/GTNewHorizons/Default-Configs/blob/master/src/main/java/net/blay09/mods/defaultkeys/EventHandler.java#L149-L214) and `./config/modpack-update` is [deleted](https://github.com/GTNewHorizons/Default-Configs/blob/master/src/main/java/net/blay09/mods/defaultkeys/EventHandler.java#L82-L85)<sup>1</sup>. Otherwise a [backup](https://github.com/GTNewHorizons/Default-Configs/blob/master/src/main/java/net/blay09/mods/defaultkeys/EventHandler.java#L92-L147) of all properties defined by `./config/localconfig.txt` is created in `./localconfig.cfg`.
6. The first time the main menu is reached, the [keybindings are set](https://github.com/GTNewHorizons/Default-Configs/blob/master/src/main/java/net/blay09/mods/defaultkeys/EventHandler.java#L311-L376) to the values defined in `./config/defaultkeys.txt`, unless they are in `./knownkeys.txt`.

Any file used to safe information specific to the pack's instance is not saved in the `./config/` directory but in the root pack directory. Thus, the pack can be updated as before and everything defined as local is restored at startup. Obviously, this only fully works with the next update *after* Default Configs is included because no backup can be created if the instance isn't run with the mod.

<sup>1</sup> This section is currently commented out but I don't know why. This needs further investigation.

chochem:
so there is a `localconfig.cfg` and a 'knownkeys.txt' in the main folder that the player should keep from their old instance basically? Please add that to install wiki then after this is in.

I dont understand what `/config/localconfig.txt` and `config/defaultkeys.txt` are supposed to do though as they arent carried over.

chochem:
its not just keeping. its also copying over when migrating

Dream-Master (approved):

PR #13986, opened 2023-07-18
fix #10531

### LINKED ISSUE #10531: Thaumonomicon UI rescale by itself on open/close and windows resizes

Tralomine#5809

### Your Pack Version

2.1.2.3qf

### Your Server

private server

### Type of Server

Vanilla Forge

### Your Expectation

the thaumonomicon UI is large and doesn't change by itself is no resize action or ui scale is done by the player

### The Reality

the UI size change depending on multiple things
by default it starts small, but if you resize the windows to larger than fullscreen then back to fullscreen, it will take the entire screen, or if you change the UI scale to something else then back to the ui scale you had just before the change (so everything should be the same), or if you change some TC4 tweaks render parameters
then, once it take the entire screen, if the UI closes and reopen (if the windows is resized (even by a few pixels), if the thaumonomicon is closed then reopened, or if any research is opened then closed) the ui will slightly shrink, and shrink again, and again, until it goes back to the initial small (and not very useable) resolution.
basically this mean that the size of the previous time the thaumonomicon was opened is saved somewhere and used in the calculations of the new windows size, which *shouldn't be the case at all* since the user has no way of resizing it

https://youtu.be/J7Di8QSVY_8 

this apply with large ui scale at 1920x1080 resolution, for normal ui scale at a resolution slightly smaller (and for small resolutions something even smaller, it probably apply at larger scale on larger screens) 

this issue is basically the same *but* the bug and its causes weren't clearly identified and the reason not to fix doesn't actually apply, ie the calculation *should* always be the same at the same windows size, and *should not* depend on the previous size of the ui
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/8381

note: this is *not* quick successions, if I don't open the book for hours it will just shrink one step when I open the book, even if it's just once and don't open it for 5 more hours after that 

### Your Proposal

make the ui only depend on the screen size and take the maximum amount of space, which is *a constant value* and *do not change* when the windows size don't change

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [X] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

Tralomine:


Dream-Master (commented):
Why set it to false ?

This is how we use it in GTNH

    browser_scale {
        # The search result area, even if it's not disabled, will be considered while inferring browserScale.
        # [default: true]
        B:considerSearchArea=true

Tralomine:
because it's not calculated properly and causes issues, see #10531

Dream-Master:
@Tralomine you can close this or and wait for gleese make a fix.

PR #14012, opened 2023-07-21
Three quests for Thorium232 line

Reflex18:
Three new quests outlining the options to obtain thorium 232

Location:
image

EBF side:
image

Chemical plant side:
image


chochem (changes requested):
thanks a lot. Got a few questions and comments though.

- Why are they locked behind the ABS? I dont see any reason for that.
- Capitalization is all over the place. You also use three variants of thorium 232/Thorium 232/Thorium-232. I would say the standard is Thorium 232, Boric Acid, EV tier, Neutron Activator, etc.
- I guess I am insane for using the LCR/EBF route. (that can stay though, nothing to change here)

Reflex18:
Thanks for the response back. 

The reasoning for the ABS lock was because I assumed that the main reason for Thorium 232 was for the Molten Arcanite for making a Large processing factory. However on looking at the recipe for the Molten Arcanite again I do see that it can be make in a mixer as well.

Because of that I will change the requirement to Tier 5 (IV).

Also yes the six step processing chain for the EBF side is crazy compared to chemical plant 2 step.

chochem (approved):

PR #14013, opened 2023-07-21
AE2 16384k fluid cell quest

Reflex18:
Adding in a few quests for the next tier of fluid cell for ae2.

image


chochem (approved):
thanks.

PR #14014, opened 2023-07-21
Yot tank quest

Reflex18:
A quest to explain the Yot tank


chochem (changes requested):
You are saying Yot tank and Yotta tank. But it's really called the YOTTank, isn't it?

chochem (approved):

Issue #14025, opened 2023-07-23
Quest "How to Get Sodium and Lithium"


Maximnety:
### Your GTNH Discord Username

Maximnety

### Your Pack Version

2.3.2

### Your Server

SP

### Java Version

Java 17

### Type of Server

Single Player

### Your Expectation

Quest should required 32x primary ore and 16x sporadic ore

### The Reality

![2023-07-23_11 47 02](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/84129980/6f366cfe-5456-4f76-9a52-4cbefc0ca591)
Rock salt ore - primary ore
Spodumene ore - sporadic ore
But quest required 16x Rock salt ore (primary) and 32x Spodumene ore (sporadic)

### Your Proposal

Change quest - set requirements of Rock salt ore - 32, set requirements of Spodumene ore - 16
![2023-07-23_11 52 21](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/84129980/9f5cc7d9-1041-43bd-841c-fb725522d3d6)

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [X] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

Dream-Master:
@Maximnety just push the change as pr if you like

PR #14027, opened 2023-07-23
adjust how to get lithium quest

Pxx500:
resolves https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14025


Pxx500:
image

Pxx500:
only spodumene is 16 because it's the rarest, I also changed the order of ores so it's the same as in the vein info in NEI

kuba6000 (approved):
:tulip:

PR #14123, opened 2023-08-01
Minor change to Wireless hatch quest

Reflex18:
Added in some extra information, changed the main requirement to better suit the tier and highlighted the commands to active the wireless power transfer.


chochem (changes requested):
For the first change: that is not actually accurate. there is a bunch of multis that can take 8 normal hatches. Well, prior to GPL they could. who really knows now but supposedly its unchanged.

I assume you are thinking of the AL so maybe talk about that specifically instead.

Also have to say this bright light blue is pretty annoying to read on the default theme. best avoid it.

For the second change: the ulv ones werent just a 'reminder' but they are quite literally what you should make. so this is all a bit strange. at least the it should be the first task and not an afterthought as the second.

Reflex18:
I was chatting with a few late game players about the later game quests and they advised to take a look at this quest.

I should of instead kept to the tiers that I understood rather then relying on second hand information. Sorry about the changes.

PR #14230, opened 2023-08-14
Fix wand cap in lootbag

chochem:
was giving the old (and unusable) dreamcraft one instead of thaumcraft.


kuba6000 (approved):

Alastors (changes requested):
Yeah we def need this, we probably need to mark the dreamcraft cap items for termination as well to stop the confusion

Alastors (approved):
Yeah we def need this, we probably need to mark the dreamcraft cap items for termination as well to stop the confusion. (Hit the wrong button, freaking mobile)

PR #14253, opened 2023-08-18
Oil quest adjustments

Pxx500:
image
Deleted the part that suggested the player to take a certain route

image
image
Fixed typo dessulfurization->desulfurization

image
image
Explained how distillation actually works

image
Redirect players to the proper quest instead of saying the same things (also wrong numbers for some reason)

image
In this quest I deleted the part "many of them aren't" (about different oil paths being unviable), we should let players figure out these things, finding the optimal path is also a part of the gt experience.

Ideally I would like to delete also parts that suggest best routes e.g.:
image
there are 2 reasons:
1) player should look for ways to create what is needed instead of everything being suggested by questbook
2) as PRs are merged numbers and directions the questbook provides become obsolete, we should avoid adding them so player can look through nei at the version they are playing on to ensure they won't get wrong directions, this lowers the amount of work needed to keep questbook up-to-date


Steelux8 (changes requested):
I ask that you do not delete the parts where I told players which options are best, and which are not good, unless the information is wrong. I deliberately did that so that players could progress with the good powergen choices without having to research them. Anyone can read these quests and still choose whichever path they like.

Pxx500:
image
Also deleted the part that says "ridiculous bio vat nonsense" and the part that says that underground oil is the "real answer"

Pxx500:
> I ask that you do not delete the parts where I told players which options are best, and which are not good, unless the information is wrong. I deliberately did that so that players could progress with the good powergen choices without having to research them. Anyone can read these quests and still choose whichever path they like.

As of now:
-suggestion on how to make ammonia is deleted
-oil prospecting is deleted and instead redirects to the proper LV quest
-part "many of them aren't" suggesting some of the paths are unviable

Last screen I sent is what I would delete, but I didn't

Which one do you want back?

dgealow:
Some suggestions I have, copied over from the discord:

* The LV semifluid generator quest seems to depend on MV fluid extractor--that seems like a mistake, since the primary early-game use is burning creosote, which doesn't require any fluid extractor, much less an MV one.
* "Light sulfuric fuel" quest requires an unwieldy amount of sulfuric light fuel cells--48 to be precise. At that point, just making that many cells is more cumbersome than actually setting up the distillation, and it make the quest very annoying to complete if you think a step ahead and want to pipe all that sulfuric fuel directly into a chemical reactor for desulfication. I'd recommend bumping that down to maybe 32 at most--the other comparable quests only require 8-10.
* "Basic cracking" should depend directly on "desulfurizing" rather than being gated behind 2 (expensive) combustion engines in "light fuel power".
* A full stack of mining pipes ("Hello Baron") is a pretty large steel requirement to get started with oil--perhaps the oil tab should open up just with the "build a distillery" quest, or with a new quest for just getting a couple oil cells.

Steelux8:
> > I ask that you do not delete the parts where I told players which options are best, and which are not good, unless the information is wrong. I deliberately did that so that players could progress with the good powergen choices without having to research them. Anyone can read these quests and still choose whichever path they like.
> 
> As of now: -suggestion on how to make ammonia is deleted image -oil prospecting is deleted and instead redirects to the proper LV quest -part "many of them aren't" suggesting some of the paths are unviable

After reviewing all the changes again, I think only this one should be reverted. I'll try to improve the writing on these quests this weekend, to better clarify options and catch up to changes in setups that didn't exist when I first wrote these quests.

Pxx500:
> > As of now: -suggestion on how to make ammonia is deleted image
> 
> After reviewing all the changes again, I think only this one should be reverted. I'll try to improve the writing on these quests this weekend, to better clarify options and catch up to changes in setups that didn't exist when I first wrote these quests.

Done

Pxx500:
> Some suggestions I have, copied over from the discord:
> 
> * The LV semifluid generator quest seems to depend on MV fluid extractor--that seems like a mistake, since the primary early-game use is burning creosote, which doesn't require any fluid extractor, much less an MV one.
> * "Light sulfuric fuel" quest requires an unwieldy amount of sulfuric light fuel cells--48 to be precise. At that point, just making that many cells is more cumbersome than actually setting up the distillation, and it make the quest very annoying to complete if you think a step ahead and want to pipe all that sulfuric fuel directly into a chemical reactor for desulfication. I'd recommend bumping that down to maybe 32 at most--the other comparable quests only require 8-10.
> * "Basic cracking" should depend directly on "desulfurizing" rather than being gated behind 2 (expensive) combustion engines in "light fuel power".
> * A full stack of mining pipes ("Hello Baron") is a pretty large steel requirement to get started with oil--perhaps the oil tab should open up just with the "build a distillery" quest, or with a new quest for just getting a couple oil cells.

@Steelux8 what do you think about this?

Steelux8:
> @Steelux8 what do you think about this?

Those aren't all quests that I made, so I'm not as sure about what they should be like. I don't have a problem with those changes, though.

Pxx500:
> 
@dgealow 
1) this change would affect balance, I can't change it in this PR since it's only quest PR
2) dream disagrees, it's only 4 operations of chem reactor, while it is quite high it is still a small part of what you'll need anyway
3) done
4) same as 2, also 64+ is needed in pretty much all mining cases except twilight forest

Pxx500:
@Steelux8 ready for review

Steelux8 (approved):

Cake12356:
In the "Exploding Fluids for Power" quest in LV, there's another typo of "Dessulfurize", unsure if I should make a new issue, as this PR fixes similar typos.

this is on 2.6.1

Screenshot 2024-06-19 at 5 14 09 PM

PR #14322, opened 2023-08-24
Improve the Naquadah Fuel quests and add Mk VI quest

S4mpsa:
Makes the numbers match the real values, adds some tips to making the fuels and introduces the Mk VI fuel that was recently added.


chochem (approved):
could use some update of coil quest in the future too but this is all good. thanks!

Issue #14245, opened 2023-08-16
Bee Changes RFC


hacatu:
### Your GTNH Discord Username

hacatu

### Your Pack Version

2.3.4

### Your Proposal

I have some ideas for changes to bees that I'd like to implement:

### Basic Balance Changes:
- alveary needs 3 stimulators instead of 2 to hit the soft cap now, so buffing them a bit would be nice but it's not necessary
- magic, resilient, and basic blood frames should be buffed.  After the additive production modifier change, frames that were 2x are now +1.  This makes them much weaker, which is fine in most cases but these frames are particularly expensive so I think they should be +2 in the case of magic/resilient/base blood
- The Lapotron bee should not be dim locked, the indium bee should not be Mutatron blacklisted, and the Drake Bee should be mutatron blacklisted.  For the indium bee in particular, it's currently not obtainable for no-space players, and this seems particularly egregious because while the mutatron is early LuV and the t5 (?) rocket is late LuV, it's really not that much of a progression skip to allow the mutatron to make indium bees when you include the bacterial sludge -> mutagen line.  If that's an unpopular opinion, it could also remain mutatron blacklisted but have its dim lock removed and instead require a base block of europium or something late LuV+

### New Bee Features:
- I'd like to add a new effect for the barnardac bee that transmutes saplings into barnardac saplings under certain circumstances.  My ideas include only working in the spectre dimension or only working on tree of transformation saplings, but maybe no such restrictions are necessary

### Your Goal

I'd like feedback on these proposed changes, mainly on how/if I should implement the mutation related changes and barnardac bee change

### Your Vision

I will implement these changes after @Runakai1, @kuba6000, @chochem, and if needed Dreammaster et al lmk anything I should change/skip

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

miaowwwwww:
“the indium bee should not be Lapotron blacklisted”， u mean mutatron blacklisted?

miaowwwwww:
indium bees need venus, which is t4, so yes, not much of a skip. it's early LuV

Runakai1:
[...]

PR #14339, opened 2023-08-27
Mutatron blacklist Drake bee, improve gene sample tooltips

hacatu:
See [here](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14245).  This PR makes the Drake bee mutatron blacklisted, so that players must get the awakened draconium coil to breed it as intended.  Even for no space / skyblock players this is still possible via breeding late planet bees that take way more breeding and have a much slower production of awakened draconium.

Also adds tooltips to ALL speed, territory, lifespan, and effect gene samples from gendustry.  This provides valuable in game information about annoying to remember bee data.

Also adds tooltips to the apiary, magic apiary, alveary, and bee house explaining the bee formula.

### DIFF
--- a/config/GTNewHorizons/CustomToolTips.xml
+++ b/config/GTNewHorizons/CustomToolTips.xml
@@ -1,14 +1,23 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <CustomToolTips>
 //Misc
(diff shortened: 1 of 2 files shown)


hacatu:
Example tooltips for gene samples
lifespanEon
speedSlower

hacatu:
The screenshot for Speed: Slower is wrong (I accidentally used an exponent of 0.52 instead of 0.37 when calculating it) but it is fixed in the PR

chochem (approved):
both good changes. thank you

Issue #14318, opened 2023-08-24
The player jumps while pressing space with NEI open IF wearing a QuantumSuit Bodyarmor


chochem:
### Your GTNH Discord Username

chochem

### Your Pack Version

2.3.7

### Your Server

SP

### Java Version

Java 20

### Type of Server

None

### Your Expectation

The player should not jump. This does not happen with other armor/jetpack.

### The Reality

the player jumps. more precisely:

step 1: put on QuantumSuit Bodyarmor and turn on the jetpack
step 2: stand freely and open NEI (so NOT at a AE terminal or with a chest/machine/other gui open)
step 3: press space
result: the player also jumps! (if the searchbar is selected it additionally types a space)

### Your Proposal

fix it!

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [x] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

PR #14389, opened 2023-08-31
Adjust quantum chestplate quest

Pxx500:
image

New text, everything tested in game  (2.4.0RC1)


Pxx500:
resolves 
image
in monthly ticket for quest changes @chochem

Pxx500:
> hmm. so you dropped all mention of speed? I guess that works, its too much of a hot topic for some devs that its hard to even talk about it normally.

I tested them all, they have the same speed

Pxx500:
> However it also has another bug currently that completely drove me mad when I tried it #14318 I went back to nano and still wouldnt recommend this one to anybody despite that hodgepodge buff.

I'd just ignore that, more funny than annoying. Not gamebreaking or anything

chochem:
maybe just me then, I found it very annoying.

chochem (approved):

PR #14509, opened 2023-09-13
Increase Dyson module consumption, especially without computation

S4mpsa:
With the Dyson formula fixed, revisiting the consumption is in order, especially as the old balancing was done with an incorrect script (It had module count and consumption swapped).

"Current" values with existing coefficients and a fixed consumption function, although the consumption has always been at a static 1 module per hour due to a bug in the formula.

image

Increasing computation is fairly inconsequential, and the module consumption still makes the Dyson spammable even without computation.

Module consumption with new values:

image

Adding computation becomes almost mandatory, with each quantum computer (at 25k computation/s) reducing the module consumption by roughly 30% (additive). The computation is capped at 100 000/s to still keep a minimum running cost of 64 modules per hour (One recipe/hour)

This should reduce the viability of spamming dozens of Dysons, add some running cost to properly-supplied Dysons, while still keeping them a competitive option (Naqfuel I am looking at you).


GDCloudstrike (approved):
looks good to me

PR #14532, opened 2023-09-16
Fixing dyson and lategame power gen

meepmopper:
Players will finally be able to overclock their aal and coals.


meepmopper:
This is a bug fix btw

glowredman:
We will not buff the Dyson x1000.

boubou19:
i don't get why people are trying to impose their own view of the pack to everyone else without even taking the balance into consideration...

PR #14835, opened 2023-11-05
Quest changes 2 electric bogaloo

Pxx500:
Moved prospecting quest and changed dependency to a better one
image

Added a mention of small ores to the prospecting quest
image

Reworded EBF quest
image


Pxx500:
Ah I forgot about heavy oil

chochem (approved):
looks good

PR #14841, opened 2023-11-06
another typo :D

Rexicon226:
image
this plasma actually has 150M value.


Dream-Master (approved):

chochem:
you shouldnt change lang files in your quest PRs. that part is automated.

Also: if the fuel value changed, isnt the efficiency number different too?

chochem:
alright, I asked and 150 is indeed the intended value but the efficiency with inf cat is still wrong. its not full then but 53.5% reduced.

PR #14861, opened 2023-11-09
Enable crashOnNullRecipeInput

miozune:
To make spotting recipe error easier. I can load the pack without issue.


bombcar:
I'm fine with this as LONG as we make sure someone at least tries to load the pack once before we ship a .0

chochem:
I think dream always does that for releases, yes.

Dream-Master:
i ship this in new nightly and we test it out

Dream-Master (approved):

PR #14920, opened 2023-11-22
Delete config/RTG directory

Caedis:
RTG is not used. We use RWG.


Dream-Master:
@Caedis let it in. We use it on one official server and I save the standard settings. Sure we can put it in a separate branch but we need to always update it with master

Caedis:
Ahh, didnt know that.

PR #15065, opened 2023-12-11
Rename override directory to give it priority

ghostflyby:
It turns out that filesystem order is different from ascii one.


OneEyeMaker (approved):
I think that 4 underscores is kinda overkill (2 is enough), but anyway...

PR #15151, opened 2023-12-19
Fix small typo (GNTH -> GTNH)

dirkschut:


bombcar:
Should this be fixed in the main quest? Not sure why it’s in txloader

dirkschut:
It is where it is because I'm not familiar with this repo and just looked around and found it here. 😄 What location should I fix it instead?

bombcar (approved):

bombcar:
Yep, that's where it goes!

dirkschut:
> Yep, that's where it goes!

Great. I was just editing in github.dev to just get it done. Didn't want to clone etc. to just fix a simple typo xD

Issue #15322, opened 2024-01-11
Quest wrongly implies rapier affects shielded Twilight Lich


HenryClones:
### Your Pack Version

2.5.0

### Your Server

private server

### Java Version

Java 17

### Type of Server

Vanilla Forge

### Your Expectation

1. Read the "Making a Rapier" quest in the in-game questbook, provided by BetterQuesting.
![image](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/30604707/1e28d72e-22f9-4291-88cb-82d636f82206)
2. Attempt to attack the Lich in the Twilight Forest with the rapier, as the quest implies.
3. Lich takes chip damage until it eventually dies.

### The Reality

1. Read the "Making a Rapier" quest in the in-game questbook.
![image](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/30604707/1e28d72e-22f9-4291-88cb-82d636f82206)
2. Attempt to attack the Lich in the Twilight Forest with the rapier, as the quest implies.
3. As with other weapons, the Lich takes no damage when its shields are active. This mechanically makes sense. However, the "Making a Rapier" quest inaccurately claims that you can use the Rapier to attack the Lich instead of using the projectiles to break its shields. This was possible in older versions of Twilight Forest and Tinker's Construct.

### Your Proposal

Remove all references to the Lich in the quest named "Making a Rapier".

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [X] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

HenryClones:
I do not believe that my proposal is the only way to fix this issue, but I do believe it is a straightforward way to make the quest book not convince players to bring the wrong equipment to a fight.

boubou19:
For years you were able to ignore the shield of the lich by spamming the rapier on the lich, killing it in 15s. So it seems someone changed it.

chochem:
fixed by https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/15323

PR #15323, opened 2024-01-11
Remove references to the Lich fight from the Making a Rapier quest

HenryClones:
One possible fix for issue #15322.


Dream-Master:
The litch take damage but it was reduced and the rapir not ignore the Schild completely like it was before .

HenryClones:
Even if it does, I think that players with rapiers that can defeat the Twilight Lich through its shield reliably will have experimented and discovered that behavior emergently. It also is not as important to mention that boss fight yet, as the quest can be completed before the Twilight Forest is unlocked. I am considering how I will implement your feedback into my PR, so thank you.

chochem (approved):
I think this is fair. rapies is certainly not a magic pill against lich like the old days.

Issue #15405, opened 2024-01-24
Minor Inconvenience: Quest inconsistencies


Savallator:
### Your GTNH Discord Username

Boldar#1676

### Your Pack Version

2.5.1

### Your Server

private

### Java Version

Java 21

### Type of Server

None

### Your Expectation

Since i have not found a better place, I wan't to collect a few minor quest issues i noticed here. I will continue to collect these minor changes needed here, and if anyone else has some, please comment and i can edit it into the list.

### The Reality

- [x] Quest for Crafting Raw Crystal chip (Crystal Chips, LuV Tab) should expect 72 Exquisite Emeralds for the 8 Molten Europium, not 64.
- [x] The same quest should also not talk about the "Low Gravity" since it got removed
- At least for me the Nq skip was not working. It did trigger the hidden quest, but i still needed to cheat the "Almost success" quest in LuV tab, so i can advance further (Because the Nq Alloy quest depends on it). Maybe i missed some other hidden trigger though, can someone confirm this?

### Your Proposal

Fix these in some future release.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [X] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

chochem:
the second point sounds like a recipe bug, not a quest bug though

Savallator:
I am quite sure the low grav requirement was removed deliberately and the quest was forgotten.

chochem:
no, I checked. looks like it was by accident in a refactor.

Savallator:
Then maybe this should be used to discuss the state of them. People used multis to circumvent it anyway. To have that requirement mean anything I would then propose to actually force people to use the low grav environment at least for some time... maybe make the multi recipe one tier higher, but half the time, so in the end it does not change much, but at the start you need to setup a machine in another dim at least for some time? (And exclude pa, or make it have both recipes) Then it would actually mean something, and it could be the right point  to reference a quest for the quantum ring. To not destroy no-rocket-runs maybe give a 10% chance to succeed without kow grav.
Just a few ideas, because right now I think that mechanic is boring, because it does not matter anyways.

[...]

PR #15454, opened 2024-01-31
Small Quest Adjustments/Fixes

### LINKED ISSUE #15430: The process of collecting creosote/limitations of clay buckets should be better explained

tarhalindur

### Your Pack Version

2.5.1

### Your Proposal

In the quest introducing the uses of creosote, explain that it can be collected by either placing a metal bucket in the bottom bucket slot or right-clicking the coke oven while holding a metal bucket. In addition, clarify that you must specifically use a metal bucket - clay buckets will not work.

Possibly also, when clay buckets are introduced, say that there are some liquids that can't be collected without better buckets.

### Your Goal

There's not really a reason a new player would suspect that clay buckets can't hold creosote. When I first put one in the slot and nothing happened, I thought I was using the UI incorrectly and, until I looked up "creosote" on NEI, didn't suspect that the issue was I was using the wrong kind of bucket. Changing these quests will make it easier for new players to know how to get liquid from the coke oven.

### Your Vision

New players shouldn't have to guess at how game mechanics work.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

### LINKED ISSUE #15432: Tell players how to move Pam's Harvestcraft gardens

tarhalindur

### Your Pack Version

2.5.1

### Your Proposal

Add something telling players that it's possible to right-click gardens to move them, and that putting a bunch of them near your base can let you get more.

### Your Goal

I didn't know you could do this until I looked at the wiki, because the QB only says you can destroy them for things. This does need an explanation in-game because I shouldn't have to read the wiki to know how game mechanics work.

### Your Vision

New players would be able to take advantage of this mechanic without needing external guidance.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

Steelux8:
- Change 64 Emeralds to 72 in Crystal Chip quest for correct value;
- Clarify Creosote uses and interactions with Coke Oven;
- Mention right-click to collect Pam Gardens;
- Link to Space Elevator spreadsheet with calculated values for the multi;
- Other assorted tweaks.

Closes #15432. Closes #15430. Fixes the first problem in #15405.


chochem (approved):
lgtm

PR #15550, opened 2024-02-18
Add the Draconic Evolution Information Tablet‘s image offline resources

NealDeal34:
The Draconic Evolution Information Tablet load pictures by accessing appointed website and downloading it to appointed path,but for someone who were offline or in a special network environment that can't visit the website,those iamges could not loaded correctly，so i crate this path and files to load it directly to slove the problem.


Dream-Master:
@NealDeal34 we have the mod tx loader wich can load external stuff better or just create a ressource package. I doubt with this change the mod will be huge.

Dream-Master:
@NealDeal34 did you check the license? Can this pictures be used outside this webpage ?

NealDeal34:
> @NealDeal34 did you check the license? Can this pictures be used outside this webpage ?

@Dream-Master I think you don't need to worry about this at all，it's just a way the mod author to use image as an external resource in manual.The official cited webpages were as follows (eg.)https://raw.githubusercontent.com/brandon3055/Random-Bits/master/DE%20Tablet%20Images/ECore/ECore1.png  in en_US version and (eg.)http://imgur.com/jj2j1nY.png in zh_CN version.Actually what i did is the way also provided by official in https://www.curseforge.com/minecraft/mc-mods/draconic-evolution/files/2234347 to slove the problem.
The mod official license about that is "Art and sound assets are released under the Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International (CC BY-NC-SA 4.0), a summary of which can be found at: https://creativecommons.org/licenses/by-nc-sa/4.0/"，What I said is quite verbose because i know exactly what I'm doing and do hope you can permit it，although this change significantly affected the size of the modpack.Thank you very much.

mitchej123 (changes requested):
License seems fine; however; I'd rather that we either:
1) Offer a link to download and do this
2) Make a separate mod that can be downloaded and makes these available
3) Some other option for the subset of people with this problem to be able to download and have available offline


Increasing the size of this repo with binary objects, and the size of the pack for something that impacts a very small subset of the player base isn't worth it IMHO.  How many people get to draconic in the pack, and how many of those are running offline and want to use the guide book?  I'd wager less than 0.1%

Ethryan:
Maybe make it available as a resource pack instead?

They are adding them on mod/textures/gui/ so it should be possible.

boubou19:
This is a non issue because we already do it for BLS with each major release: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/tree/master/config/txloader/forceload/betterloadingscreen/textures/backgrounds And it takes only a minute to produce a release with the github action. We have time before reaching a point where it becomes unbearable.

For hot mods repos i can understand, but for this one, it's not a big deal.

NealDeal34:
> This is a non issue because we already do it for BLS with each major release: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/tree/master/config/txloader/forceload/betterloadingscreen/textures/backgrounds And it takes only a minute to produce a release with the github action. We have time before reaching a point where it becomes unbearable.
> 
> For hot mods repos i can understand, but for this one, it's not a big deal.

Sorry,but i'm confused that you mean do it or not,for this mod? Thanks for guys advises @mitchej123 because i'm think about it and try to redirect the source url for special regions player who can't download the images normally.But it's not easy for searching a stable image hosting service and need some time.
And，just to clarify，no one's feeling and needs in game should be ignore if the problem could be solve，right？:）

boubou19:
@NealDeal34 After some discussion with @Dream-Master we think it's better to have this as a ressource pack. We will offer it to download somewhere for the people interested in it. Can you change this into a ressource pack please?

PR #15593, opened 2024-02-25
Quest changes 3: the return of the king

Pxx500:
image
image
image
image
 (changed link in space elevator quest so it leads to the first tab of the calculator and not the second one)
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
(2 and 3 are the same as 1)
image
image
image
image
image
image
image
image
image


chochem (commented):
Great work!

only got nitpicks:
- 'new tier MKVI fuel' sounds weird. maybe just 'MKVI fuel', or 'the new MKVI fuel' if by new you mean its a recent addition
- would probably give unenchanted lootbags for the thauminite quest. that is the default so that the enchanting mechanic is meaningful. (also the fusion prereq does nothing, you dont get on the tab without anyway)
- would probably make the text color in the macerator quest the note one, not a warning. red warning is like 'careful or it will explode' and stuff like that.

Pxx500:
image
image
image

Pxx500:
Old:
image
New:
image

chochem (approved):
great work! thank you :)

PR #15617, opened 2024-02-28
Added a quest for the TGS.

AbdielKavash:
Added a quest for the TGS rework here: https://github.com/GTNewHorizons/GTplusplus/pull/839

Questline:

Tasks and rewards:

Full description:
<details><summary>Click to expand</summary>

> This multiblock is the perfect solution to high-speed tree farming. Do you need wood for charcoal or benzene power? Nuts for seed oil? Leaf blocks for decoration? Look no further!
> 
> The Tree Growth Simulator can farm almost any tree in the game, and produces all the items you would get by chopping it down normally. All this in a compact 3x3x3 cube, without any lag from growing a real tree.
> 
> First, place any sapling in the controller slot of the multiblock. Then, insert tools appropriate to the part of the tree you want to harvest in an input bus. Give the machine power, turn it on, and be ready to get flooded by the outputs.
> 
> Different tools are required to gather different parts of the tree:
>  \- Saw for logs
>  \- Branch Cutter for saplings
>  \- Shears for leaves
>  \- Knife for fruit
> Better versions of the tools (Chainsaw, Grafter, ...) will increase the yield! Check the tooltip of the controller block for details.
> 
> You can even use multiple different tools together, and the machine will harvest several different products at the same time!
> 
> Note that the tools used will slowly lose durability, so for long-term operation you might want to set up some automatic replacement.
> 
> [note]While this machine only requires IV, it is quite expensive to make early in the tier. You will need 16 ZPM circuits and some fancy alloys.[/note]

</details>


chochem (approved):
nice!

AdolMors:
Just want to leave a note here about existing behaviour of electric GT tools in TGS not mentioned in tooltip and quest. Electric GT tools consume EU with some chance of durability (much like if they are used normally by hand) if possible. Tools at 0EU are ejected to output bus to allow charging the tools and then re-inserting through input bus. Not sure if the TGS rework here have changed the behaviour or not.

AbdielKavash:
> Just want to leave a note here about existing behaviour of electric GT tools in TGS not mentioned in tooltip and quest. Electric GT tools consume EU with some chance of durability (much like if they are used normally by hand) if possible. Tools at 0EU are ejected to output bus to allow charging the tools and then re-inserting through input bus. Not sure if the TGS rework here have changed the behaviour or not, because after the rework the tools will stay in the input bus. So if the eject mechanic is kept, it should be able to move 0 charge tools from input bus to output bus.

Yes, the reworked TGS still keeps this behavior (I have tested this right now to make sure I have not broken it accidentally meanwhile - is it not working for you?)

I can add it to the quest description too, if you'd like.

AdolMors:
> Yes, the reworked TGS still keeps this behavior (I have tested this right now to make sure I have not broken it accidentally at some point - is it not working for you?)
> 
> I can add it to the quest description too, if you'd like.

That mechanic was just something I did not know when I did test runs of TGS in single player. GT tools with working battery have to be crafted, so using the spawned in tools did not show the behaviour. I learned about the ejection only after seeing it being used with actual tools.

PR #15633, opened 2024-03-02
add angelica cfgs

Dream-Master:
<s>disable  HudCaching to not shuttering with Holo Inventory</s>


mitchej123 (changes requested):
This is not a proper fix

Dream-Master:
sure it isnt but it is a workaround so long this is not fixed. Take this on by default will let your eyes bleeding :trollface:

Issue #15517, opened 2024-02-12
Increase OpenComputer Execution Timeout


Hikari1nVoid:
### Your GTNH Discord Username

Hiakri1nVoid#9527

### Your Pack Version

nightly-310

### Your Proposal

Increase the default Timeout for OC execution to 20s instead of 5s.
As ME networks in GTNH has grown exponentially comparing to vanillas we need extra time for OC to interact with them.

### Your Goal

getAllItemsInNetwork()will fail as it took more than 5 secs.

### Your Vision

Make OC-AE2 interaction more smoothly.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

PR #15668, opened 2024-03-07
Increase execution time limit to 20s

Hikari1nVoid:
Target this:https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15517
20sec should be adequate for general OC usage on large scale ME Networks and not get affected too hard from lnfinite loop apps.
Feel free to edit the number as long as AE2 getItemsInNetwork([filter:table]):table method does not fail in a Stargate base.

### DIFF
--- a/config/OpenComputers.cfg
+++ b/config/OpenComputers.cfg
@@ -269,7 +269,7 @@ opencomputers {
     # programs blocking other computers by locking down the executor threads.
     # Note that changing this won't have any effect on computers that are
     # already running - they'll have to be rebooted for this to take effect.
-    timeout=5
+    timeout=20
   }
 
   # Settings that are intended for debugging issues, not for normal use.


boubou19:
5 seconds is already big, so 20s has the potential of destroying all the computer execution code. Feel free to do that in SP or on your own server, but we can't merge this in the pack because big server owners will then suddently have issues with their player experiencing weird behaviors with OC.

Hikari1nVoid:
> 5 seconds is already big,

This is for vanilla.

> 20s has the potential of destroying all the computer execution code.

This will not affect all OCs that are currently running.Changes will take effect once they are rebooted.And 20s is a "general number"and open for any modification.

And get root access for every GTNH server to modify this code is not that practical.And Server owners will not accept random guys' suggestion to change this config.

Just a random shitposting after I found out I'm the only one believes the execution limit is way too short on discord...this is closed and I probably cannot do anything about this.

PR #15708, opened 2024-03-12
Change Quest Formatting to Visible Tags

Steelux8:
This PR serves to change the most used formatting options in quest descriptions to the tags I implemented in [this PR](https://github.com/GTNewHorizons/BetterQuesting/pull/122). These are the dark red text used for warnings, the blue text used for notes, and the URL formatting which combines blue with underlined.

The purpose of this change is to reduce the amount of invisible formatting in descriptions as possible, given that the current formatting relies on text that is hidden right away and is hard to then change if necessary. On top of that, the new tags can automatically generate different colors for better readability with specific themes.

I will made the changes for those popular colors across the whole questbook, so that the tags are used consistently.


Steelux8:
This PR only changes quest descriptions, to remove invisible formatting and use the new features from the visible formatting I added, so it should be fine to merge. My testing showed no problems so far.

D-Cysteine:
It's probably not intended, but it looks like this PR is deleting some quests:
* https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/15708/files#diff-e72fe81e1aa17d75c23f1ff577d9055de3c97622606d66687ce84cdd4a8c4242
* https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/15708/files#diff-1c74ac5254b0929f37a702f3605b60b7715668607f4e0ef9022d0540a925c215
* https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/15708/files#diff-d2e24500660ae4f2bf89714690a4b94014b6e1bf2d91a5d9f61c34523663ce7b
* https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/15708/files#diff-be7809e25b36ef5a995d4c35481d12113be0f301c8f3ccccf4aafe33221a22a1

(I think that's all of the deletions, but I might have missed some)

Steelux8:
> It's probably not intended, but it looks like this PR is deleting some quests:

I checked, and none of the files that were deleted in this PR exist in master, because they were moved to other tabs or renamed.

Steelux8:
Latest nightly is crashing, I'd rather re-do this PR than trying to fix the problems that came up after format change and months of freeze.

PR #15729, opened 2024-03-15
Thaumcraft quest tab overhaul

chochem:
Draft of a full Thaumcraft quest tab overhaul, authored by AcaMetis. (also thanks to notapengiun for the help in transfering it here)

This is not finished. Besides some checking of contents and progression, known issues are

- [ ] some quests are deleted that maybe shouldnt be. (Or maybe they should be, but then we should be clear about it)
- [ ] some quests were deleted and remade which changes their quest id and thus removes progress from players. This should be fixed. (e.g., the scepter quest. no full list right now)


Dream-Master:
@chochem @NotAPenguin0 any news about this ?

chochem:
will start on redoing it soonish. might wait for after 2.6.1 and that dcysteine pr though

Dream-Master:
> will start on redoing it soonish. might wait for after 2.6.1 and that dcysteine pr though

@D-Cysteine wait for you with the quest format change .

chochem:
yea I just wrote them to go ahead.

PR #15822, opened 2024-03-25
Remove pools and pylons from elven gateway trophy

chochem:
Fixes https://github.com/GTNewHorizons/Angelica/issues/370

This does not need to be the permanent solution. Maybe somebody fixes the issues between amazing trophies+botania+angelica properly at some point, then the pools and pylons can be readded. But for now we dont want such a crash in the stable release.

Tested and doesnt freeze/crash anyway.


config/amazingtrophies/trophies/magic/botania/04_elven_gateway.json
```diff
@@ -25,8 +25,8 @@
       ["       ", "       ", "       ", "  BDB  "],
       ["       ", "       ", "       ", " B   B "],
       ["       ", "       ", "       ", " D   D "],
-      ["C     C", "       ", "       ", " B   B "],
-      ["A     A", "       ", "       ", "  B~B  "]
+      ["       ", "       ", "       ", " B   B "],
+      ["       ", "       ", "       ", "  B~B  "]
```
glowredman, line 29: Could you remove the first three strings in the inner arrays and each outer blank space in the remaining strings?  Like this:
```json
      [" BDB "],
      ["B   B"],
      ["D   D"],
      ["B   B"],
      [" B~B "]
```
Otherwise the structure doesn't appear centered on the pedestal.

glowredman (approved):

PR #16016, opened 2024-04-20
Update InGameInfo.xml - Fix Slime Chunk Designation

PlayfulPiano:
Noticed that the default config doesn't have a functioning slime chunk HUD even though it is included. So I replaced it with one that worked (tested personally) that fit with the "This is an ore chunk." and "This is the center of the chunk." formats.


PlayfulPiano:
> Do you have a screenshot? Looks good from what I can tell though.

Sure, here: 
image
image

(don't mind the moon phase and warp addons that's just from the wiki)

PlayfulPiano:
Actually, maybe it would be better to move this line to be above the center chunk info? Since that doesn't have an image but the ore chunk one does.

NotAPenguin0 (approved):

Alexdoru:
good work

Issue #16126, opened 2024-05-01
distributing stacks by shift-dragging in inventory or crafting grid doesn't distribute evenly


shoobedoo926:
### Your GTNH Discord Username

shoo_be_doo

### Your Pack Version

2.6.0

### Your Server

SP

### Java Version

Java 8

### Type of Server

None

### Your Expectation

distributing two half stacks of coal into four quarter stacks. expected behavior is four stacks of 16.

### The Reality

the items distribute wrong; you shouldn't end up with one stack of 22 and three of 14
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/88210579/5c0a17db-8fd0-46b0-ab3c-97e17bec76c2
I should note that this isn't my video, but I had the same issue in my own save; this video is from another user on the Discord who had this issue

### Your Proposal

actually distribute the items equally (in this case into equal stacks of 16), like in 2.5.1

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [X] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

PR #16145, opened 2024-05-03
Readd mousetweaks config and fix incorrect stack splitting

chochem:
readds the mousetweaks config in the new format with the LMBTweakWithItem setting it had in 2.5.0. I am not gonna touch wheeltweaks here after all that is a bigger can of worms including defaultconfig mod considerations.

specifically:
- this disables LMBTweakWithItem again. which is a buggy thing we always had disabled. This fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16126

### DIFF
--- a/config/MouseTweaks.cfg
+++ b/config/MouseTweaks.cfg
@@ -1,0 +1,35 @@
+# Configuration file
+
+general {
+    # Scroll Scaling
+    # 0 - Multiple Wheel Clicks Move Multiple Items
+    # 1 - Always Move One Item (macOS Compatibility) [range: 0 ~ 1, default: 0]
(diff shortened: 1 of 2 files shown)


Caedis (dismissed):

chochem:
need to think about defaultconfig mod for a second

chochem:
alright, if somebody else wants to change the mod for wheeltweak thats fine but its a bit more involved (see for example https://discord.com/channels/181078474394566657/939305179524792340/1236056489655926885)
this PR is now just focussed on the lmbtweakwithitem

Caedis (approved):

PR #16228, opened 2024-05-10
Update InGameInfo.xml - Weather & Moon Phases, Improving Slime Chunk Commit

PlayfulPiano:
Improves the weather information to be accurate based on localtemperature (which is how weather *actually* works, that being a .15 temperature threshold which is dynamic based on y level) instead of the previous raining/snowing/thundering variables. Also shares the remaining length of the current storm / time until the next storm.

The big series of biomes are me trying to catch every biome which does not have any precipitation, as uh, there is no easy way to add that without including every case (and there are less no precipitation biomes than there are precipitation biomes).

Examples:
image
image
image
image

As for moon phases, this is a modification from what was grabbed in this page here: https://gtnh.miraheze.org/wiki/InGame_Info_XML
Instead of color coding based on Day/Night cycle on the moon, it's color coded based on the fullness of the moon (as the terms for gibbous and crescent might be unfamiliar for users). Instead, the Day/Night cycle is just added to the side separately.

Examples:
image
image
image

Lastly, after realizing there were a uh, few issues with consistency and spacing, I modified the slime chunk line (which I added in my earlier commit here: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/16016) so it's way more in line and consistent with the rest of the HUD. Same goes for ore chunks / center of the chunk.

Examples:
image
image
image

### DIFF
--- a/config/InGameInfoXML/InGameInfo.xml
+++ b/config/InGameInfoXML/InGameInfo.xml
@@ -129,7 +129,33 @@
             <str>{white})</str>
         </line>
         <line>
-            <str>{black}</str>
+            <icon>
+                <str>witchery:mooncharm</str>
+            </icon>
+			<str> Moon phase: </str>
+			<op>
+				<str>eq</str>
+				<modi>
+					<var>day</var>
+					<num>8</num>
+				</modi>
+				<num>0</num>
+				<num>1</num>
+				<num>2</num>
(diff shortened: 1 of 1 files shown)


Dream-Master:
i like the change

Dream-Master (approved):

PR #16177, opened 2024-05-05
Adding new applied greg quests

Hungerya1:
Quests added:
-Crafting input bus
-Crafting input slave
-Crafting input buffer
-Advanced stocking bus
-Stocking Hatches
resim
resim
resim
resim
resim
resim

PR #16382, opened 2024-05-26
Fix Applied Greg quest requisite

Rejyr:
Fixes the Applied Greg quest requisite from #16177.

Applied Greg should only require Interfacing with Machines and not Stocking Fluids. 
Stocking Fluids should require Applied Greg and not the other way around.

Before:
image
image
After:
image
image


chochem (approved):
thanks!

PR #16444, opened 2024-06-02
Botania mushroom coin quest

Reflex18:
Adds a coin quest with a 20 min cooldown and provides the player with their choice of a botania mushroom. It is great for that early game with botania to help the player not have to comb the twilight forest looking for them.


Dream-Master (dismissed):
This quest must be locked after Twilight portal . On my phone rn can’t check .

Reflex18:
> This quest must be locked after Twilight portal . On my phone rn can’t check .

The quest is currently locked behind flower grafting which is apart of a chain of quests that is locked behind the twilight portal.

The player would need to enter the twilight portal, find at least one mushroom and then make the petal apothecary which is gated  behind silver to then access the coin quest.

Dream-Master (approved):

PR #16498, opened 2024-06-09
Update BetterQuesting Readme with New Formatting Tag Information

Steelux8:
Requested by @chochem. The readme now explains how to use the formatting tags and how they set colors on their own.


chochem (approved):
thanks!

PR #16624, opened 2024-06-29
ASoC Quest Relocation

OlliedeLeeuw:
Relocated the ASoc quest in LuV tab:
- Clarification in Precise Assembler quest about requiring ASoC 
- moved it a bit in LuV tab for a better quest flow.


OlliedeLeeuw:
image

chochem (approved):
thanks, thats a good change!

PR #16635, opened 2024-06-30
Trigger Quests inconsistency fix

OlliedeLeeuw:
Epoxid and Styrene-Butadiene work fine, but these 4 quests don't check for Tier Unlock.
- 3 other HV plastics
- Naquadah dust

Naq dust will check for LuV Tier unlock and HV plastics will check for HV chem react (like how Styrene works).


chochem (approved):
ah, thanks for looking through all of them

Issue #15373, opened 2024-01-20
Monthly Ticket for Quest Changes


chochem:
### Your GTNH Discord Username

chochem

### Your Pack Version

2.5.1

### Overview

Well, maybe it is the _quarterly_ ticket for quest changes. A ticket for any large or small problems with any quests. Just post below. Of course, you can also make a dedicated ticket. In particular for more indepth suggestion or problems a dedicated one is better.

The most up-to-date quest files can always be found here: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/tree/master/config/betterquesting/DefaultQuests
Depending on your version, you might also need to update BQ to use them.

Below are the remaining things from the last ticket (https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14789):

### Quest Changes

- [ ] https://github.com/GTNewHorizons/GT5-Unofficial/pull/1146 reworked turbine materials, so quests related to turbine materials need to be checked to confirm if any changes are needed.
- [x] **Quest.3165 UV Power Per Turbine**
  > Large Ichorium 170% 342,720 EU/t (very durable)
  Ichorium turbine has been buffed. So this data is outdated.
- [ ] Thaumcraft quests: the two quest after silverwood core need a core or core options
- [ ] Thaumcraft quests: the last wand quest should point to kami-ha-me-ha tab, the wand quest is there.
- [x] update existing coil quest of Naq Refinery based on the Naq overhaul (e.g. new alternate recipes)
- [x] adjust quest 'two big ovens' to account for and mention this railcraft boiler change https://github.com/GTNewHorizons/Railcraft/pull/17
- [x] add link to a space elevator spreadsheet link or something like that https://discord.com/channels/181078474394566657/1189994332686651682/1195155645532471326
- [x] hydroponics unit quest: Since this machine needs mithril (i think venus is a source, so probably Tier 4?) and naquadah doped wafer, I don't think it's position in the QB is accurate (just after T2 rocket).
- [x] The "Hulk Smash!" Quest in GT++ Tab has a typo, it says: Build a maceration statck ... (which should be stack)
- [x] Quest "Powerful Armor" in the BM tab is asking for 27 Empty and Filled Sockets. This is slightly off, you need 28 sockets to make a full armor set. (4 to make the forge, 5 + 8 + 7 + 4 = 24 for the actual armor.)
- [x] fix canola quest (again), after the name change revert
[...]

PR #16637, opened 2024-06-30
Fixed wrong info in Data Bank Quest

OlliedeLeeuw:
Data Bank quest mentions that LuV energy hatch is enough, but it is not anymore. 
Fixes (https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15373#issuecomment-2041528843)


chochem (approved):
looks good

PR #16684, opened 2024-07-05
Fixed Solar Quest pre-reqs

OlliedeLeeuw:
Solar panel quests had some wrong/outdated requirements, such as wetware mainframe being required for LuV tier.
- Fixed certain tiers not having correct prerequisites.
- Added some missing prerequisites to some tiers.


chochem:
luv solar is already fixed in https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/16657.

OlliedeLeeuw:
> luv solar is already fixed in #16657.

oops i missed that

Issue #7063, opened 2020-12-06
Time to make some steel quest should be streamlined


McFrugal:
#### Which modpack version are you using?
2.0.9.0
#
#### What did you try to do, and what did you expect to happen?
I looked at the quest for the bricked blast furnace and noticed it has a lot of requirements.  Most of them are unnecessary, in fact the clay brick ones directly get in the way since you can get brick dust by macerating bricks harvested from worldgen structures.

#
#### What do you suggest instead/what changes do you propose?
Remove most of the basic material steps.  The only parts that need to stay are the firebricks and the bricked blast furnace block, since those are the end result of your crafting steps.

I saw someone complain about this quest in a different way: that you have to gather all the materials before you start combining them, which is a little silly.  It should be possible, for instance, to start macerating the clay before you gather all of it.  I expect there are more quests like this in the quest book.

Prometheus0000:
Complain to dream

Dream-Master:
No why i would to change this. It explain what eaterials you need. If you go and loot Villages its you way to play the game. Some mp worlds have villages disabled.
The quests are optinal so if you dont like it skip it.

McFrugal:
You do understand that people like to complete quests, right?  It's a bad quest.  You should change it.

alphaNOVAircraft:
The Bricks from a village are a workaround. You get a reward for making the quest the way it is intended. Getting the clay isnt that bad. 
Yes a lot of quests are done that way. 

> The quests are optinal so if you dont like it skip it.
Not all of them, iirc there is still not a way to obtain draconic cores without quests?!

Prometheus0000:
This isn't an optional quest, nor can you find the parts, the problem is that you need a huge number of them and if you start on the next step before you finish the previous, you end up without enough at once to complete it.

The solution would be to require 1 of each and then say in the text how many you need.

0lafe:
Maybe we should stop using the thing that helps people, as a method of rewarding people as well. Really dumb you need to intentionally play worse to receive better rewards

Uristqwerty:
Is it possible to make specific quest steps optional, or automatically complete if a specified later step is completed first? Or even have a checkbox next to specific crafting steps deemed non-critical for "I've already done that", including the ability to de-check it if they later decide they *want* the reminder to craft that bit.

[...]

PR #16692, opened 2024-07-06
Update TimetoGetSomeSte-AAAAAAAAAAAAAAAAAAAATA==.json

Eldrinn-Elantey:


chochem (dismissed):
the bricks look wrong.

but more importantly: did you look who set it all to one and spoke with them? just going back and forth without communication is not good.

Eldrinn-Elantey:
We need 36 firebricks for each firebrick we need 6 firebricks (36*6=216). For 216 "firebricks" you need to make its dust by mixing brick dust and clay dust for each kraft you get 2 dusts (216/2=108). So for 216 firebrick you need 108 clat dust and 108 brick dust.

What do you mean by "just going back and forth"?

chochem:
I said bricks, not firebricks.

back and forth means these used to have amounts but were then set to just 1. (pretty sure anyway)

Eldrinn-Elantey:
I do not know who previously set the value for these quest stages to 1. I think it does not make sense. Accurate values in the quest will make it easier for players to calculate early on. If you think that this is unnecessary, you can close

HoleFish:
you have to make extra bricks for the quest when you accidentally turn a few into firebricks in advance

chochem:
I would assume these were set to 1 when they werent optional? that would make the most sense. but I have not checked, you certainly should. If thats the only reason then it would be fine to change.

of course your numbers are not correct right now anyway.

Eldrinn-Elantey:
I changed the number of bricks to 432 in the last commit

chochem:
yea but not the unfired lol

Dream-Master:
Why not set the amounts to the bricks you need in total ? Since the quest part are optimal it will not harm anyone. Every time I make a bbf I need to make the math how much bricks I use to not waste to much stuff or run out of materials .

Sure the amounts need to be correct .

Eldrinn-Elantey:
The number of bricks has been adjusted

chochem:
so you want me to do the work dream? :(
alright Ill have a look as to where this came from

chochem:
here you here https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/7063

Dream-Master:
what you mean @chochem change the other ingredients as well ?

Eldrinn-Elantey:
So what's the solution?

Dream-Master:
I mean how much clay you need in total etc.

Eldrinn-Elantey:
> I mean how much clay you need in total etc.

A total of 720 clay pieces.

chochem (dismissed):
please dont change numbers to random incorrect values.

Eldrinn-Elantey:
> please dont change numbers to random incorrect values.

Why random? The number is calculated using JEC

chochem:
jec isnt magic. if you dont ask the right question you cant expect the right answer.

Eldrinn-Elantey:
The quest is to make 36 firebricks where 4 will be used to craft the controller and the other 32 for the furnace itself. Where's the mistake?

Dream-Master:
i think it was 1008 or so but more than 720

Eldrinn-Elantey:
Previously, it really was 1008. Initially, I did not touch this value, but changed only those that were equal to 1. Perhaps this value is due to the fact that when crushing with a mortar, 1 small dust comes out and when crushing with a macerator, 2 small clay dust comes out

Eldrinn-Elantey:
If you count through the mortar, you will really need 1008 clay. That's how it was originally, I wasn't going to change these values, I made the changes after this message. https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/16692#issuecomment-2219268226

Eldrinn-Elantey:
Which figure is miscounted?

Eldrinn-Elantey:
> Just change it to 1008

Done

Dream-Master (approved):

PR #16701, opened 2024-07-07
Update CuttingaTreelogb-AAAAAAAAAAAAAAAAAAAC5w==.json

Eldrinn-Elantey:


chochem:
if it doesn't work, then it should be fixed and not just the text changed to 'should'.

Eldrinn-Elantey:
I forgot to state the reason for the change. I think it will be logical if in similar quests will be specified equally this point

Eldrinn-Elantey:
Now I made a big axe out of bronze and the quest counted.

chochem:
I am confused by what you are saying, but I guess that means we can close this?

Dream-Master:
yes close it

Eldrinn-Elantey:
I suggested changing the line that the quest will accept a tool made of any material to the same line that is written in the hammer quest

Issue #16505, opened 2024-06-10
AE Tier 1 Fluid Storage quest has incorrect description


nivanda:
### Your GTNH Discord Username

nivanda

### Your Pack Version

2.6.1

### Your Server

private server

### Java Version

Java 8

### Type of Server

Vanilla Forge

### Your Expectation

The quest said "The main processor fo 1K, 4K, 16K and 64K fluid cells". The recipe needed Engraved Laptron Chips and i thought i needed many of them so i could make a large fluid storage system. So i mass produced them.
![expectation](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/58223153/2c23e671-ce04-4e07-9c7f-a9aaac58f35a)

### The Reality

None of the Fluid Storage Cells and their components need those processors. You only need to make one so you could unlock rest of the quests. So in the end i had wasted a lot of time and resources to make the items that the questbook had said i need a lot.
![reality1](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/58223153/2357ba8d-0ba8-4de7-93e7-50c4cadf073a)
![reality2](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/58223153/18f26ae7-4110-4f26-bb40-ef6381c7a414)
![reality3](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/58223153/16eb70c3-45fc-43b8-a840-8e364413acf5)
![reality4](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/58223153/00e53036-6ad8-4c3f-9164-f0849d002e99)

### Your Proposal

Change the desctiption, so the information is valid.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [ ] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

PR #16702, opened 2024-07-07
Change misinformation regarding AE2FC mod

Baerschen:
This pr aims to fix the following problems:
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15373#issuecomment-1979509272
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15373#issuecomment-1979512989
and fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16505

Fluid processor has updated info since they are not used to craft fluid storage cells. Tier 2 circuits has been removed since its ONLY used to craft 2 blocks from the T.F.F.T and thus should not have its own quest.

I think the change to the requirements of the quest is up to discussion. Below is a screenshot with the 1024k cell as an example. Since this cell is pretty much gated by IV circuits it has the quest for the first IV circ as a requirement. The same has been added to all cells except singularity and quantum. The 1k fluid cell needs no circuit so it just has the AE2 Fluid Storage quest as a requirement.

image


Baerschen:
image

Dream-Master (approved):

PR #16715, opened 2024-07-09
Update redstone quest in stone age

Baerschen:
Cant progress on my new run since you dont get redstone ore anymore. To still make it clear what ore has to be found and mined it will still be in the quest but be optional. This way they can be blicked to see the ore information in NEI.

The second commit updates the quest explaining how to process it since you no longer place the block but rather put it in the crafting table with a hammer 2 times now.

image

This change has to be made for the other ore quests aswell unless having to get amber tools for silk touch is the intended way of finishing these quests, which i highly doubt. Will do them the coming days if this way of changing the quest is accepted


Caedis:
Raw ores are detecting fine for me

Baerschen:
Tested it again and your right i think i cheated in the wrong ore on my test world will close the pr but thanks for checking! sry for the trouble

PR #16716, opened 2024-07-09
Update the redstone processing quest

Baerschen:
Updates the instructions on how to process redstone ore since you will no longer place it in-world as you dont get the block anymore but instead raw ore


Dream-Master (approved):

Dream-Master:
@Baerschen it works also use a hammer to mine redstone in a vein maybe you can add this info as well ?

PR #16572, opened 2024-06-21
Update InGameInfo.xml - Fix Server/Local Discrepency

PlayfulPiano:
Fixes the issue of the broken {raintime} tag for users on multiplayer by giving a non-raintime variant on serverport check for -1.

Tested whether it would detect as the serverport -1 or not on a local singleplayer world (inverting behavior to see if the lines functioned correctly), and I had AnrDaemon help on the discord with finding a better tag for detection. I didn't test it on a personal server file host (as I only play through singleplayer). Might need someone to double check in case connecting to a server hosted on the same machine will still list it as -1 or not. This should work fine though for at minimum external server hostings not on the same pc/ip, so for that it'll be a lot better for users.

Compared to 2.6.1's lines (which would stay persistent for singleplayer users), new lines are as follows:
It is {darkaqua}Raining{reset}. (While Raining) 
It is {darkaqua}Raining{reset} and {gold}Thundering{reset}. (While Raining and Thundering) 
It is {darkaqua}Snowing{reset}. (While Snowing)
It is {darkaqua}Snowing{reset} and {gold}Thundering{reset}. (While Snowing and Thundering)
There is currently no {aqua}precipitation{reset}. (When in a biome that has precipitation but no precipitation is occurring)
This {darkgreen}biome{reset} does not have {aqua}precipitation{reset}. (When in a biome that does not have precipitation, unchanged)

If you do want to test it yourself, copy lines 160 through 913 and paste that in your own IGI config (or just copy the file itself and replace it).

PR #16727, opened 2024-07-10
Modified IGI config to fall back to less informative rain changes

Kortako:
…from https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/16572 if IGI is not installed on the server, not based on a server port check.

This should be paired with https://github.com/GTNewHorizons/InGame-Info-XML/pull/21, and IGI should be included in the server files zip after that PR is merged.  If IGI is not included in the server though, it is still fine, as it will just fall back to the changes @PlayfulPiano made in the previous PR.

### DIFF
--- a/config/InGameInfoXML/InGameInfo.xml
+++ b/config/InGameInfoXML/InGameInfo.xml
@@ -159,10 +159,12 @@
         </line>
         <line>
             <if>
+                <not>
                 <equal>
-                    <var>serverport</var>
-                    <str>-1</str>
+                    <var>nextrain</var>
+                    <str>?</str>
                 </equal>
+                </not>
                 <if>
                     <or>
                         <equal>
@@ -216,12 +218,10 @@
                 </if>
             </if>
             <if>
-                <not>
                 <equal>
-                    <var>serverport</var>
-                    <str>-1</str>
+                    <var>nextrain</var>
+                    <str>?</str>
                 </equal>
-                </not>
                 <if>
                     <or>
                         <equal>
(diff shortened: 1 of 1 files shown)


chochem:
I would say we remove it. (and moon and slime)
permanent screen space is precious and this is clear bloat

PlayfulPiano:
That could be an argument for the entirety of igixml. Rather prefer more useful UI additions show up that ppl can simply disable if they don't want it.

AnrDaemon:
Moon phases are important at certain points in progression for some of the mods.

A real good solution would be a management addon that enable/disable certain features, write new config and call reload.

The idea is: Have a number of XML blocks saved in the `~/config/InGameInfoXML/blocks/`. Mod show a list of these blocks and against each, a combo selector for its location (disabled/top/left/center/etc.). Include simple logic to prevent two blocks taking same place. Click `Confirm` to reload IGI.

Dream-Master:
@Kortako

Kortako:
Honestly, that kind of modification you're talking about is a little out of my league.  This PR along with the linked one I made for IGI was solely intended to fix data inconsistencies between servers instances and SP instances with the PR that was merged in the 2.6.1 update [here](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/16228), which is the one that added the moon phases and rain info, and this accomplishes that goal.  I would imagine that any further additions or removals to the actual info displayed in IGI would probably be better served in its own PR.

Dream-Master (approved):

PR #16752, opened 2024-07-14
Remove old destructopack quest in MV tab

Baerschen:
Since ServerUtilities is now shipped with the pack by default and it offers a by default turned on trash can in the inventory having a quest for a item that costs 4 HV circ for a mobile trashcan seems pointless. Since the MV tab has a lot of quest already i think removing pointless quest like this is a good idea.

image


Dream-Master (changes requested):
let the people decide which trash can they use. I would not remove it.

Baerschen:
New descriptions is as follows and mentions both alternatives (in case someone did not find or use the SU trashcan):
image

Dream-Master (approved):

PR #16781, opened 2024-07-17
Add quest for Eternity, edit Applied Greg for the output cache changes, and make the EOH quest a main quest

srdr2k3:
image
^^ My math was wrong, the chronic singularity amount has been updated to 5
image
image


Dream-Master (approved):

Pxx500:
You didn't mention that the ME output bus/hatch can store more than it's capacity as long as it's not full, but won't when it's above limit. (https://discord.com/channels/181078474394566657/295669878222880769/1261323017456324730)

srdr2k3:
> You didn't mention that the ME output bus/hatch can store more than it's capacity as long as it's not full, but won't when it's above limit. (https://discord.com/channels/181078474394566657/295669878222880769/1261323017456324730)

I assumed that would be a known thing, but I can make a separate PR for it

PR #16796, opened 2024-07-19
Multiblock Goals small revamp

OlliedeLeeuw:
Moved some quests around in the tab, so most multiblock are grouped by tier and not randomness
image
Added two new quests
- Clean Implosions?!: Adds a much needed quest for Electric Implosion Compressor.
image
- Draconic Development: Adds a quest for casings for the Draconic Evolution Fusion Crafter.
image

Old look of Multiblock Goals:
image


Pxx500 (approved):
looks good

Issue #16814, opened 2024-07-21
Suggest removing servers that break laws from the default server list


CatsLeftEar:
### Your GTNH Discord Username

Keyla3300

### Your Pack Version

2.6.1

### Your Proposal

Remove MineYourMind server from the default servers list (as well as several other servers i have mentioned in comments). This server contains heavy pay2win, such as: Tiers, which give fly and god mode. You can buy credits, with which you can buy items in item shop and claim chunks, so, basically, a direct real money to in-game currency transaction.
They also sell base restoration due to inactivity, i suppose they are so low that they remove your base for being inactive.

### Your Goal

Goal is cut support or a flow of new players to bad servers with predatory tactics.

### Your Vision

My vision is to follow EULA agreement and not having P2W servers with kits, that not only can cost 80 dollars, but also have the text "We have the right to revoke them." 

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

CatsLeftEar:
Second server that breaks rules/laws is APOC Gaming GTNH, which sells crate keys for real money. Gambling for EV items in GTNH has never been easier!
![image](https://github.com/user-attachments/assets/90a11c79-53c9-49c1-9ae0-31060b0c6cd9)
And with the acccess to a $274.99 dollar pack you can get a veinminer unlocked for you.

CatsLeftEar:
MyFTB server offeres slightly cheaper and less predatory rewards, for 10 euro you can get a kit with the loot depending on your tier. For example, screenshot:
![image](https://github.com/user-attachments/assets/6da6424d-bd20-4041-8d40-45be7a5e1d39)

RealSilverMoon:
I believe that commercial server operation of this pack is strictly prohibited by the EULA. They may require not only removal, but further investigation and condemnation.
Only non-profit donations are compliant.

Einhornyordle:
> I believe that commercial server operation of this pack is strictly prohibited by the EULA. They may require not only removal, but further investigation and condemnation. Only non-profit donations are compliant.

[...]

PR #16821, opened 2024-07-22
Delete servers breaking eula

Pxx500:
closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16814


Dream-Master (changes requested):
Before we are doing this I would like to contact these server owner and ask about this.

Ethryan:
While removing these, we should probably go through and take a quick look so there isn’t any more examples of this kind of monetization used in the other servers.

Dream-Master (approved):

Einhornyordle:
> Before we are doing this I would like to contact these server owner and ask about this.

May I ask who of us (MyFTB) you contacted?

PR #16850, opened 2024-07-27
Adds a quest for the new water pump

Baerschen:
This adds a quest for the new waterpump. See images below for more details. Since its a quest for early game it is a crafting task to keep in line with other early game quests (there is no assembler recipe anyway).

image
image


Dream-Master (approved):

chochem:
thats a strange name did you confuse it with the other pump?

steam consumption might be bugged as well, iirc the agreed value was 1500

Dream-Master:
@Baerschen yes the value was changed after first set. Did you test it with old or with the new version ?

Issue #16658, opened 2024-07-01
Monthly Ticket for Quest Changes


chochem:
### Your GTNH Discord Username

chochem

### Your Pack Version

2.6.1

### Overview

The monthly ticket for quest changes. Well, maybe it is the _quarterly_ ticket for quest changes. A ticket for any large or small problems with any quests. Just post below. Of course, you can also make a dedicated ticket. In particular for more indepth suggestion or problems a dedicated one is better. This ticket is more focussed on small and clear problems and suggestions that don't really need any larger discussion.

The most up-to-date quest files can always be found here: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/tree/master/config/betterquesting/DefaultQuests
Depending on your version, you might also need to update BQ to use them.

Below are the remaining things from the last ticket (https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15373):

### Quest Changes
- [ ] https://github.com/GTNewHorizons/GT5-Unofficial/pull/1146 reworked turbine materials, so quests related to turbine materials need to be checked to confirm if any changes are needed.
- [x] quest "Smaller Fusion Is Bigger" in "Powerful Nuclear Physics" should mention that compacts follow different overclocking rules, needing double the startup power and double the voltage instead of just the voltage.
- [x] some potential issues with AE2FC quests https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15373#issuecomment-1979509272, https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15373#issuecomment-1979512989
- [x] add T9 planets to some quests https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15373#issuecomment-2094254113
- [x] remove samarium oxide and mention of gangue from IV lanth quests
- [x] remove cynical comment at the end of arc furnace multi quest

### Wishlist for New Quests

- [x] quest regarding NEI's 2.6.0+ recipe group & linking mechanics?
- [x] quest regarding NEI's 2.7.0+ new search syntax options?
- [x] quests for new steam multis
- [ ] add quests for bastline (maybe not full detail but at least something?)
- [x] add Eye of Harmony quest + related stuff
- [ ] Add quest line or some tips on how to do Botania potions, specifically for Warp award flasks/alfglass/incense stick.
- [x] add quest for dark steel chest or even an entire ironchest progression.
[...]

PR #16910, opened 2024-08-05
Remove questbook reload quest since its outdated

Baerschen:
fixes (https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16658#issuecomment-2267916633)
Hope putting it in brackets does not close the monthly ticket again.

The removed quest is outdated since the questbook auto updates on loading worlds since some time.


Dream-Master:
sure you not to reload the quest but mentioned the command blocks would be nice. There are others as well.

PR #16949, opened 2024-08-09
Update reload quest instead of removing it

Baerschen:
Updates the quest reload quest to not spread misinformation. This is a follow up to https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/16910#issue-2448175764 that does not remove it as requested in discord dms.
image


Dream-Master:
Btw this command block was added to use the command if you not be an admin .

Dream-Master (approved):

PR #16955, opened 2024-08-09
Fix better questing bees

vehsamrak:
Fix in Bee Breeding section of better questing texts.
There seems to be a typo, because name of the bee type is "Forest".

Снимок экрана от 2024-08-09 19-12-43


chochem (dismissed):

chochem (changes requested):
you should only make changes in the default quests folder.

vehsamrak:
> you should only make changes in the default quests folder.

Done. Reverted fixes in `.lang` files.
Those `config/txloader/load/betterquesting/lang/template.lang` and `config/txloader/load/minecraft/lang/ru_RU.lang` would be generated from default quests json on building stage, right? @chochem

chochem (approved):
the translation file is automatic yea

PR #16999, opened 2024-08-14
CollapsibleItems: More Groups

slprime:
### New Groups:
image

+ All Tinkers Tools what was shown after https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/16907

### If INpureProjects Mod are disabled
image

### Fix:
image

### DIFF
--- a/config/NEI/collapsibleitems.cfg
+++ b/config/NEI/collapsibleitems.cfg
@@ -367,12 +367,12 @@ gregtoolparttoughbind | tconstruct:toughbinding
 gregtoolparttoolrod | tconstruct:toolrod
 gregtoolparttoughrod | tconstruct:toughrod
 gregtoolpartlargeplate
-gregtoolpartlargeguard
+gregtoolpartlargeguard | tconstruct:wideguard
 gregtoolpartswordblade | tconstruct:swordblade
-gregtoolpartmediumguard
+gregtoolpartmediumguard | tconstruct:handguard
 gregtoolpartcrossbar | tconstruct:crossbar
 gregtoolpartknifeblade | tconstruct:knifeblade
-gregtoolpartfullguard
+gregtoolpartfullguard | TConstruct:fullGuard
 gregtoolpartfrypanhead
 gregtoolpartsignhead | tconstruct:signhead
 gregtoolpartchiselhead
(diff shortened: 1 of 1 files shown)


config/NEI/collapsibleitems.cfg
```diff
@@ -223,6 +241,14 @@ $pipeHexadecuple
 :gt.metatool.01 180-192
 #gregtech coffe
 :gt.metaitem.02 32000-32009
+#Exquisite
+:gt.metaitem.02 30208,30209,30500-30505,30507-30514,30527,30528,30890 | dreamcraft:item. GemExquisite
```
chochem, line 245: wouldnt it be more helpful to group by material instead of by oreprefix?

slprime: I don't think so. Groups are more useful if used to group by type rather than by material.

slprime: P.S. We group the same Storage Drawers by type and not by the material they are made of

chochem: because they are interchangeable and practically the same item. nothing could be further from the truth for different dusts/plates/etc

slprime: Gems are no different from plates, dust, gears, wires, pipes, etc.

slprime: When I started creating groups I tried to follow simple grouping rules:

1. objects of the same type but different colors (example: glass)
2. items of the same tier but from different materials (example: drawers,gears,wires)
3. association by topic. not advisable. rarely used (example: flower)

S4mpsa: If I need Tungstensteel Double Plates, I would like to search for "Tungstensteel" and expand the "Tungstensteel" group instead of having to search for "Double Plates", expand the "Double Plate" group and *additonally* search for "Tungstensteel Double Plate" in that group.

slprime: this is what you get if you search "Tungstensteel"
image

You won't have to expand anything. But if you make groups by material, you'll have to expand the material every time

Caedis (approved):

Caedis:
@S4mpsa @chochem everything good to merge?

PR #17164, opened 2024-08-31
Fix 'Dig Deeper for Water" quest

Pxx500:
Fixes quest requirement to not use a deleted quest


chochem (approved):
gonna trust you on that

PR #17177, opened 2024-09-02
Add storage drawers pack configuration

Cleptomania:
This adds configuration that is being added with https://github.com/GTNewHorizons/StorageDrawers/pull/33. This controls which add-on packs are enabled in the pack.

I've disabled the Erebus add-on as that hasn't been shipped with GTNH previously.

### DIFF
--- a/config/StorageDrawers.cfg
+++ b/config/StorageDrawers.cfg
@@ -109,11 +109,13 @@ integration {
 
 
 packs {
-    B:enableBiomesOPlenty=true
+    # If enabled, all explicit pack options will be ignored. Packs will be enabled if their corresponding mod is present (Not including misc pack)
+    B:autoEnablePacks=true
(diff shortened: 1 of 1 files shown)


Caedis:
Could an AUTO option be useful? On if detected, off if not

Cleptomania:
That could definitely be done, I'll something to that for the storage drawers PR and update this accordingly

Cleptomania:
@Caedis I added a config option called `autoEnablePacks`. Which makes it ignore the other settings, and load if the target mod is present or not.

However this doesn't apply to the misc pack, that must be explicitly enabled/disabled(defaulted to false in StorageDrawers) as it targets like 6 different mods. It could probably be broken up in the future though.

Caedis:
That works

Caedis (approved):

PR #17191, opened 2024-09-02
Quests for Artificial Cells & unification of item/fluid/essentia AE2 quests

Baerschen:
This PR changes placement/requirement/rewards of some late game AE2 quests.
First off with all changes the tab as a whole looks like this (changes are highlighted with my amazing drawing skills):
image
The most important changes are in the blue block. Changes that have been made:

1. Adds 2 new quests for the new Artificial Universe Cell & its fluid version. These are gated behind singularity and UXV circuit quest as they are needed for crafting them
image
image
3. Scale down icon size of the quest because of the amount of endgame singularity type cells now IMO they look out of place being double the size of all other quests
4. Change the position of the Singularity Crafting from top to bottom to be inline with the other crafting storage quests
5. Add requirements that are in place for the digital singularity cell to the fluid & essentia variant. Instead of the UEV circuit quest which is needed for quantum tanks/chests 5 the essentia one gets the quest for the essentia reservoir since it is used in crafting
image

The other part of this PR deals with the thaumometric essentia cell. As it turns out we have got two of those that are both different in prereqs and in information given. So instead one was deleted, some text taken from it and merged with the other quest which is now a multiple questline quest as it sits in both UIV and AE2 tabs.
image
image

Thanks for coming to my TED Talk, as always if there are any changes that need to be done comment on github or ping me on discord.


chochem (approved):
lgtm

PR #17232, opened 2024-09-06
More tgregworks removals

chochem:
cleanup on useless and terrible tinker materials.


Dream-Master (approved):
ok from my side

boubou19:
what happens if you already have a tool with this material before config update? does it crash?

chochem:
one second, people are giving me some feedback

chochem:
alright, ready now. combined with the previous pr that is 88 disabled materials.

Baerschen:
> what happens if you already have a tool with this material before config update? does it crash?

Tested it with one of the removed materials and it does not crash. This was the result:
image

combusterf:
Some items that are removed as useless but might not have been intended to be useless:
- `Silicone`, `StyreneButadieneRubber`: Synthetic rubber makes one of the few soft mallets we have in game, and its the best non-magical material for it. Are these affected by this change?
- `Livingwood` and/or `Dreamwood`, would love to see those balanced as better tool handles instead of being unceremoniously removed and give people more reasons to divert into magic.

Baerschen:
>     * `Silicone`, `StyreneButadieneRubber`: Synthetic rubber makes one of the few soft mallets we have in game, and its the best non-magical material for it. Are these affected by this change?
They are not affected

combusterf:
I've added a ticket and plan to pick it up: #17235

PR #17236, opened 2024-09-06
Gregworks round 3

Baerschen:
17 more materials bite the dust. Most of them are unobtainable relics like Angmallen or Atlarus but this also disables some late game mats since they are not needed. Namely Spacetime, Universum, Magnetohydrodynamically Constrained Star Matter and Transcendent Metal


kstvr32:
Why are you disabling obtainable late game materials? I don’t see a problem with spacetime or magmatter tools.

Baerschen:
> Why are you disabling obtainable late game materials? I don’t see a problem with spacetime or magmatter tools.
> 
Because they are not really needed. The only thing changing with those mats is durability which does not matter at that point, they clutter NEI, they need to be loaded in which takes time, their texture is off since gregwork cant handle the custom textures for the most part. Also asked late game devs and they agreed as can be seen https://discord.com/channels/181078474394566657/401118216228831252/1281655026569318416
> Does it break saves with existing tools of materials you are disabling?
> 
It does not as was tested for 2 other PRs that removed 88 materials. If you happen to have a tool made of an disabled material you will get this https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/17232#issuecomment-2334194106

boubou19 (approved):

PR #17241, opened 2024-09-06
Re-enable mutatron for indium bee

StaffiX:
After https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/967, the indium bee is not as overpowered as it was before, so the mutatron blacklist is not necessary anymore.


Dream-Master:
why not keep requirement ? before it was disabled @StaffiX

StaffiX:
My idea was about having 3 ways of getting indium bee
Mutatron needs naquadah bee (so a decent amount of bee progression is required) and LuV components.
Venus needs T4 rocket so smaller assembly line than mutatron and less bee breeding
Lots of Thaumcraft progression to get it in IV

Dream-Master (approved):

PR #17729, opened 2024-10-23
Update InGameInfo.xml - Adding center chunk icon & removing unecessaey chunk checks

PlayfulPiano:
Noticed that the current .xml file has an unnecessary chunk coordinate check (chunk coordinates, not relative coordinates within a chunk) for the "center of the chunk" line. I removed that part and added in the obsidian tough binding icon to be placed next to this line.

I would've added some sort of red tough binding icon but I could not figure out the proper file path for like an ignis tough binding. So obsidian it is.

Also uhhh ignore that second commit. I tried to fix one of the lines using a different tab size but it literally did nothing and could still be saved as a commit so /shrug

### DIFF
--- a/config/InGameInfoXML/InGameInfo.xml
+++ b/config/InGameInfoXML/InGameInfo.xml
@@ -1039,10 +1039,10 @@
                     </equal>
                 </and>
                 <concat>
-					<icon>
-                        <str>tinker:textures/items/parts/obsidian_toughbind.png</str>
-                    </icon>
-                    <str> You are at the {darkred}center {white}of the chunk.</str>
+                        <icon>
+                                <str>tinker:textures/items/parts/obsidian_toughbind.png</str>
+                        </icon>
+                <str> You are at the {darkred}center {white}of the chunk.</str>
                 </concat>
             </if>
         </line>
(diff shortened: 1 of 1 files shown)


PlayfulPiano:
trying to put this into draft, now realizing that the PR https://github.com/GTNewHorizons/InGame-Info-XML/pull/24 seems to have caused a lot of issues with icons displaying. can't get this pushed until that first is fixed. (this code worked fine in beta 2 but beta 3 seems to have multiple icons broken)

PlayfulPiano:
This can be undrafted as of the latest igixml version. Technically a day after deadline but the PR was made before so /shrug

config/InGameInfoXML/InGameInfo.xml
```diff
@@ -1039,44 +1039,10 @@
...
-                                        <num>3</num>
-                                    </modi>
-                                    <num>1</num>
-                                </equal>
-                            </or>
-                        </and>
-                        <str>      </str>
-                    </if>
-                    <str>   You are at the {darkred}center {white}of the chunk.</str>
+					<icon>
```
wlhlm, line 1042: Please use spaces instead of tabs for indendation.

wlhlm (changes requested):
Looks good other than the minor formatting nit.

config/InGameInfoXML/InGameInfo.xml
PlayfulPiano: done, new commit has the changes to indentation

wlhlm (approved):

PR #17736, opened 2024-10-23
Enable a few SU options

Lyfts:
This aims to lessen the amount of things players need to turn on in config if they want to use SU features.

The newly enabled values are

/ranks - Does nothing if ranks are turned off (which they are by default)
/chunks - Does nothing if chunk claiming is turned off (it is by default)
/dump_chunkloaders - Useful for debugging and is an admin command anyways
/killall - Admin command
/leaderboard - Why was this turned off?
/nbtedit - very useful for debugging and is still only usable by OPs
/tpl - admin command
/inv - admin command
and finally it enables chunk loading, this does nothing if claims are turned off as it says in the comment above the setting but leaves one less option that the players need to manually enable.

Most of the useful stuff that people use is still disabled but hopefully this will help alleviate some confusion, which is most often caused by /ranks or /chunks being disabled after they have already enabled chunk claiming/ranks.

### DIFF
--- a/serverutilities/serverutilities.cfg
+++ b/serverutilities/serverutilities.cfg
@@ -85,10 +85,10 @@ commands {
     B:backup=true
 
     #  [default: true]
-    B:chunks=false
+    B:chunks=true
 
     #  [default: true]
-    B:dump_chunkloaders=false
+    B:dump_chunkloaders=true
 
     #  [default: true]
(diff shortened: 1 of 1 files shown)


Ethryan (approved):

Alexdoru:
thanks

PR #17844, opened 2024-11-03
Update questbook page for Industrial Precision Lathe

JMan-Zx:
The item pipe casing parallel/speed factor is not very intuitive, and doesn't really fit well in a tool-tip. I added some quick explanations.

Planning to update the controller tooltip afterwards due to redundant info, and also because of missing paren in the formula for speed boost in the tooltip.


JMan-Zx:
Idk, I was possessed by something and imagined the wrong letters.

Dream-Master (approved):

PR #18064, opened 2024-11-22
Fix for tengam meteor to make it match old numbers

koolkrafter5:
The expected yield for this config and the old one are both 26 tengam ore (with the updated Blood Magic version, the previous version significantly underestimated the ore yield for all meteors so the numbers from beta 4 and before were shown to be a bit lower).

### DIFF
--- a/config/BloodMagic/meteors/CheatyVeryLowQuantityRawTengam.json
+++ b/config/BloodMagic/meteors/CheatyVeryLowQuantityRawTengam.json
@@ -1,14 +1,15 @@
 {
   "ores": [
-    "OREDICT:oreTengamRaw:4"
+    "OREDICT:oreMarbleSalt:69"
   ],
   "filler": [
-    "OREDICT:oreMarbleSalt:3000"
+    "OREDICT:oreMarbleSalt:72",
+    "OREDICT:oreTengamRaw:1"
   ],
   "radius": 16,
   "cost": 1000000001,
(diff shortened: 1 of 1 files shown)


AbdielKavash (approved):
A bit more detail for those who missed the discord discussion:

Previously tengam was a part of the meteor with an approximately 1/1000 chance (exactly 4/3004). Converting all salt blocks to filler in https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/18062 accidentally made the tengam chance way too high, as the maximum possible ratio for filler is 99. Thus tengam had a 1% chance of appearing.

Further, since adding Terrae or Orbis Terrae to the ritual increases filler chance, this would mean that the chance goes to 100 and *no* tengam would be generated. This is both unintuitive and undesirable.

We decided to instead make tengam part of the *filler* blocks -- it does not have to make sense as this is a "meme" meteor anyway; and it also means that actively engaging with the reagent mechanic gives you a little more tengam on average.

Comparison of expected average tengam yields:

| | Old version (2.6.x) | New version (after this change) |
| --- | --- | --- |
| No reagents | 25 | 26 |
| Terrae | 30 | 31 |
| Orbis Terrae | 35 | 40 |

Note that the old values don't exactly match NEI, as the 2.6.x calculation for that underestimated the radius of the meteor. This has been already fixed in https://github.com/GTNewHorizons/BloodMagic/pull/60.

PR #18084, opened 2024-11-23
Fix mismatching config values and indicated default value in server utilities

Worive:
The server utilities's config have several properties which their actual default value is not the one indicated by the comment above them within `serverutilities/serverutilities.cfg`.

This PR fix that changing the indicated default value to match the actual value,

### DIFF
--- a/serverutilities/serverutilities.cfg
+++ b/serverutilities/serverutilities.cfg
@@ -78,7 +78,7 @@ chat {
 
 
 commands {
-    #  [default: true]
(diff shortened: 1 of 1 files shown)


Worive:
> Those are all autogenerated from the mod defaults and will likely change back to their previous value once the game is launched.

Oh, I mixed up the default values of the mod with the default values of the modpack, make sense then, I'll then just close the PR.

Thanks!

PR #18100, opened 2024-11-24
tooltips for preconfigured storage buses

boubou19:
it's not ideal, but at least it works for base configurations.
Requires https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1093


serenibyss (approved):
Not ideal but gets the job done for now for the recipes

PR #18156, opened 2024-11-28
Update for LootGames and TooMuchLoot Chest3

EagleEyeLogic:
Actual config file for LootGames is a very old one and correspond not with the new created config, so delete this old file.

Also provide an edited rewards.cfg because TooMuchLoot provide only the content for the chests 1-4 but do not overwrite the rolls/picks per chest correctly.

With the default rewards.cfg we get

4-8 rolls/picks for chest 1 (wanted only 2-4)
5-10 rolls/picks for chest 2 (wanted only 3-6)
6-11 rolls/picks for chest 3 (wanted only 4-8)
7-12 rolls/picks for chest 4 (wanted only 5-10)

Update chest 3 as Forestry Farm Blocks are

:0 for Farm Block
:2 for Gear Box
:3 for Hatch
:4 for Valve
:5 for Controller

and goal should be to have a chance for an early MultiFarm. So most FarmBlocks, than hatch, valve, gear and lowest amount for controller.


EagleEyeLogic:
Only to be clear for that, you say from chest1.xml to chest4.xml how much picks you like to have at min and max.

Dream-Master (approved):

PR #18171, opened 2024-11-30
EBF Coil quest description: Unify kelvin symbol to K

purchan:
In EBF Coil quests, the symbol of temperature unit kelvin is not unified (`K`, `k` or no symbol at all).
Every symbol of kelvin will be unified to `K` after this PR.


wlhlm (approved):
Nice! I hate inconsistencies.

PR #18179, opened 2024-12-01
removed wrong information about hile from quest

Nockyx:
Follow-up to https://github.com/GTNewHorizons/GT5-Unofficial/pull/3584


FourIsTheNumber (approved):
Thanks

PR #18386, opened 2024-12-14
add MY FTB server back to list

Dream-Master:
Selling kits and everyone on the server get the benefit are not against Eula so far ik.

### DIFF
--- a/config/defaultserverlist.json
+++ b/config/defaultserverlist.json
@@ -9,7 +9,7 @@
     "Test Server GTNH 4 EU": "zeta.gtnewhorizons.com",
     "Prospercraft GTNH Server US": "newhorizonsold.prospercraft.com",
     "StoneLegion.com GTNH (Canada)": "mc.stonelegion.com",
-	"MyFTB GTNH German Server": "newhorizons.myftb.de",
+	"MyFTB GTNH German Server": "gtnewhorizons.myftb.de",
     "StoneHuts GTNH EU whitelisted": "gtnh.stonehuts.net",
     "Zvezdolet GTNH Server 1": "gtnh.114-7.com",
     "Zvezdolet GTNH Server 2": "gtnh2.114-7.com",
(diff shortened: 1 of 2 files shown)


config/defaultserverlist.json
```diff
@@ -9,6 +9,7 @@
     "Test Server GTNH 4 EU": "zeta.gtnewhorizons.com",
     "Prospercraft GTNH Server US": "newhorizonsold.prospercraft.com",
     "StoneLegion.com GTNH (Canada)": "mc.stonelegion.com",
+	"MyFTB GTNH German Server": "newhorizons.myftb.de",
```
wlhlm, line 12: Just to clarify: this domain is different to [the one we remove](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/16821/files#diff-d09fbafc9d969165d4e2718faf79e02b17c6bd63db392f30b749a072295be28dL9). Is that correct?

chochem:
this post is not correct? https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16814#issuecomment-2241762175

config/defaultserverlist.json
Einhornyordle: gtnewhorizons.myftb.de is the right one

Einhornyordle: yes, just as it was before the removal. It's also on the website if  you open up the "Serveradressen" part https://myftb.de/servers/gt-new-horizons

chochem:
> this post is not correct? [#16814 (comment)](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/16814#issuecomment-2241762175)

ok yea apparently it was incorrect or misleading at best
image
jeder spieler = every player on the server gets that

config/defaultserverlist.json
Dream-Master: Ok will correct later

chochem (approved):

PR #18431, opened 2024-12-18
Update Coolant-AAAAAAAAAAAAAAAAAAAMAQ==.json

Willshaper:
Fixed Quest text.

With Distilled Water you need 8 Lapis dust, and with normal water you need 64. This fixes the quest text to reflect that properly.


chochem (approved):
the diagnosis is not correct, but the solution sure is.

PR #18492, opened 2024-12-23
Removed unnecessary NBT tag from Quantum Star quest radium reward

C-Remilian:
Radium obtained from Quantum Star quest has an additional NBT tag that makes it different from radium you would obtain from processing (eg sifting uranium ores), which makes it not stackable in barrel-like containers.

Normally it wouldn't matter, as the player who claimed the reward would get it into their inventory where the item would start ticking and immediately become non-stackable via new NBT data. But I though I was smart and claimed it with full inventory over an item collector, only to realize that it doesn't stack anyways due to reward already having NBT data in it. Repeating this trick with edited quest reward made the reward barrel-stackable.


Dream-Master:
Are you sure that the ticking starts at 0 without an nbt tag and does not immediately change to the next state?

C-Remilian:
Sifted radium item starts with no NBT. Confirmed via /nbtedit on output bus.

PR #18599, opened 2025-01-01
Allow crafting a TiCo Flint Sword Blade & Bone Wide Guard

YannickMG:
With the removal of the Gregtech Flint Sword there's a capability gap between the Wooden Sword & the Bronze Broadsword.

image
image
image

There is already a Flint Dagger which can be crafted, but it lacks a block ability (you throw it instead) and has lower damage than the Wooden Sword so I don't think it fits the needs of new players.

image

As a possible short-term fix, we could re-enable the Flint Sword Blade & Bone Wide Guard to allow crafting a Flint Broadsword. Incidentally this also allows crafting a Flint Rapier.

image
image

### DIFF
--- a/config/IguanaTinkerTweaks/restrictions.cfg
+++ b/config/IguanaTinkerTweaks/restrictions.cfg
@@ -24,12 +24,14 @@ toolparts {
         Flint:pickaxe
         Flint:shovel
         Flint:axe
+        Flint:swordblade
         Flint:knifeblade
         Flint:arrowhead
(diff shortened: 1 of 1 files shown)


Ethryan (approved):
Looks good, but it should definitely be added to the progression as a quest somewhere.

Dream-Master:
I think if it is limited to this I am fine with it. But it need to be mentioned in a quest . I can look into it later .

PR #18605, opened 2025-01-02
Fix twilight forest quest

Pxx500:
added a copy of the "What's that...?" quest in the novice tab so that it opens earlier and shows that the portal has a dependency

changed the icon in the "Searching for answers" quest because the icon was an end portal and its texture got changed

resolves [this discord message](https://discord.com/channels/181078474394566657/949447391587733504/1324156299432296630)


Ethryan:
242k edits?, my tablet would melt if I even tried to review that.

Dream-Master:
I had this yesterday as well when I add the sword quest. I just grabbed the two dueled and ignore the other lines. I am not sure if this is maybe some bug reformate all stuff after save the data base

YannickMG:
Could you both describe what you did to produce that diff? Did you use  some kind of bq_admin command to update all the quest files?

Anyway this PR has to be remade to only include the intended changes. If we want to reformat all the BQ files that should be part of a concerted effort and have it's own PR.

YannickMG (dismissed):
This PR has to be remade to only include the intended changes. If we want to reformat all the BQ files that should be part of a concerted effort and have it's own PR.

Pxx500:
> Could you both describe what you did to produce those automatic changes? Did you use  some kind of bq_admin command to update all the quest files?

literally just "/bq-admin default save"
it's pretty much impossible to make any quest changes without this unless you edit files directly 

the qb just formatted the lines upon saving

Dream-Master (approved):

Issue #18764, opened 2025-01-14
Coke Oven Brick Should Be Mineable By Hand, But Is Not


BeaulieuT:
### Your GTNH Discord Username

GoatsWithPants

### Your Pack Version

2.7.2

### Your Server

SP

### Java Version

Java 8

### Type of Server

Single Player

### Your Expectation

I misplaced a coke oven brick, and tried to mine it by hand. With empty hands, it says the block is "Harvestable". 

### The Reality

The block is not harvestable, clicking on it does nothing. Tried mining for several minutes, nothing happened.

### Your Proposal

I would like the coke oven brick to be mineable with empty hands. 

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [X] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

FourIsTheNumber:
Behavior broken between 2.7.1 and 2.7.2, presumably by https://github.com/GTNewHorizons/Railcraft/pull/65

Caedis:
IguanaTinkerTweaks' config file needs updating as its overriding it with a tool level of iron

PR #18776, opened 2025-01-15
Set tool level for RC water tank and coke brick to 0

Caedis:
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/18764

7 = Coke Oven Brick
14 = Water Tank Siding

### DIFF
--- a/config/IguanaTinkerTweaks/BlockDefaults.cfg
+++ b/config/IguanaTinkerTweaks/BlockDefaults.cfg
@@ -992,7 +992,7 @@ blocks_axe {
     I:"OpenBlocks:sponge:9"=1
     I:"Railcraft:cube:8"=0
     I:"Railcraft:machine.alpha:11"=1
-    I:"Railcraft:machine.alpha:14"=1
+    I:"Railcraft:machine.alpha:14"=0
     I:"Railcraft:post:0"=0
     I:"Railcraft:post:4"=0
     I:"TConstruct:slime.gel:1"=0
@@ -5953,7 +5953,7 @@ blocks_pickaxe {
     I:"Railcraft:machine.alpha:4"=2
     I:"Railcraft:machine.alpha:5"=2
     I:"Railcraft:machine.alpha:6"=2
-    I:"Railcraft:machine.alpha:7"=2
+    I:"Railcraft:machine.alpha:7"=0
     I:"Railcraft:machine.alpha:8"=2
     I:"Railcraft:machine.alpha:9"=2
     I:"Railcraft:machine.beta:0"=2


YannickMG (changes requested):
I forgot this needed fixing, sorry about that. I'm afraid this is an incomplete fix because every single railcraft machine block is affected (multiblock tanks, boilers, etc).

Copy pasting my post from elsewhere:
> I have a problem that needs insight into how we design blocks and want them to be harvested in this pack so the right decision can be made. We have 2 things that are in conflict:
> - Railcraft machines are hand harvestable and also have preferred tools and mining levels
> - IgnanaTweaksTConstruct has an event handler that dictates that railcraft's behavior is invalid. It treats all blocks which have a hand-harvestable material and preferred tools and mining levels as if the hand-harvestable flag was obviously there as an oversight and makes the block unminable by hand. (ExtraHarvestLevelHandler)
> 
> I think one of these 2 might have to go. My theory is that what ExtraHarvestLevelHandler accomplishes probably only matters for a few blocks across the pack which have an unfortunate combination of materials, tool levels and drop behaviors. I'd be inclined to add a config to turn off that handler in the pack and if need be correct materials of the affected blocks. (edited)

So that left us with 3 proper fixes:
- Disable ExtraHarvestLevelHandler
- Make all railcraft machines hand-harvestable (not require tools with specific mining levels)
- Give railcraft machines back a requiresTools material and change their harvesting code to only drop the machine if the proper tool was used.

edit: That being said we could also reduce the tool level of these 2 specific stone age blocks, but yeah it doesn't fix the underlying issue.

serenibyss:
> So that left us with 3 proper fixes:
> 
>     * Disable ExtraHarvestLevelHandler
> 
>     * Make all railcraft machines hand-harvestable (not require tools with specific mining levels)
> 
>     * Give railcraft machines back a requiresTools material and change their harvesting code to only drop the machine if the proper tool was used.
> 
> 
> edit: That being said we could also reduce the tool level of these 2 specific stone age blocks, but yeah it doesn't fix the underlying issue.

In my opinion, the bug is now fixed in Railcraft and retained previous behavior for all blocks with better WAILA behavior. Now we decide what should the pack actually do. I don't think there's a big issue in requiring a tool to mine "more advanced" Railcraft parts, and regardless I think this fix is good to get merged for 2.7.3 and potentially do something wider later, unless we can figure this out relatively quickly

Caedis:
I just tried this locally and it didnt seem to fix the issue.
I wonder what would happen if we straight up removed those 2 from the config file

YannickMG:
> In my opinion, the bug is now fixed in Railcraft and retained previous behavior for all blocks with better WAILA behavior. Now we decide what should the pack actually do. I don't think there's a big issue in requiring a tool to mine "more advanced" Railcraft parts, and regardless I think this fix is good to get merged for 2.7.3 and potentially do something wider later, unless we can figure this out relatively quickly

If we want to completely "require" a tool to mine "more advanced" Railcraft parts that's perfectly fine with me, but unless we want to complicate the code we should require one for all their parts. This would be consistent with Gregtech and with the previous waila behavior.

I propose that we make railcraft parts unminable without tools but apply this current PR to make stone age parts minable with stone age tools, which players will have available anwyay.

Caedis:
Looks like this file is regenerated every time the pack loads. I'll do some more in depth searching to figure out a solution.

PR #18922, opened 2025-01-31
Disable GalaxySpace ore generation

alphaomega21:
Set enableOresGeneration to false in the GalaxySpace config so GalaxySpace ores will not generate.


glowredman (approved):
FYI in some dimensions this config is not checked. I'm working on a PR which will fix this (among other things).

PR #19021, opened 2025-02-11
Fix some Typos in configs

Yoshy2002:
it came to my attention that many configs have typos, so it had to be done.
If anything would break anything, lmk.


Dream-Master:
You edit the config but I guess you better edit the errors inside the mods cfg ?

boubou19:
Did you boot the pack with those? Are they still corrected after a pack boot or do they reset to what they were?

Yoshy2002:
It seems like i misunderstood the configs, just went off of https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/19017 , which after investigation, doesnt change the actual config if you restart the game, due to it being generated inside ae2. I will rework this in the future

PR #19045, opened 2025-02-12
Add Salis Arcana Configs

koolkrafter5:
Add configs for Salis Arcana.

The only change so far is to disable the deprecated biome colors module.

Before adding SA to the pack, remember that tcinventoryscan should be removed because SA includes that functionality.


config/salisarcana/enhancements.cfg
```diff
@@ -0,0 +1,117 @@
...
+        16666
+        16666
+        16666
+        5555
+     >
+    B:stopCreativeModeItemConsumption=true
+    B:suppressWarpEventsInCreative=true
+    B:thaumcraftCommandTabCompletion=true
+    B:thaumcraftCommandWarpArgAll=true
+    B:thaumicInventoryScanning=false
```
Caedis, line 33: This is what takes the place of https://github.com/GTNewHorizons/ThaumicInventoryScanning

koolkrafter5: Accidentally left that off from when I added it to my gtnh instance. Fixed in [52d108a](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/19045/commits/52d108a145a016e50298116e2e63608b17a7cf3c)

Caedis:
@Dream-Master do we want chest content scanning behind a research?
If we do, i would suggest it being one of the first researches available.

https://github.com/rndmorris/Salis-Arcana/blob/main/src/main/java/dev/rndmorris/salisarcana/config/modules/EnhancementsModule.java#L212-L221

Dream-Master:
yes you can add it and make it early available

Dream-Master (approved):

PR #19259, opened 2025-03-07
Make DE Reactor viable for chaos shard generation

StaffiX:
Requires https://github.com/GTNewHorizons/Draconic-Evolution/pull/61

Makes the reactor be able to convert awakened draconium to chaos shard significantly faster.
With this, the reactor can generate 50 chaos shard every 6 hours, and can store 20 chaos shard as fuel.

The ratio of block of awakened draconium to chaos shard is 162 to 1.
The energy output is unchanged due to the reactor generating rf, with no way to convert it to eu.
Disables the world destroying explosion, with it off, it only explodes the reactor and the automation setup close to it


EnderProyects:
There was a draft/closed pr somewhere that turns the rf output from the reactor to eu

FourIsTheNumber:
Closed for good reason

config/DraconicEvolution.cfg
```diff
@@ -2,7 +2,7 @@
 
 "draconic reactor" {
     # Setting this to false will reduce the reactor explosion to little more then a tnt blast
-    B:EnableBigExplosion=true
+    B:EnableBigExplosion=false
```
Dream-Master, line 5: no why? gt machines explode as well. Thi is a reactor if you dont handle it with care it need to go boooom.

YannickMG: For what it's worth, we're hoping to make GT explosions much more manageable so they result in approachable repair gameplay rather than rollbacks.

EnderProyects: But if we are going to use draconic reactors to make shards instead of hunting chaos islands I think we should leave the normal explosion of the reactor

FourIsTheNumber: Why? What value does it add? Just annoying rollbacks

EnderProyects: If you setup the reactor correctly it will not explode, the value it adds its just make things correctly

FourIsTheNumber: It already explodes now. Changing this config to true just makes the explosion destroy your entire base completely unavoidably.

Dream-Master (approved):

config/DraconicEvolution.cfg
Dream-Master: not happy with the change

PR #19441, opened 2025-03-25
Synchronize IC2 reactors

ah-OOG-ah:
Enable IC2 nuke synchronization. Requires the [Hodgepodge PR](https://github.com/GTNewHorizons/Hodgepodge/pull/502)


ah-OOG-ah:
Drafting the config change until further investigation into TPS impact can be done

boubou19:
This really shouldn't be an issue except on big servers with many IC² nuke enjoyers, and even then most of the lag will come from inventory check and IC²->GT Enet conversion, as usual. If this is proven to be an issue on big servers, well, hopefully it's a config and it can be disabled.

boubou19 (approved):

Issue #19454, opened 2025-03-27
EFR adds Amethyst Geodes to Personal Dimensions (Nightly 912)


ogdae:
### Your GTNH Discord Username

rilliko

### Your Pack Version

Nightly 912

### Your Server

Elak93's OCE Server

### Java Version

Java 21

### Type of Server

Crucible

### Your Expectation

Amethyst Geodes should not spawn inside personal dimensions at all.

### The Reality

The new Amethyst Geodes from EFR do indeed spawn consistently inside personal dimensions. Scouting new chunks loaded more of them. 

![Image](https://github.com/user-attachments/assets/58347a8d-5135-4b7d-8572-6c1b976f92fe)

### Your Proposal

From looking at the configs it seems like dimensions can be blacklisted from spawning geodes, adding personal dimensions to this filter should fix the issue.

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [x] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

PR #19456, opened 2025-03-27
Amethyst only spawn in OW

Caedis:
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19454

### DIFF
--- a/config/etfuturum/world.cfg
+++ b/config/etfuturum/world.cfg
@@ -57,10 +57,11 @@ client {
 generation {
     # What dimensions should we ban amethyst geodes from generating in?
     I:amethystDimensionBlacklist <
+        0
      >
 
     # Treat the amethyst dimension blacklist as a whitelist instead, so geodes will ONLY generate in those dimensions, instead of excluding those dimensions from generation. [default: false]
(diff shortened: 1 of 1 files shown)


EnderProyects:
Ups, btw could you also add twilight forest and non ice planets? :3

Caedis:
Twilight Forest's y level is too low (think you spawn at y40 or something)

Caedis:
Get me a list of dim ids and I can add them, or add them yourself :P

EnderProyects:
Since the max level amethyst can generate is 46 I don't think there should be many problems, it will also add some decoration to the magic of tforest

EnderProyects:
> Get me a list of dim ids and I can add them, or add them yourself :P

I'm not at home ( and I don't have a laptop) so is going to be hard for me 😅

Dream-Master:
> Twilight Forest's y level is too low (think you spawn at y40 or something)

but so far ik i saw some Geodes in TF already

serenibyss (approved):
This is a good fix on its own, we can do a followup PR if we decide to add amethyst to any other dimensions

PR #19537, opened 2025-04-08
Updated config defaults to nightly 962

YannickMG:
To create this commit, I checked out the modpack repository on top of a prism instance of nightly 962 and ran the game a few times. I included all automatic changes to game configs except for timestamp updates, for non-whitespace changes to `defaultserverlist.json` which has a section helpfully labeled `DO_NOT_EDIT` (might want to move that one to a separate file) and the `config/modpack-update` file which is meant to be included in the pack but deleted on pack load. (The other way around would be nice so that we don't have to track it). I also added auto-generated config files from config folders where we had done so in the past.

Finally, I updated .gitignore to ignore all newly created files it didn't already account for and that I did not add to the repository.

### Uncommited changes after launching the pack now
image

### Uncommitted changes after launching the pack before
image


serenibyss (approved):
I trust that these changes are all automatically made when you launch the pack (aside from the gitignore change, of course)

PR #19563, opened 2025-04-11
New BM Meteors

EnderProyects:
Quartz meteor
image

Metamorphic stones meteor
image

Ice meteor
image

Amethyst meteor
image

### DIFF
--- a/config/BloodMagic/meteors/EFRAmethyst.json
+++ b/config/BloodMagic/meteors/EFRAmethyst.json
@@ -1,11 +1,11 @@
 {
   "ores": [
-    "etfuturum:amethyst_block:95",
-    "etfuturum:calcite:90",
-    "etfuturum:deepslate:80",
-    "etfuturum:blackstone:80",
-    "etfuturum:sculk:40",
-    "etfuturum:gilded_blackstone:20",
+    "etfuturum:amethyst_block:0:95",
+    "etfuturum:calcite:0:90",
+    "etfuturum:deepslate:0:80",
+    "etfuturum:blackstone:0:80",
+    "etfuturum:sculk:0:40",
+    "etfuturum:gilded_blackstone:0:20",
     "minecraft:monster_egg:1:10"
   ],
   "filler": [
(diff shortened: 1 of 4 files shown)


Nockyx:
remove ancient debris

EnderProyects:
> remove ancient debris

The point of that meteor is to give ancient debris, also it has a 2% chance

EnderProyects:
> > The point of that meteor is to give ancient debris, also it has a 2% chance
> 
> ok, so remove the whole meteor 👍

changed

AdityaVG13:
Hi, i think the Hive Meteor is a good meteor to include in the pack, but not at the current place where it's gated to in this PR. 

For one, the size of the meteor would mean a fairly large hive clump, which you give you hundreds of princesses and drones that you would have to scoop by hand (as far as I'm aware, the known meteor mining methods will not grab the hives). If the radius was reduced to like 5 or 6, i believe that the meteor would be in a much better position for players to actually get all the princesses and be able to store them all since they wouldn't have to spend a significant amount of time manually harvesting. 

2nd, I think that it's gated way too late in progression. The Advanced Mutatron, which is end ZPM to early UV is way, way too expensive for this type of meteor. The Advanced Mutatron already requires going down a significant portion of the breeding tree (to get Neutronium Comb), which by then you would have found other methods to gather a large amount of princesses, And while there's still a small jump between Gendustry stuff in ZPM/UV and the Mega Apiary, most players aren't going to be using hundreds of bees around that tier to have passive generation in iApiaries/Alvearies.

When you get a Mega Apiary, which you can get shortly after getting the advanced mutatron, having hundreds of bees is great, but the Swarmer mode on the Multiblock itself is far faster and easier since you don't have to do extra breeding. I propose that you change the item to an **Apiarist Chest** as that would be more HV-EV appropriate and would fit in with a smaller meteor which would provide the player a decent amount of princesses that would be beneficial to starting their bee journey. 

You are already able to find hives, buy hives from the QuestBook and get them from lootbags, so I don't think that this meteor needs to be gated so far into the progression and be as big as it is.

Dream-Master (approved):

PR #19592, opened 2025-04-14
Add assline note for >ULV busses

jurrejelle:
Adds questline note for https://github.com/GTNewHorizons/GT5-Unofficial/pull/4190

PR #19595, opened 2025-04-14
Revert "Add assline note for >ULV busses"

Dream-Master:
Reverts GTNewHorizons/GT-New-Horizons-Modpack#19592


Dream-Master:
i was to fast

PR #19646, opened 2025-04-22
Make geodes more rare and allow budding amethyst to be silktouched

felixfour:
I think that geodes should be spaced out a lot more, the highest value the config lets you do is to make them 2.4 times more rare

Also, if stuff like budding amethyst can be transported in AE2 spatial cells and with a dislocation focus and the like, I don't really see harm in letting them be silk touched


EnderProyects:
Check the discord thread

EnderProyects:
Also if calcite ratio is going to be nerfed from 9 to 1 I dont see the point of making geodes rarer

WanderingHero:
I would not make them more rare, a 1:1 nerf seems a good enough nerf.

felixfour:
oh whatever

PR #19672, opened 2025-04-25
Update meteor ritual to integer limit to work with my Armok pr

felixfour:
Still requires one capacity rune, Armok orb is (integer limit)-1
https://github.com/GTNewHorizons/Avaritia/pull/52

### DIFF
--- a/config/BloodMagic/meteors/T9Ores.json
+++ b/config/BloodMagic/meteors/T9Ores.json
@@ -8,6 +8,6 @@
     "OREDICT:oreTengamRaw:1"
   ],
   "radius": 16,
-  "cost": 1000000001,
+  "cost": 2147483647,
   "focusItem": "gregtech:gt.blockmachines:14009"
 }


Ethryan:
Not really a fan of just boosting the cost by about 1 billion without a solid reason for it.

felixfour:
The reason is so that rituals that have a really high scaling cost are usable with it again.

felixfour:
You would actually need two armok orbs, one in your hotbar and one charging in an altar with a capacity rune, to have the full integer limit and use it as the activation for a couple rituals that require 2.1 billion on top of the costs, but for the mark of the falling tower ones, just having one armok orb in your altar charging should be enough, i tested it

LazyFleshWasTaken:
> You would actually need two armok orbs, one in your hotbar and one charging in an altar with a capacity rune, to have the full integer limit and use it as the activation for a couple rituals that require 2.1 billion on top of the costs, but for the mark of the falling tower ones, just having one armok orb in your altar charging should be enough, i tested it

Would that even work? Afaik armok sets your network to 1bil, no matter what it was.

felixfour:
> > You would actually need two armok orbs, one in your hotbar and one charging in an altar with a capacity rune, to have the full integer limit and use it as the activation for a couple rituals that require 2.1 billion on top of the costs, but for the mark of the falling tower ones, just having one armok orb in your altar charging should be enough, i tested it
> 
> Would that even work? Afaik armok sets your network to 1bil, no matter what it was.

yes it would, if the armok value is increased to 2E31-2

PR #19687, opened 2025-04-27
Add optional resourcepack to fix fast leaves

donottellmetonottellyou:
The zip is [this](https://github.com/donottellmetonottellyou/GTNewHorizonsConfig/tree/main/resourcepacks/FastLeafOverride) with the GTNH logo added.

## Why

Angelica is a backport of very modern performance mods (and some other things). As such, there are bound to be incompatibilities where Minecraft itself handles graphics differently compared to older Minecraft versions. In addition, some older mods just don't handle graphics correctly in all instances, even in vanilla.

I have [made an issue](https://github.com/GTNewHorizons/Angelica/issues/934) for one problem related to fast graphics. This pull request is both a stopgap until (if?) that issue gets fixed there (since the solution involves using fancy graphics, and that messes even more with leaves), and fixes some other problems that may never be fixed (such as IC2 & Pam's leaves). The resource pack was originally used by me, and I just thought others might want to use it.

Some people, like me, grew up with bad computers, and are just used to and prefer the fast graphics leaves. Also, the fancy settings in Angelica don't work for everything (looking at you, IC2). Actually, IC2 is the main reason I created this resource pack, since its rubber trees are needed early game!

Note that this resource pack is intended for use with fast leaves on, it doesn't make sense to use it with fancy leaves (since it would just be slower for no reason). ~~This resource pack as it is represents the minimal changes needed for full consistency with fast leaves. This also fixes X-ray issues with fast leaves. If it is wanted for BOP, Natura, Thaumcraft, and special Twilight Forest leaves to appear fast with fancy leaves for consistency, I can easily duplicate their fast textures (I might force push those changes anyway).~~ However, nothing breaks consistency because all fancy leaves are overridden with fast textures.

## What

This pull request fixes many instances where fast leaves don't render correctly. How? For some (Forestry, BOP), an overriding texture is added with the fast graphics imposed behind the original texture. For some (IC2, Pam's), an alternative texture is needed because they simply don't provide fast graphics. GT++ and Witchery just have their fancy graphics overridden with fast. The Twilight Forest uses the vanilla fancy textures, which are also overridden to fix some trees being fast and some fancy (and the vanilla items always use the textures following graphics settings, so this fixes that too).

I think I got all the broken leaves, but feel free to let me know if I missed any. ~~Also, I think Pam's leaves look OK-ish in the nether, but I might put a rotated texture behind instead of the logs to maybe make it better?~~ Pam's texture has been updated to make it less garish. Let me know what yall think.

Vanilla:
VanillaBefore
VanillaAfter

BOP:
BOPBefore
BOPAfter

Forestry:
ForestryBefore
ForestryAfter

Giant Leaves:
GiantLeavesBefore
GiantLeavesAfter

IC2:
IC2Before
IC2After

Pam's:
PamsBefore
PamsAfter
(new texture)
netherLeaves

Witchery / Twilight:
WitcheryTwilightBefore
WitcheryTwilightAfter


donottellmetonottellyou:
My discord is the same as my username, @donottellmetonottellyou

Dream-Master (approved):

PR #19720, opened 2025-05-01
Lootbag tiered trash & general cleanup

YannickMG:
- Uses https://github.com/GTNewHorizons/EnhancedLootBags/pull/12 to give tech & other lootbags their previous lootbag as "trash" instead of the default trash bag.
- Gave witchery & HEE bags trash, including their previous tiers.
- Removed trash from the Legendary lootbag
- Removed all unobtainable lootbags except for the UHV lootbag (it's too beautiful not to be used)
- Removed an unevenly applied misleading tooltip suggesting that you enchant your lootbags with a Fortune III book, which led me to try to do that in an anvil as a noob.
- Re-order the tech lootbags in the config so they show up as a continuous line in NEI.

This PR doesn't depend on https://github.com/GTNewHorizons/EnhancedLootBags/pull/13 but it is required to achieve the removal of superfluous enchanted lootbags shown here:
image

### Merging notes

Please only merge this after https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1248 is merged.


Ruling-0:
I worry that the discoverability of the enchanting mechanic might be quite low with the tooltip removed - could it potentially be rephrased, or do you think something like an early quest callout might be better?

YannickMG:
I expect players to use NEI to figure out how to craft stuff but it's true that it might not be intuitive in this case. 

Telling you to enchant the bag was misleading and I'd rather not say that unless we actually let you enchant bags naturally and remove the tier component from the cost.

As you suggest an early quest to explain this might be appropriate.

Ruling-0:
> I expect players to use NEI to figure out how to craft stuff but it's true that it might not be intuitive in this case.
> 
> Telling you to enchant the bag was misleading and I'd rather not say that unless we actually let you enchant bags naturally and remove the tier component from the cost.
> 
> As you suggest an early quest to explain this might be appropriate.

Maybe a tooltip like:
"Combine with Fortune III and a dedicated
component to improve your odds"

YannickMG:
> > I expect players to use NEI to figure out how to craft stuff but it's true that it might not be intuitive in this case.
> > Telling you to enchant the bag was misleading and I'd rather not say that unless we actually let you enchant bags naturally and remove the tier component from the cost.
> > As you suggest an early quest to explain this might be appropriate.
> 
> Maybe a tooltip like: "Combine with Fortune III and a dedicated component to improve your odds"

So far I'm thinking of the following text: "Craft with a Fortune III book and a component to improve!"

One problem is that this tooltip is applied to all lootbags regardless of their enchantment status. I liked the consistency of not showing this message on already enchanted lootbags.

Ruling-0 (approved):
Good cleanup. Discussed on Discord implementing the fortune tooltip in Enhanced Lootbags to be mindful of enchantment level.

PR #19725, opened 2025-05-01
Disable One Ring in dungeon chests

YannickMG:
If we still want to give out, we can do so in more interesting loot tables like that of Roguelike Dungeons.

Alternatively, we could also re-enable crafting this item in the pack.


Dream-Master (approved):

YannickMG:
> I am strongly against this change. If we change the place where this item will appear, first of all we need to make that new place, and not remove the old one before it.
> 
> For now it looks like the removing the item. And it's unclear whether it will be added to new loot pool eventually or not.

It will be added back somewhere else as long as we're sure it's a net positive to the game experience. This being said anyone who strongly feels like they should have access to this item early on can just disable that config. 

Could you give us more context on why you think this item is important to have available to players?

Dream-Master:
We re add the ring if i understand @YannickMG right. But to not forget it we should choose a location now and make a pr.

YannickMG:
> The main positive thing here is that on such a low tier players can easily be killed by mobs, and getting the items back may be very tedious. The One Ring gives the opportunity to get them back by the cost of warp. After you achieve LV-MV the mobs are not the problem anymore, and that ring becomes useless, because you don't travel anymore, and the whole base is lit. Players also can deal more damage, have an armor, which also makes the life much less dangerous

Thanks for breaking that down, unfortunately while your use case as described is relatively benign, it intersects precisely with another use case which we do not want to enable. Let me explain why we're even making this change to begin with.

### The motivation for a change
First why are we changing anything at all? It's because we've started putting effort into combat content in the pack, notably with a total rework of the loot pool available in Roguelike Dungeons. The primary goal of that rework is for players who fight their way through such a dungeon to feel like it was worth their time. It is also explicitly a non-goal to change the meta in terms of what the best way of getting early tier metals like steel is.

### The effect of The One Ring on combat content

The way The One Ring is coded right now, it allows you to completely ignore enemies for as many spans of 5 minutes as you like, as long as you take a moment to die every 5 minutes. Indeed your mind corruption is reset on death and the ring was changed in 2018 to let you remove it anytime you like as long as you spam removing it, including before any permanent consequences set in. The ring is also easily obtainable in stone age within minutes of starting the game.

The effect this has on the pack is that you **cannot** design content at any stage of the game which relies on placing dangerous enemies close to items the players want to obtain, since they can always just wear the ring, obtain the loot, walk off and die somewhere secure.

### The One Ring combined with new dungeon loot

At this point I hope you will understand the concern this has. In beta testing, a tester found The One Ring, put it on and cleaned out a dungeon of most of it's loot within 5 minutes, netting enough steel (let alone other materials) to make cheesing dungeons that way by far the best way to obtain those materials. We know word of this would get around pretty quickly as a "loophope" so we're closing it early.

### How can we fix this?
- We can make The One Ring's corruption persist through deaths.
- We can give it a crafting recipe that's late enough that we think combat doesn't matter anymore, and push that recipe further into the game every time we manage to push combat content further into the pack.
- We can simply stop giving out the ring altogether because by definition, when combat doesn't matter the ring also won't matter to players. Unless there's strong demand for The One Ring to be made available in a way that doesn't entirely invalidate early game combat, this is likely to be where this ends.

I hope you can understand why we made the decision we did. In hindsight I should probably have included this write-up as part of the pull request.

### What you can do about it
Players are free to play exactly as they like. If you don't want to engage in combat content at all, feel free to turn on peaceful mode whenever it suits you. If you want to turn combat off with the ring the way you're used to, feel free to re-enable that config and use the ring however you see fit.

Rakdarian:
What.
No im strongly against this change, and I was strongly against the initial change back in 2020 that made the warp effects significantly more damaging.

As it currently stands, roguelike dungeons have had terrible loot. And it doesnt sound like thats going to be changing. Minecraft just as a whole isnt made to be particularly good at being a challenging combat pack. The ability to place and break blocks will just trivialize getting access to any of the dungeon loot if youre planning on upgrading it at all unless you intend to do something like RLcraft's implementation. 

The point at which most combat content becomes trivialized for most players of the game is right around the tool forge thanks to how strong the crossbow is, at its best, your point about having to die every 5 minutes is a serious commitment and pain in the ass to do, engaging with the game normally is significantly faster, and less frustrating then doing what youre suggesting, and at its best it hastens advancement to "combat doesnt matter for most applications" by a couple of hours at most.

Finally, there is at least one mob that punches through ring protection, which means its possible to implement if you really wanted to rework combat content. The chaos dragon of draconic evolution does kill you while wearing the one ring, so if you wanted to put loot next to a dangerous mob encounter you could both do so and do so without removing the fun early game puzzle that the one ring provides.


Now as far as bug fixing goes i wouldn't mind having the mental corruption persist through death, managing those warp effects and giving a strong incentive to beeline warp clearing magic research is one of the biggest reasons why I like the ring so much

WanderingHero:
I definitely think combat and tinkers weapons should be reworked, but the one ring does not help, it is an extreme bandaid fix of circumventing all of the combat instead of just fixing combat in nh.

Rakdarian:
Okay, but as it stands, you go from two viable ways to get passed the early game combat, to just one. And between the two of them the crossbow is far more straight forward and has no drawbacks.

WanderingHero:
The drawback is crossbow is actual combat and is not a guarantee way of not dying, i have died and seen people die going through rld with carbon crossbows

WanderingHero:
Also crossbow is gated much later than the one ring.

Miklebe:
> Okay, but as it stands, you go from two viable ways to get passed the early game combat, to just one. And between the two of them the crossbow is far more straightforward and has no drawbacks.

And you still have Bows, Shurikens and Throwing knives, which can work exactly the same and acquired as soon as you get tool station, unlike Crossbow. You can also gear appropriately (make iron armor as soon as you get hammer, etc) or play safe. I just started my fresh playthrough, found RLD dungeon and been raiding it just fine with 0 gear except stone sword and Throwing knife, 0 deaths so far, cleared 2 floors. No tunnel ratting, just simply going forward, stopping at intersections by blocking them off, and clearing them 1 by 1, no death rushing to place torches either. I kill mobs, go in, light up the area, repeat. Didn't manage to go deeper yet because generation was broken and part of the dungeon got cut off......

If you just refuse to deal with the hostiles, doesn't mean removing The One Ring makes it "now there is only 1 path".

WanderingHero:
The first two floors of rld are rather easy to beat, then it turns into armored inferno spam on the 3rd floor, might nerf the floor and nerf number of chests on that floor. But the main point does stand, there are more weapons, and you can still cheese stuff in other ways.

Rakdarian:
> > Okay, but as it stands, you go from two viable ways to get passed the early game combat, to just one. And between the two of them the crossbow is far more straightforward and has no drawbacks.
> 
> And you still have Bows, Shurikens and Throwing knives, which can work exactly the same and acquired as soon as you get tool station, unlike Crossbow. You can also gear appropriately (make iron armor as soon as you get hammer, etc) or play safe. I just started my fresh playthrough, found RLD dungeon and been raiding it just fine with 0 gear except stone sword and Throwing knife, 0 deaths so far, cleared 2 floors. No tunnel ratting, just simply going forward, stopping at intersections by blocking them off, and clearing them 1 by 1, no death rushing to place torches either. I kill mobs, go in, light up the area, repeat. Didn't manage to go deeper yet because generation was broken and part of the dungeon got cut off......
> 
> If you just refuse to deal with the hostiles, doesn't mean removing The One Ring makes it "now there is only 1 path".

Theres not a meaningful difference between one ticon weapon over another. Its the same gameplay loop between them. As far as specifically the roguelikes why are you spending so much time breaking into one. Youre putting a lot of effort for some non-standard ingots that you'll likely never use. Some low tier alloys that you would just be better off mining and smelting yourself.

None of the devs are going to make dungeons genuinely worth chasing. Youre not going to be getting enough steel to skip the bbf youre not gonna get a portal generator or traveler's gloves or a diamond lens or a diamond dolly, or hardened leather or aluminum bars or even the low tier thaumcraft equipment like thaumium fortress armor. 

Because if they did this would suddenly be the best way to get all of those items, ring or no ring, incredibly exploitably. I mean hell lootgame rooms were far less exploitable and they gutted those to be worthless for progression.

As I see it, your choices were for early game, build a powerful tinker tool to bypass combat, or find a one ring and build a tool later, at the cost of needing to either micromanage it or beeline warp cleansing research. Because dying every 5 minutes is a lot of deaths.

WanderingHero:
Rld got updated in 2.8 they have decent loot

WanderingHero:
But rld being almost semi viable is not the issue, an item that just avoids combat cant be balanced.

Miklebe:
New loot from some dessert themed RLD (i forgot to collect chests, so there is no counter for first and 2nd floor, they probably had like 7 and 10 chests)
image
image
image
image
image

Rakdarian:
> But rld being almost semi viable is not the issue, an item that just avoids combat cant be balanced.

It absolutely can, and this is a main problem with Minecraft as a whole. The humble bed and torch invalidates most combat. Torches negate the passive combat and is generally the  meta for most applications, tile entities, anti mob blocks and other vanilla and vanilla adjacent and then after that dream master armor, nano, high durability fortress armor, pdims, and basically everything after MV.

You strike me as someone who hasn't played GTNH all that much, most of the pack through one way or another invalidates combat. The ring is just one more alternate solution, and handling its drawbacks, especially after the nerf in 2020, is a fun, optional puzzle to solve for those of us willing to engage with it.

Rakdarian:
> New loot from some dessert themed RLD (i forgot to collect chests, so there is no counter for first and 2nd floor, they probably had like 7 and 10 chests) image image image image image

Interesting, this is definitely a little better than I was expecting, if thats just the first two floors dungeon loot that would be pretty reasonable for it depending on the odds for steel, wrought iron, silver, bows, and lootbag odds. But i have a feeling thats for the entire dungeon from top to bottom.

PR #19831, opened 2025-05-10
Update all lootbag icons to bappybags

YannickMG:
image

The default lootbag is unaffected, it is from the base mod and is never obtained in-game. Not worth touching.


Ruling-0 (approved):
Looks good, I can't imagine any controversy over this texture upgrade.

Issue #17005, opened 2024-08-14
Increase default FindIt range


leumasme:
### Your GTNH Discord Username

leumasme

### Your Pack Version

2.6.1

### Your Proposal

The current default range for FindIt (That's responsible for the "Find Item in nearby Inventories" hotkey) is currently 16 blocks.
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/blob/88d10f254e46db06c1408e1633df7eaea3fdfe7b/config/findit.cfg#L14
I propose increasing this default, for example to 32 or 64.

From looking at the code [ItemFindItService.handleRequest](https://github.com/GTNewHorizons/FindIt/blob/master/src/main/java/com/gtnh/findit/service/itemfinder/ItemFindService.java#L39-L50), [WorldUtils.getTileEntitiesAround](https://github.com/GTNewHorizons/FindIt/blob/master/src/main/java/com/gtnh/findit/util/WorldUtils.java#L15-L20), the range on its own doesn't have a performance impact, but instead performance cost scales with total amount of tile entites within all chunks within the scan range, performance scaling isn't actually $O(n^3)$ as one may assume. Should maybe do a bit of performance testing for situations where players have a very vertical base = many tile entities in the chunks in range?

### Your Goal

A range of 16 is very small. With how useful FindIt is, the small range punishes building prettier, more spread-out bases in the early ages. 

Increasing the FindIt range would have been a problem previously as the highlighting was hard to see from far away. With the new block highlighting and rotating the players head, this is no longer an issue and increasing the range to 64 and still being able to recognize all the highlighted blocks is possible.

### Your Vision

This is just some free quality of life. The Stockroom Catalogue sadly doesn't offer a comparable amount of QoL for finding items.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [X] I understand this change request may not attract enough attention and thus not be implemented.
- [X] I understand this change request may be rejected due to other community members think it's inappropriate.
- [X] I believe this feature would make the pack better.

leumasme:
still relevant

Dream-Master:
@leumasme make a pr to increase the value. i support this

PR #19861, opened 2025-05-13
Increase default FindIt range

meislucas:
Just commiting for leumasme since he has no access to a computer at the moment. 

Discussion/Explanation see here: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/17005

### DIFF
--- a/config/findit.cfg
+++ b/config/findit.cfg
@@ -33,7 +33,7 @@ general {
     S:SearchItemsOnGround=true
 
     # Radius to search within
-    S:SearchRadius=16
+    S:SearchRadius=64
 
     # Use Particle for Block Highlighter
     S:UseParticleHighlighter=false


YannickMG (changes requested):
While Dream's approval is important, I believe this is a major change with severe content balance and performance implications that needs serious discussion before it goes in before we determine which specific value is best for the pack.

Searching all inventories in range is not free and you are proposing to multiply that cost by 64 for very large densely packed bases .I think this at the very least needs to be measured so we can be sure the change isn't problematic from a performance perspective.

There's also a huge impact on people's ability to use this feature to remotely scan the contents of treasure chests in dungeons which I don't think are blacklisted from this feature.

Let's hold on until we know what the best value is and we're satisfied it won't hurt content balance. Individual players are of course free to change this setting in their own instance.

mitchej123:
I'd want to see some heavier profiles or some adjustments done to the code to self time box.  I'm not ok with this increase without proving it won't negatively impact things.

If dream wants it in NH he can do it via config until then.

boubou19:
This going from a 33x33x33 cuboid to a 129x129x129 cuboid scan. It's computing the chunks, then for each TE in the searched chunks, it checks for coordinates being inside the cuboid or not. So effectively, this makes it go from 3x3 chunks of search area to 9x9 chunks. Maybe there is a more reasonable middle ground, like just doubling the radius? That would make the search area of 5x5 chunks, It'd be still possible to cheese for dungeon loot search, but you need to be somewhat closer, so i think it's a reasonable tradeoff.

Dream-Master:
Sure I said 16 is to small but I am not sure 64 ist a bit much . Guess 32 fit better ?

Dream-Master:
About treasure Chests idk how much this can be abused

YannickMG:
> The chests aren't really hidden and I don't think we have much in the way of super rare loot, so it's not that abusable. Performance is a much more important concern.

Well we are actively working on adding much better loot to dungeons to make them meaningful while not overpowered, so I'm wary of any change which hurts that effort.

Dream-Master:
maybe we can try with 24 first so its a minimal increase and change to 32.

Dream-Master:
@meislucas any update here?

Dream-Master:
@meislucas any news? if not i close this now

LunarLambda:
Anything higher than 32 is IMO overkill.

If this is going to be closed due to lack of feedback, perhaps instead a note could be added to
https://wiki.gtnewhorizons.com/wiki/Commands_and_Configurations ?

Dream-Master:
the dev not asnwering or not test it so i close this for now

PR #19883, opened 2025-05-16
Make dirty portal generator optional, so you don't have to make it again

Ciborg085:
Problem:
The "Searching for Answers" quest has two requirements, the "Welcome to Tier 1,LV" will probably be completed before you can complete this quest (unless you free ride through someones portal in MP), and the "What's That...?" quest that requires you to collect a silver wood sap and log and greater wood sap and log, the tasks from the latter quest are not a strict requirement to make a dirty portal gen, so if you by any chance forget to complete this quest and make a dirty portal gen and use it to open the portal, you are not going to be able to complete this quest because you don't have the dirty portal gen anymore but have entered the twilight forest, of course unless you make another dirty portal gen, but then yo just wasted 4 circuits and a diamond and can't even recycle it.

image

So this change makes it so the dirty portal gen is a optional requirement to complete the quest.

Checked if it worked:
image

### DIFF
--- a/config/betterquesting/DefaultQuests/Quests/NoviceThaumaturg-AAAAAAAAAAAAAAAAAAAAFg==/SearchingforAnsw-AAAAAAAAAAAAAAAAAAAA4Q==.json
+++ b/config/betterquesting/DefaultQuests/Quests/NoviceThaumaturg-AAAAAAAAAAAAAAAAAAAAFg==/SearchingforAnsw-AAAAAAAAAAAAAAAAAAAA4Q==.json
@@ -100,7 +100,7 @@
           "id:8": "dreamcraft:item.TwilightCrystal"
         }
       },
-      "taskID:8": "bq_standard:retrieval"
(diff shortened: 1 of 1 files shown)


Ruling-0:
I think a better approach would be to just make the trees quest instead gate the wand. It makes far more sense - I have no idea why the TF quest is gated behind them.

Ciborg085:
> I think a better approach would be to just make the trees quest instead gate the wand. It makes far more sense - I have no idea why the TF quest is gated behind them.

That's true yea.
The other argument to have the dirty portal gen a optional item in this quest is in case there's a public TF portal or something in a MP server. It's not like having to make the item will make a difference, if you have already unlocked LV you can prob make the item but if there's already a portal you can use it seems silly to have to craft the item.

I'm busy right now but maybe Monday i can check out the quest book again and change the requirements like you said so it makes more sense and maybe add it to this PR or make a new one idk.

Dream-Master (approved):

Issue #19892, opened 2025-05-17
Add Proper Sapling Quantity to Forestry Quests


BillyBobJoe12345:
### Your GTNH Discord Username

fairbanks2126

### Your Pack Version

2.7.4

### Your Proposal

The Forestry Tree Breeding Quests have some trees that are non-1x1 girth saplings, this means that when you mutate/treebreed them you need more than 1, but the quests only show you need 1. I ask that the select few quests that it applies to get it changed so that the sapling count matches the proper amount needed per the girth gene

### Your Goal

To prevent confusion/annoyance.
IE if someone gets a sapling, the quest completes, they thing "Cool, I've bred the sapling, I'm done." Then they try to plant it and find out that they need 3 more since its a 2x2 girth  or 8 more since its 3x3 girth etc.

### Your Vision

It prevents confusion.

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [x] I understand this change request may not attract enough attention and thus not be implemented.
- [x] I understand this change request may be rejected due to other community members think it's inappropriate.
- [x] I believe this feature would make the pack better.

PR #19893, opened 2025-05-17
Forestry QB Changes

AdityaVG13:
Based on the suggestion of issue #19892  :

I changed the affected saplings in the Forestry tab to respect this suggestion.

This will help any plays doing Tree Breeding to understand the correct amount of saplings they will need to grow the requisite tree.

wenge before
wenge after

coast before
coast after

grandidier before
grandidier after

giant before
giant after


AdityaVG13:
For some reason when I pushed it, it deleted a bunch of other stuff but I have no idea how that happened. I'm still trying to learn. Will close, re asses what happened, and then try to do it without all the other stuff

PR #19894, opened 2025-05-17
Tree Breeding QB Changes

AdityaVG13:
Based on the suggestion in ticket #19892  , I've updated the Saplings in the Forestry Tab to reflect how many you actually need to grow the trees at the end of the line.

This should give the player a proper idea of how many saplings they need to make if they want to grow the tree that they are aiming for.

closes: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19892

wenge before
wenge after
coast before
coast after
grandidier before
grandidier after
giant before
giant after


Ruling-0 (approved):
Deletions appear to be the result of previous PRs not properly deleting.

PR #20032, opened 2025-05-29
backport red sandstone 3rd floor to the desert dungeon

WanderingHero:
In the code of roguelike dungeons, the 3rd floor desert dungeon should be red, but red sandstone does not exist in 1.7.10 until it was back ported by EFR. This uses the EFR red sandstone to recreate what it might look like.


WanderingHero:
image

PR #20102, opened 2025-06-02
Solar Factory Quest

AdityaVG13:
Added a quest for the Solar Factory on the Multiblock Goals Tab. I linked it to the HV Assembler quest, as you require one to actually create the multi itself.

I didn't add a quest reward because I'm not sure where this multi is really "placed." Since you won't make it till* ZPM, you shouldn't receive a ZPM loot bag if it's an HV unlockable multi. This could be changed, however, if necessary.


AdityaVG13:
What the quest looks like for reference: 

image

PR #20184, opened 2025-06-09
MagLev Pylon and Harness Quests

AdityaVG13:
Added 4 Quests for the MagLev Pylon and Harness. 1st Quest in MV to show the player, and then the additional pylons added in "build better bases"

@Caedis, for your review, good sir

image


Caedis:
Love it

Caedis (approved):

Pxx500:
You can also use multiplequestlines folder to put the first quest both into MV and build better bases tab

PR #20198, opened 2025-06-11
Permanent OffHand Quest

AdityaVG13:
Adds to the Checkmark quests at the start of the pack, "And so, it begin,s" by mentioning to the player that permanent offhand is now a feature and that the BG slots are removed.

image


Caedis:
Maybe add a note that says any items that were in the Battlegear slots should have been returned to your inventory or dropped if its full

Caedis:
And mention that `F` is used to swap the mainhand and offhand item

Dream-Master:
@AdityaVG13 just add as caedis suggested and we can merge the quest

AdityaVG13:
> @AdityaVG13 just add as caedis suggested and we can merge the quest

Will get to it tomorrow! I will reply to the thread when I'm done.

AdityaVG13:
@Dream-Master The changes have been made!

PR #20207, opened 2025-06-12
Add default configs for journeymap

Alexdoru:
Add default configs for journeymap to remove the useless info and make it not take so much space on the screen
Changed configs:
Minimap:
- shape : circle -> square
- size : 30 -> 25
- reticle : true -> false
- location format x y z (v) -> x y z 
- show mob heading : true -> false
- show grid -> truye -> false
- entity scale : 100 -> 135

Fullscreen map:
- location format x y z (v) -> x y z 
- show mob heading : true -> false
- show grid : true -> false
- entity scale : 100 -> 135

A comparison screenshot:
image


Ruling-0:
I think default should be kept circle - that's the original default and the choice between them is entirely personal. Personally I'd also keep grid, since it's not obtrusive and gives a good sense of distance. The rest seem good.

Alexdoru:
> I think default should be kept circle - that's the original default and the choice between them is entirely personal.

A circle is a waste of space on the screen, you occupy the whole corner while bringing less information than a square would. In fact the ratio between the area of both shapes is `π/4 = 0.78..`, that's about 20% of wasted space when using a circle over a square.

all the area in red is wasted :
image

Alexdoru:
> Personally I'd also keep grid, since it's not obtrusive and gives a good sense of distance.

Realistically do you use the grid on the minimap to estimate distance or do you just look in front of you in the world or better use the chunk overlay keybind ?

Ethryan:
I use the F9 to see it in world, the grid on the minimap is too small to be of much use.

PR #20229, opened 2025-06-16
Thaumcraft Research

koolkrafter5:
This makes some Thaumonomicon researches take research points instead of giving a research note to play the minigame with for every research.

### DIFF
--- a/config/Thaumcraft.cfg
+++ b/config/Thaumcraft.cfg
@@ -148,7 +148,7 @@ research {
     I:aspect_total_cap=500
 
     # 0 = normal, -1 = easy (all research items are directly purchased with RP), 1 = Hard (all research items need to be solved via the research table)
-    I:research_difficulty=1
+    I:research_difficulty=0
 }
 
 


Dream-Master:
<s>why this was easy. I remember it was set to hard when i start with GTNH

it was set 2016 wrong it need to be -1</s>

Caedis:
It was set to hard. Check #mod-dev, we have been discussing it

Dream-Master (changes requested):
i dont want this change

Caedis:
All "hard mode" does is make it so ALL researches require the tc4 minigame.
Default thaumcraft is fine, but the pack as over 500 researches to do. All normal mode does is allow some of them to be completable via paying aspects.

koolkrafter5:
> i dont want this change

Can you explain why? If difficulty is what you're after, having a few of them be directly bought with certain knowledge points can sometimes even be *harder* for players, since they would need to get and spend points for specific aspects, and the costs are often higher than if they had used their other points to connect the aspects in the regular researches. This setting just speeds up the process a bit for what is probably about 20% of the researches (pure guess on that number, but the large majority will still involve the research notes, especially since none of the custom GTNH ones have been set up to be the point-buy research type).

boubou19:
I think it's not the correct order of things to proceed:
- we don't know how many recipe it changes
- we don't have a clear view on what should and what shouldn't have a research minigame

clearing up those two things could help greatly in the decision making.

On a personal level, as much as i hate thaumcraft research minigame (online solver my beloved), removing in some researches the minigame and replace it by points does not solve the issue at all. You make only the research minigame rarer, but it's still super annoying to do. To tackle this, i'd be more in favor of a mix between a research minigame and some brandnew ones, as well as research points for some researches.

The two main issues that makes magic hated by a good chunk of the community is the research minigame and the kilo tons of warp making it not worth to do before having the possibility to craft the pure tear so you can just speedrun the researches and never care about it again. So it's time we do something about it.

Caedis:
2.9: The Magic Rework

Dream-Master:
remove the minigame by boring click game will not improve the game at all. If there is a nice new mechanic i am open for changes but this should be made in a AOI first and added later.

nshepperd:
Replace ordo spam minigame with a ritual where you craft a certain item (i guess one with related aspects) and then obliterate it dramatically to complete the research note.

serenibyss:
Closing for now, we should try to approach this with an AoI to look at the bigger picture than make this change hastily

Issue #20252, opened 2025-06-21
Add tuff, calcite to forestry backpacks


davqvist:
### Your GTNH Discord Username

Qvist

### Your Pack Version

Daily 06-14 (but I checked, it was not added in the last week on github)

### Your Proposal

Please add tuff, calcite (and potentially amethyst blocks) to the forestry diggers backpack. 
I have no silk touch so I cannot test, but looking at the config file deepslate is missing too. Cobbled Deepslate works already as it is oredict to cobble. 

### Your Goal

Make the mining experience in the bottom layers more enjoyable ...

### Your Vision

... and it would be more in line on what the diggers backpack did before and is supposed to do.

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [x] I understand this change request may not attract enough attention and thus not be implemented.
- [x] I understand this change request may be rejected due to other community members think it's inappropriate.
- [x] I believe this feature would make the pack better.

dibbydoda:
Closed by: https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/20294

PR #20294, opened 2025-06-28
Add EFR blocks to biggers backpack

MalTeeez:
This PR adds relevant worldgen blocks to the diggers backpack, as requested in #20252 .

Specifically:
            - deepslate
            - tuff
            - calcite
            - amethyst_block
            - coarse_dirt

### DIFF
--- a/config/forestry/backpacks.cfg
+++ b/config/forestry/backpacks.cfg
@@ -12258,6 +12258,12 @@ backpacks {
             appliedenergistics2:tile.BlockSkyStone
             appliedenergistics2:tile.BlockSkyStone:1
             etfuturum:magma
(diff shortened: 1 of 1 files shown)


Ethryan:
Personally, if the Amethyst Block is a block of amethyst, I think it belongs in the Miner’s Backpack and not the Digger’s one.

MalTeeez:
> Personally, if the Amethyst Block is a block of amethyst, I think it belongs in the Miner’s Backpack and not the Digger’s one.

I thought about that, since it also was in the original issue - Should we ask another person?

dibbydoda (dismissed):
I would agree that Amethyst is better suited in Miner's not diggers

PR #20346, opened 2025-07-05
typo chlorbenzene

Eldrinn-Elantey:


serenibyss (approved):
Updating the lang manually is fine, but just fyi it will happen automatically if you dont do it manually

PR #20382, opened 2025-07-09
Remove chem plant manual

koolkrafter5:
I can't test it right now, but this should remove the task for the chem plant manual


Dream-Master (approved):

koolkrafter5:
Tested it and it does indeed fix the problem when opening the quest and when searching

PR #20407, opened 2025-07-13
Enables StorageDrawers Fancy Item Render Mode

leagris:
Fancy Render Mode has been significantly optimized:
- Over 10× faster rendering
- 100× fewer temporary object allocations

With these improvements, enabling 3D item display over drawers is now practical and efficient.

See details and benchmarks in:
https://github.com/GTNewHorizons/StorageDrawers/pull/44


Dream-Master:
Did it mean we can allow bigger cubes of drawers per controller ?

boubou19:
No, but since Léa improved the rendering of the items inside the drawers, we now can enable fancy rendering without impacting FPS

PR #20647, opened 2025-07-29
Correct brittle charcoal fortune hint

crumblbee:
According to the [wiki](https://wiki.gtnewhorizons.com/wiki/Charcoal_Pile_Igniter#Usage) and my own testing (v2.7.4), using fortune on brittle charcoal does _not_ increase drops. That is in contradiction to the loading-screen hint stating the opposite.

This PR mends that, the hint now states that fortune does not increase drops from brittle charcoal.


crumblbee:
Feel free to change the message arbitrarily.

There are also some other loading-screen txt-files that wrongfully state fortune increases drops. These seem to be for other languages (left in English because maybe the hints aren't translated into that language yet, maybe) and use a different wording. I did not change those. You can find them easily by searching the repo for "brittle" and "fortune".

This PR also does not address any loading-screen hints that are _already_ translated into other languages.

koolkrafter5 (approved):
Looks like it always drops either 2 or 3 charcoal and does not check fortune:
https://github.com/GTNewHorizons/GT5-Unofficial/blob/master/src/main/java/gregtech/common/blocks/BlockReinforced.java#L325

Message text seems fine. I don't see any other language files like you mentioned.

crumblbee:
Sorry, I should have been more specific when mentioning other languages. Searching [across the entire GTNewHorizons org](https://github.com/search?q=org%3AGTNewHorizons%20fortune%20brittle%20charcoal&type=code) yields other loading screen txts in other repos. This includes:
- Translations of the loading screen in the GTNH-Translations repo, e.g. the [japanese translation](https://github.com/GTNewHorizons/GTNH-Translations/blob/1900ce231436020ad2824ec18e101d6a93ddb7d1/ja_JP/config/Betterloadingscreen/tips/ja_JP.txt#L62) (which seems to be an old copy of the english tips), but also other translations. I don't know how you like to handle these files, otherwise I would consider submitting a PR in that repo.
- [This `en_US.txt` file](https://github.com/GTNewHorizons/BetterLoadingScreen/blob/75d477a4822f4dbba5662d2dba41926c8f471c5e/src/main/resources/assets/betterloadingscreen/tips/en_US.txt#L61) in the BetterLoadingScreen repo. I don't know if this file is used at all, its most recent commit was two years ago. This seems to be the only relevant file in that repo.

PR #20663, opened 2025-07-31
Enable the enchantability of EMT tols

YaGirlAlexhupp:
Attempt to enable enchanting for the non Thaumic boots mod boots, this might also enable enchantability of other things, if that is an issue it might be best to try and rectify this at the source code which could be done as well, which would simply require removing the config option and changing it on the boots to be always enabled

### DIFF
--- a/config/EMT.cfg
+++ b/config/EMT.cfg
@@ -131,7 +131,7 @@
     B:"Enable capes"=true
 
     # Warning: the enchantability is low.
-    B:"Enable enchanting tools"=false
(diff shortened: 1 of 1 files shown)


UltraProdigy:
I like the idea. Though I'll wait for some other opinions, I'm not an EMT expert so I don't know if there is a reason we keep this off by default

YaGirlAlexhupp:
thats why I suggested the potential of editing the source code if needed, as it would bring them in line with the rest of the Boots

PR #20698, opened 2025-08-03
Update Electric Jetpack quest

AndrewCoon:
- Fixed old reference to GT++ Charging Pack in the Electric Jetpack quest
- Clarified how to charge the item


UltraProdigy (approved):

danyadev:
I'm currently on 2.7.3 and the note about the Charging Pack was super helpful. The thing is, (I guess) the item's name changed to Charge Pack, which is craftable in MV (even the MV one with 768k EU capacity) and after being placed in a belt slot charges armor and items in hotbar

Unless something has changed about this item in 2.8, I think we should return back mention about the Charge Pack

AndrewCoon:
That would make sense, I didn't see the name change anywhere, I can update it to reflect that in a few hours. The quest was confusing to me in 2.7.4 because I couldn't find the item.

PR #20745, opened 2025-08-06
remove large cells from the blocked items

boubou19:
This PR removes the large cells from being blocked by the game when you try to right click with them. They work correctly from what i can tell, and they were blacklisted 9 years ago, so probably something that got fixed without us noticing.


Alexdoru (approved):

chrombread:
Does this allow big cells to be right clicked on anything (including tanks)?

boubou19:
yes

Dream-Master:
lets test this a bit to be sure we have not any bug missed

PR #20941, opened 2025-08-24
Quick Quest Changes

UltraProdigy:
### Changes:
- Removes mention of battlegear from quest (I double checked it's no longer mentioned anywhere)
- Makes the epoxy board skip more clear


sisyphussy (approved):
Small desc changes

PR #20966, opened 2025-08-27
Quest changes

HFPTetraUro:
Mass Processing:
* Move Industrial 3D Copying machine to EV, since it can be crafted then and doesn't actually require the ABS
* Add missing rewards to quests (lootbag of appropriate tier)
image

* Make Dangote upgrade chip & 200 casing requirement optional, only the controller needed to complete the quest
image

IV & EV:
* More "Supersized Buses" quest to EV, since the 1st tier of super buses can be crafted in EV.
image
image

Storing and Transforming EU:
* Add quests for EV  and IV 4A, 16A and 64A dynamo hatches, to let players know of their existence for their LSC
image
image
image


UltraProdigy:
Awesome changes. Just one minor thing, could you possibly align the 3D copy machine to be on the symmetrical corner like the others are.image

HFPTetraUro:
Made the fixes, and also used the missing space for a quest for the t2 maceration stack :p
image


image
image

Issue #14265, opened 2023-08-19
No mining level in primal crusher


Kriion:
### Your GTNH Discord Username

Kriione[Kirill]#4672

### Your Pack Version

2.3.6

### Your Server

Zeta

### Java Version

Java 19

### Type of Server

Open To LAN

### Your Expectation

The expectation when crafting a Primal crusher was that its mining level would be at least as high as a fire pickaxe. 
![image](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/142689341/10090138-2f4f-48cc-ae0c-37143b7b8ea2)

### The Reality

There is no mining level at all. Anything that is not stone or earth does not allow mining at all. 
![image (1)](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/assets/142689341/52eb8360-f666-4201-964b-d89293642b9d)

### Your Proposal

Issue a primal crusher mining level equal to a fire pickaxe or higher. Otherwise, it is meaningless.

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [X] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

desagas:
Is this not the item that can mine bedrock? Or is that another item? I was think this because, if it was, would it truly require displaying a mining level when it can literally mine everything? I might be thinking of a different item.

Kriion:
HI.
Another item. This tool cannot dig anything at the moment except stone,
cobblestone, sand, earth and gravel with a 3x3 block area. Blocks such as
painted clay and so on, only 1x1 (Apparently there is not enough mining
level). Ores and everything that requires a waile mining level above 0
cannot be extracted in principle.
The object you are talking about is the ichor pickaxe.
[image: image.png][image: image.png][image: image.png][image: image.png]

вс, 20 авг. 2023 г. в 22:09, Denver ***@***.***>:

Nikolay-Sitnikov:
Closed in https://github.com/rndmorris/Salis-Arcana/pull/206 (if properly configured.) Can you confirm on the latest nightly?

PR #21037, opened 2025-09-01
Makes primal crusher tooltip visible and functional

UltraProdigy:
### Changes:
- This updates the mining level of the Primal Crusher to level 8. This makes concurrent sense given that it's a combination of 4 strong level 7 tools, having it at 5 SA & -1 ITT made it invisible.
- Updates both the ITTweaks & Salis Arcana config with this information. (SA says 6, that's directly akin to 8) 

Currently all of the mining level info is concurrent within them both with the same and accurate information despite them doing the same thing. If this is something we wish to change by moving all items from one config to the other then we can do that but this is the status quo as it stands so I'm just maintaining that. This is open to change however, I'm not particularly swayed in one direction.

This will close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/14265
This was briefly discussed with @Nikolay-Sitnikov : https://discord.com/channels/181078474394566657/522098956491030558/1411856035593064448

###In-Game:
image


Nikolay-Sitnikov (approved):
This should work fine. I still don't know why a Tool Override is necessary for the tool's mining level to be visible in its tooltip (even though the override is the same as the original), but only Salis Arcana properly affects WAILA's display of the mining level.

PR #21047, opened 2025-09-01
Make thaumometer scanning faster

sisyphussy:
GTNH has a ton of items, and scanning them all with the thaumometer takes ages, especially considering you unlock it in LV, at which point you've already accumulated a lot of unique items & machines that all need to be scanned. This makes it feel more like a repetitive chore instead of actual fun engaging gameplay.

Combining that with the fact that thaumcraft has hundreds of researches, this makes the whole early game of thaumcraft quite boring. Making the scanning faster would speed up the boring part and incentivize more people to explore the magic mods (all magic mods rely on the fact that the player has some early-game thaumcraft research)

Since there is no documentation on the new feature from salis arcana, where you can scan all aspects inside of a chest at once, many players won't do the research required for it until after they spent their entire time scanning every chest.


UltraProdigy (dismissed):
Approving because the code is good. Freezing so that Devs can discuss it. I personally think this is a good idea.

UltraProdigy:
Okay so the prevailing opinion here is that rather than doing a bandaid fix for something that already has a fix but is poorly documented is simply to document it better. Maybe instead of moving forward with this we can add a quest or a note early on in the Thaumonomicon (or both).

sisyphussy:
I also think more documentation would be good, but at the same time i don't think reducing the tick duration would be a bad thing. You still need some aspects to complete the research for that feature (and also having to scan like 30 items to unlock all aspects), and it's not like you'll just stop using the inventory scanning feature once you unlock that research. Every new item that you unlock afterwards should also get scanned.

wlhlm:
@UltraProdigy 

> Okay so the prevailing opinion here is that rather than doing a bandaid fix for something that already has a fix but is poorly documented is simply to document it better.

This comment makes no sense in the context of this PR here with two contributors arguing for the change. I assume you are referencing a conversation on Discord.

UltraProdigy:
> This comment makes no sense in the context of this PR here with two contributors arguing for the change. I assume you are referencing a conversation on Discord.
> 
That is exactly what I'm doing.

@Dream-Master Can we get Sisyphus stakeholder so they can see MetaDev

sisyphussy:
After discussion, we've agreed that it'd be better to instead add more documentation on the salis arcana feature, where you can scan the entire container at once.

PR #21060, opened 2025-09-01
added thaumometer scanning documentation

sisyphussy:
- added the container scanning feature (discussed in https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/21047)
- made warp quest not worded as scarily (i've genuinely seen people in general who didn't want to start with thaumcraft because of warp)


Miklebe:
I know that some researches give different warp values (i think blood magic stuff is one of those? the one that are taboo), but maybe worth mentioning how much warp each level of forbidden knowledge gives?
So if anything, player could also calculate their own risks on gaining some warp from "mostly harmless" and etc stuff

sisyphussy:
Good idea, however I'm too lazy to do it today. If you want to you can fix it but i won't until tomorrow

UltraProdigy (commented):
For the first quest, in the first section use 'them' not 'it'. Instead of it'll just use 'it will' (just trust me on that one) and instead of 2nd just type out second. Also I usually change the color of the text explaining the research path to be a [/note] but if you want to keep it the way it is I won't argue

UltraProdigy (approved):

PR #21067, opened 2025-09-02
Pt1 of "Reward Area" Quest Tab Decommission

UltraProdigy:
### Changes:
- Deletes all lootbag conversion quests, besides the stone age -) steam one.
- Moves the stone age -) steam lootbag conversion to the steam age tab
- Moves the "Chef Training" quest to the fishing farming cooking tab
- Adds new quest in Tips and Tricks for Uptiering lootbags 

This is the first part of a 3 PR chain. This empties the Reward area quest tab entirely, the next PR will remove it from the questbook altogether. The third PR will add assembler recipes for missing lootbags in addition to the regular recipes that exist already.

This was discussed with @Dream-Master 
This idea was brought up by @LewisSaber 

### Why:
- This tab is effectively just clutter and causes issues with obtaining quest master legitimately. There are assembler recipes that make this an actual balanced feature due to the EU cost and tiering. The Basic to steam bag quest was kept as you can imagine because there is no steam age assembler, and the chef training quest was kept because it has nothing to do with lootbags. There was a quest added to explain the uptiering of lootbags so the player is aware of the existence of the recipes now, which is more than could be said before.

### Lootbag Uptier Quest (In Tips & Tricks - making it a perfect rectangle now)
image

### DIFF
--- a/config/EnhancedLootBags/LootBags.xml
+++ b/config/EnhancedLootBags/LootBags.xml
@@ -2075,7 +2075,7 @@
         <Loot Identifier="f09323b5-b81a-4472-af03-72acad499184" ItemName="witchery:ingredient:96" Amount="1" NBTTag="" Chance="25" ItemGroup="" LimitedDropCount="0" RandomAmount="false"/>
         <Loot Identifier="2a722f83-c9c0-480b-9d6d-858fd65d08d4" ItemName="witchery:ingredient:99" Amount="8" NBTTag="" Chance="85" ItemGroup="" LimitedDropCount="0" RandomAmount="false"/>
     </LootGroup>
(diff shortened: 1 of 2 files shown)


UltraProdigy:
Making this a draft until I have follow up PR with new lootbag assembler recipes.

cubefury:
I shall only be migrating the coins tab to the vending machine with this in mind!

UltraProdigy:
This is ready for review. For the person reviewing; don't bother looking at the code in detail it's just quest removals and a couple file movements. The only thing that may need edits is the new quest which is depicted above for convenience.

LewisSaber (approved):
LGTM

PR #21147, opened 2025-09-06
Enable shulker spawning in end

Caedis:
No end cities so they spawn like enderman in the end, just not as common.


EnderProyects:
Maybe it could be changed to spawn in HEE estructures? or some HEE biome?

Caedis:
but thats more work :P
Feel free to submit that pr. Would that go into coremod?

EnderProyects (approved):

EnderProyects:
I suppose it will be the same as your pr of lich tower

Dream-Master:
Are End cities not generated at all or not implemented yet?

Caedis:
Not implemented

Dream-Master:
but they will in the future ?

PR #21216, opened 2025-09-10
Update "The Easy Way to Change Biomes"

DylanTaylor1:
The current quest and description is awkward and confusing. After having helped people with the !seasons infographic on discord I went and updated the quest to look more like it. Added Optional Task for the optional radius increases and another retrieval task for the book of biomes, gunpowder, and glass bottle. Removed Mandrake Root from the first retrieval task and quest rewards because it is unnecessary.

image


UltraProdigy (approved):
Gorgeous

Issue #21271, opened 2025-09-14
Placed Anvils turn into et futurum requiem very damaged anvils, which causes forge hammer multiblock to be unable to form


Wokafishh:
### Your GTNH Discord Username

larvisen

### Your Pack Version

2.8 beta 4

### Your Server

SP

### Java Version

Java 21

### Type of Server

None

### Your Expectation

Build the forge hammer multi

### The Reality

Placed anvil is replaced by one from et futurum. Caused the mb to not form

### Your Proposal

 -

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [x] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

Yoshy2002:
Can confirm on Daily 132
Either MB has to accept the new slightly damaged anvil or anvils dont transform into slightly damaged anvils anymore (which would make more sense)

jordanamr:
This also affects every other anvil in the game (steel, thaumium, void metal anvil...) Their texture looks like the anvil is new, but the actual block placed becomes a Very Damaged Anvil from EFR

<img width="1912" height="1080" alt="Image" src="https://github.com/user-attachments/assets/537ea12c-53f3-4aaa-8984-dbd308a1111d" />

jordanamr:
There's a quite interesting issue with EFR actually, I just discovered that by placing the anvil in different directions, the anvil type changes, it seems like depending on the direction, the anvil is fine at first, and then instantly "rotates" and changes its durability (look at WAILA on this video):

EDIT: It seems like the durability change is a Waila bug, the anvil is indeed mint

https://github.com/user-attachments/assets/5097547a-b95f-4d75-9b8e-409de4949b73

aka13-404:
I just experienced this with the void metal anvil on 2.8.4. The video with the orientation helped - approaching from the other side placed a pristine anvil, and the forge hammer was able to form.

PR #21315, opened 2025-09-18
fix: disable new anvils from EFR

jordanamr:
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/21271

Following the discussion on Discord: https://discord.com/channels/181078474394566657/401118216228831252/1418190294410723389

I think we should disable the new anvils from EFR due to known and potentially more uncovered issues.
It might be a good idea to check other block replacements from EFR too, especially ones that are used in multiblocks for example.


boubou19:
EFR anvil needs proper integration in the pack before getting enabled

Issue #21325, opened 2025-09-18
Shulkers should not spawn (as often?) in the outer lands


ReaganAR:
### Your Pack Version

2.8-beta4

### Your Proposal

Shulkers should either not spawn in this dimension/biome or spawn at significantly reduced rates

### Your Goal

Oftentimes while exploring the outer lands maze structure, I will turn the corner and see 3-5 shulkers all within a few blocks of each other. this means that I am finding at least 5 times as many shulker boxes as I am eldritch guardians and shambling husks

### Your Vision

before EFR was added, the only mobs that would be found in this dimension were thaumcraft specific enemies meant for this area, shulkers don't fit the vibe and make the maze feel even more cluttered and annoying to deal with, especially with how many there are

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [x] I understand this change request may not attract enough attention and thus not be implemented.
- [x] I understand this change request may be rejected due to other community members think it's inappropriate.
- [x] I believe this feature would make the pack better.

PR #21327, opened 2025-09-18
Disable shulker spawns in end

Caedis:
Thaumcraft's dimension for whatever reason has the end tag in one of its biomes which is causing shulkers to spawn 1/3rd of the time due to the weights thaumcraft is using.

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/21325


chrombread:
how will players get shulker boxes if the shulkers are disabled? (if im understanding this config right)

Caedis:
The Twilight Forest Lich and Ghast towers

chrombread:
ah, they can spawn there? thats fine then I suppose. ill wait for others to give their opinion though, maybe.

wlhlm:
See https://github.com/GTNewHorizons/twilightforest/pull/121

chrombread:
ah ok. so this basically changes nothing, i see

chrombread (approved):
as a non-intrusive bugfix, lgtm

PR #21328, opened 2025-09-18
New botania quests

EnderProyects:
- Made gaia quest a required quest for DEFC quest
- Added quests for loomium, asargardandelion, creative mana pool, tablet, and 1 and 20 tier laputa shards
image
image
image
image
image
image
image


UltraProdigy:
These look awesome. Maybe though before the google doc you could just explicitly state what it's for. "Here is a list of all possible drops" or something

EnderProyects:
image

Issue #21341, opened 2025-09-20
Monthly Ticket for Quest Changes


UltraProdigy:
## Overview:

The running ticket for quest changes; a ticket for any problems with any quests. Just post below and I will get to your comment as soon as I can. For more in-depth suggestions or problems, creating a [Dedicated Ticket](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues) is better. This ticket is more focused on small and clear problems and suggestions that don't really need any larger discussion.

Below are the remaining things from the last ticket (https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/17882)

The ticket will refresh upon the release of 2.8.1

#### To See Latest Questbook:
1) Be sure your GTNH instance is reasonably up to date
2) Grab the most up-to-date quest files from [Here For 2.9+](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/tree/master/config/betterquesting/DefaultQuests) or [Here For 2.8.X](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/tree/release/2.8.x/config/betterquesting/DefaultQuests) and download by either pasting the URL in [here](https://download-directory.github.io/) or forking the pack and grabbing the folder from the appropriate branch.
3) Rename the folder "DefaultQuests" if needed and replace the folder in your `.minecraft\config\betterquesting`
4) Either reload your world, or if you hot-swapped the folder in game, use `/bq_admin default load` after new folder is in (this will **NOT** impact your quest progression).

## Quest Changes (Also see Sub-Issues):
- [x] Add info to the cactus and sugar charcoal quests that they can now be compressed to last longer
- [x] Reorder all ore requiring quests so raw ore comes before the ore block
- [x] Fix Shulker Spawn Details
- [x] Move Personal Healing Nanobooster to UV
- [x] Make Intermediate requirements optional for UV bacteria quests
- [x] Fix "You need them all part 2" quest to only require those which are craftable in stone age
- [x] Change GT Shadow Metal Ore to Shadow Ore
- [x] New quest for T4 Nano Forge

#### Post RC1:
- [x] Double MagLev MV Quests
- [x] T4 Nano Forge Misinformation
- [x] Prismarine Line
- [x] Change Shulker Quest Reqs. & Possibly Location
- [x] Change Deep Dark location to post FGen & Spell check
- [x] EV 64A Hatches Quest Fix
[...]

PR #21344, opened 2025-09-20
fix: allow to purchase zinc gravel ore before completing its quest

jordanamr:
Following https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/21341#issuecomment-3315136206

As the quest description says you can purchase some if you can't find any, allow the related coins quest to be purchased before completing the zinc gravel quest.

image


Dream-Master:
how this quest get triggered?

jordanamr:
> how this quest get triggered?

I just moved the dependency one step before the Zinc Gravel quest, so you can actually buy Zinc Gravel Ore when the Zinc Gravel quest is available, since that quest tells you that you can purchase the ore if you can't find any.
image

UltraProdigy (approved):

PR #21385, opened 2025-09-22
Reword concrete quest to indicate that Chisel and GT variants provide the same boost

wlhlm:
followup for https://github.com/GTNewHorizons/GT5-Unofficial/pull/4910


Vlamonster:
Interestingly the old text was wrong since Chisel concrete was actually faster than GT concrete.

PR #21466, opened 2025-09-26
Change Big Backpacks to 90 slots again

wlhlm:
There are some changes ([1](https://github.com/GTNewHorizons/Minecraft-Backpack-Mod/pull/19), [2](https://github.com/GTNewHorizons/Minecraft-Backpack-Mod/pull/17)) in the pipeline to improve the inventory layout with the original 90 slots, so go back to that for now. That means, the last slot in the grid will be empty with the current algorithm, but that's acceptable.

### DIFF
--- a/config/Backpack.cfg
+++ b/config/Backpack.cfg
@@ -15,7 +15,7 @@ general {
     # Number of slots a large backpack has
     # valid: integers 1-128
     # ##############
-    I:backpackSlotsL=91
+    I:backpackSlotsL=90
 
     # ##############
     # Number of slots a medium backpack has
--- a/config/GTNewHorizons/CustomToolTips.xml
+++ b/config/GTNewHorizons/CustomToolTips.xml
@@ -24,9 +24,9 @@
 	<ToolTip ItemName="StevesCarts:CartModule:61" ToolTip="§dPerpetual Locomotion" NBT=""/>
 
 //Bags
-	<ToolTip ItemName="Backpack:backpack" ToolTip="4*9 Inventory - 36 Slots" NBT="" MetaStart="0" MetaEnd="16"/>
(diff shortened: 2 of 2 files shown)


UltraProdigy:
Would reducing this take away the perfect rectangle or add it back? Not to be Captain OCD here but I'm team perfect rectangle and I think others would generally agree. The imperfect rectangle just appears sloppy imo.

wlhlm:
Yes, it will look like this for now for newly crafted backpacks (and old backpacks from before #20997 was merged):

483859242-90b70af6-8715-4db4-a01c-0930eca569a5

We'll be updating the layout algorithm for 2.8.1 so that it will again show as a perfect grid with 90 slots. This PR tries to prevent too many  91 slot backpacks from getting out into the wild.

Miklebe:
Merge algo changes right before stable and have a reason for RC3 :trollface: 
But if we can contain OCD for 1 stable release till minor version comes with algo fix, i think it would be indeed better so that players on stabled craft 90 slot backpacks, but otherwise the most glaring issue would be just backpacks having different layouts (13 slots per row with 91 vs 15 slots per row with 90 using new algo)
Also would need to change the slots here, might just add this to this pr since it's all in the configs
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/blob/756ba9d8c92963e139e37961954eaf866597351b/config/GTNewHorizons/CustomToolTips.xml#L29

wlhlm:
> Also would need to change the slots here, might just add this to this pr since it's all in the configs

Can you push a commit for this to the branch here?

Miklebe:
> > Also would need to change the slots here, might just add this to this pr since it's all in the configs
> 
> Can you push a commit for this to the branch here?

yeah, i can, gimme a sec

Miklebe:
Oh hey, it worked, cool :D

Dream-Master:
@TimeConqueror

Issue #19662, opened 2025-04-24
Things to do for 2.8.0


boubou19:
### Issues that absolutely need to be fixed before  2.8.0-beta1
- [x] https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19842

#### AE2
- [x] https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/766
- [x] https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/762
- [x] https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/759
- [x] https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/742
- [x] https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/733
- [x] https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/705
- [x] https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/756
- [x] https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/739
- [x] https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/755
- [x] https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/pull/703

#### AE2FC
- [x] https://github.com/GTNewHorizons/AE2FluidCraft-Rework/pull/302
- [x] https://github.com/GTNewHorizons/AE2FluidCraft-Rework/pull/294
- [x] https://github.com/GTNewHorizons/AE2FluidCraft-Rework/pull/278

#### Amazing Trophies
- [x] https://github.com/GTNewHorizons/Amazing-Trophies/pull/14

#### Botania
- [x] ~~https://github.com/GTNewHorizons/Botania/pull/75~~ postponed

#### Botanic Horizons
- [x] https://github.com/GTNewHorizons/Botanic-horizons/pull/33

#### Draconic Evolution
- [x] https://github.com/GTNewHorizons/Draconic-Evolution/pull/70

#### EFR (checkmark indicates whether this has been reviewed + approved by a GTNH dev)
- [x] ~~https://github.com/Roadhog360/Et-Futurum-Requiem/pull/572~~ postponed
- [x] https://github.com/Roadhog360/Et-Futurum-Requiem/pull/598
[...]

PR #21480, opened 2025-09-26
hotbar sort off by default

Pxx500:
hotbar sort off by default

### DIFF
--- a/config/bogosorter.cfg
+++ b/config/bogosorter.cfg
@@ -11,7 +11,7 @@ general {
     B:enableAutoRefill=true
 
     # Enable the hotbar sort feature. [default: true]
-    B:enableHotbarSort=true
+    B:enableHotbarSort=false
 
     # Enable the hotbar swap feature. [default: true]
     B:enableHotbarSwap=true


Pxx500:
https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/19662#issuecomment-3331111718
image

config/bogosorter.cfg
```diff
@@ -11,7 +11,7 @@ general {
     B:enableAutoRefill=true
 
     # Enable the hotbar sort feature. [default: true]
-    B:enableHotbarSort=true
+    B:enableHotbarSort=false
```
boubou19, line 19: Can you take the config file from RC2 so the new entries are also in the modpack repo?
I want `enableHotbarSort` to false and `enableNoGuiSort` to true to match inventorytweaks's previous behavior.

boubou19 (changes requested):

config/bogosorter.cfg
Pxx500: no guisort is enabled by default anyway

boubou19: it got disabled by default in the mod today, this is why it's important to sync configs, so changes in default for the mod don't impact the mod configs on mod update.

Caedis: I would want the config to also be disabled in the pack

PR #21513, opened 2025-09-27
Update GT Client.cfg, up line thickness from 2.5 to 5

PlayfulPiano:
Noticed that the current repo's version of the file was fairly outdated.

Changes listed come from a clean 2.8.0 client instance, minus one adjustment (changing line thickness for the wrench/tool overlay from 2.5 to 5 for better visibility).

### DIFF
--- a/config/GregTech/Client.cfg
+++ b/config/GregTech/Client.cfg
@@ -17,30 +17,13 @@ client {
         ##########################################################################################################
 
         cableinsulation {
-            #  [range: -2147483648 ~ 2147483647, default: 64]
+            #  [range: 0 ~ 255, default: 64]
             I:blue=64
 
-            #  [range: -2147483648 ~ 2147483647, default: 64]
+            #  [range: 0 ~ 255, default: 64]
             I:green=64
 
-            #  [range: -2147483648 ~ 2147483647, default: 64]
-            I:red=64
-        }
-
-        ##########################################################################################################
-        # constructionfoam
-        #--------------------------------------------------------------------------------------------------------#
-        # RGB values for the construction foam color modulation.
-        ##########################################################################################################
-
-        constructionfoam {
-            #  [range: -2147483648 ~ 2147483647, default: 64]
-            I:blue=64
-
-            #  [range: -2147483648 ~ 2147483647, default: 64]
-            I:green=64
-
-            #  [range: -2147483648 ~ 2147483647, default: 64]
+            #  [range: 0 ~ 255, default: 64]
             I:red=64
         }
 
@@ -51,13 +34,13 @@ client {
         ##########################################################################################################
 
         machinemetal {
-            #  [range: -2147483648 ~ 2147483647, default: 255]
+            #  [range: 0 ~ 255, default: 255]
             I:blue=255
 
-            #  [range: -2147483648 ~ 2147483647, default: 220]
+            #  [range: 0 ~ 255, default: 220]
             I:green=220
 
-            #  [range: -2147483648 ~ 2147483647, default: 210]
+            #  [range: 0 ~ 255, default: 210]
             I:red=210
         }
 
@@ -102,6 +85,9 @@ client {
         # if true, input filter will initially be on when input buses are placed in the world. [default: false]
         B:inputBusInitialFilter=false
 
+        # if true, input filter will initially be on when input hatches are placed in the world. [default: false]
+        B:inputHatchInitialFilter=false
+
         # If true, scrolling up while hovering a ghost circuit in a machine UI will increment the circuit number. [default: false]
         B:invertCircuitScrollDirection=false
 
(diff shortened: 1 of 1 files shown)


wlhlm:
We have to keep in mind that with the way the lines are drawn, this also makes the grid thicker for lower resolution. See here for a slightly exaggerated example (5.0 line thickness):



That would be a bit much for me, but I get that's better than not being able to see the grid at all. I'm interested in hearing from somebody playing at 4K with Shaders what a good value is for them.

I'm aware that this is a bit of a bikeshed on my part.

wlhlm:
@Caedis I read you play at 4K? Are you using shaders? What do you think would be a good compromise here?

Caedis:
Just check how GTCEu does it

PlayfulPiano:
> Just check how GTCEu does it

Yeah personally the GTCEu implementation would be the ideal but that's a full mechanic replacement. This PR was more a quick QoL default config change for 2.8.1 since that option is now customizable.

As for the resolution point, I think it's more a readjustment aspect than being too thick? Like I personally use a 10.0 line thickness, but I figured 5 would be a good compromise value for most players.

leagris:
Please make the line scale proportionally to GUI scale.

My argument is that no fixed value is a fit all.

GUI Scale is actually a valid indicator of user preferences of how large these things need to be drawn.

PlayfulPiano:
> Please make the line scale proportionally to GUI scale.
> 
> My argument is that no fixed value is a fit all.
> 
> GUI Scale is actually a valid indicator of user preferences of how large these things need to be drawn.

Oh huh I didn't realize this was affected by GUI scale. That would definitely be a better fix in general.

PlayfulPiano:
> Please make the line scale proportionally to GUI scale.
> 
> My argument is that no fixed value is a fit all.
> 
> GUI Scale is actually a valid indicator of user preferences of how large these things need to be drawn.

Actually just double checked, it doesn't seem like this gets affected by GUI Scale?

image
image

Kynake:
> > Please make the line scale proportionally to GUI scale.
> > My argument is that no fixed value is a fit all.
> > GUI Scale is actually a valid indicator of user preferences of how large these things need to be drawn.
> 
> Actually just double checked, it doesn't seem like this gets affected by GUI Scale?

Turns out its resolution dependent. https://github.com/GTNewHorizons/GT5-Unofficial/pull/4994 should fix this for all resolutions.

PlayfulPiano:
Still believe that adjusting to 5 should at minimum be tested out in experiments/dailies and see community feedback. Just to clarify, my issues with the thickness are independent from resolution as my screen is 1920x1080. I also personally use 10 line thickness, not 5 like this PR.

leagris:
> > Please make the line scale proportionally to GUI scale.
> > My argument is that no fixed value is a fit all.
> > GUI Scale is actually a valid indicator of user preferences of how large these things need to be drawn.
> 
> Actually just double checked, it doesn't seem like this gets affected by GUI Scale?

I was not saying that it is.

In its **current implementation**, line width is **not proportional** to the **GUI scale**,

I said that it **would be appropriate to implement the scaling of the line width proportionally to the GUID scale**.

RecursivePineapple:
Anyone want to block this PR before I approve it? It seems like everyone's concerns are satisfied with https://github.com/GTNewHorizons/GT5-Unofficial/pull/4994.

cc: @leagris

PlayfulPiano:
As I said in the other PR, I still think the default thickness should be increased in some capacity independent from the resolution-based functionality. I normally use 10 myself so I suggested 5 as a more conventional value to be more visually readable.

Dream-Master:
@PlayfulPiano @leagris @RecursivePineapple @Kynake @wlhlm @Caedis what we gonna do here ?

PlayfulPiano:
Have there been issues with the higher thickness in experimentals for others who've been using them? Or is is not as big of a deal to increase the default thickness here?

Dream-Master:
since zeta is dead i am not sure who test with experiemental?

PlayfulPiano:
> since zeta is dead i am not sure who test with experiemental?

Oh. Are dailies also dead?

chochem:
This is not saying this change shouldnt happen, but just to be clear:
This will not affect most people that update or migrate correctly as this config is saved and ported between instances by the defaultconfigmod (it is listed in /client/localconfig.txt). This PR will only affect new installs or people that migrate incorrectly.

Case in point: I made a new instance with this pr, migrated over, and after running the config is right back to 2.5

wlhlm:
With https://github.com/GTNewHorizons/GT5-Unofficial/pull/4994 implemented, I'm fine with this.

UltraProdigy (approved):
This has been on the table for a month. There are no remaining issues or objections, nor any documented issues on experimental. Looks like a solid change.

PR #21555, opened 2025-09-28
Fix "UV Lootbag has some strange items"

### LINKED ISSUE #21554: UV Lootbag has some strange items

_No response_

### Your Pack Version

2.8.0

### Your Server

SP

### Java Version

Java 8

### Type of Server

None

### Your Expectation

Open a UV lootbag

### The Reality

Image

Why there's a "drop info" in my item???
the thing i get has a really strange nbt that cannot use unnamer to remove nbt.

### Your Proposal

Fix this.
Seems like xml file is wrong.

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [x] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

SST-03:
due to some unknown reasons, some items in uv lootbags have some items with strange nbt. this patch removes them.
closes #21554


YannickMG:
This NBT is used in https://github.com/GTNewHorizons/EnhancedLootBags/blob/005bfd892ae879f7cbb2aedcbbd773c28183c2d5/src/main/java/eu/usrv/enhancedlootbags/core/LootGroupsHandler.java#L575 to communicate drop chances. Are you sure there's no visible impacts from removing it? If so, that feature may be broken.

SST-03:
> This NBT is used in https://github.com/GTNewHorizons/EnhancedLootBags/blob/005bfd892ae879f7cbb2aedcbbd773c28183c2d5/src/main/java/eu/usrv/enhancedlootbags/core/LootGroupsHandler.java#L575 to communicate drop chances. Are you sure there's no visible impacts from removing it? If so, that feature may be broken.

I think no. You see it's for fake recipe, and will be automaticly added.
This NBT **in xml file** is useless for this feature.

Other stacks doesn't have this nbt, but the feature is still working. Also i think that these nbts may from fake recipe of LuV Lootbag, seems live devs made a mistake.

Issue #21518, opened 2025-09-27
Et Futurum Requiem should focus on additions and avoid changes to existing vanilla behavior


User198263321:
### Your Pack Version

2.8.0

### Your Proposal

GT:NH is a pack that is running on Minecraft 1.7.10

However, there are assets that were changed that match modern versions of Minecraft and some (like me) do not prefer those changes. So I am going to list what I noticed when updating to 2.8.0. If this never gets changed, then it will be helpful for those who are looking to revert back to older assets. You can do this for mods by extracting the older mod's assets and making a texture pack out of it to apply to the game.

Beds now have colored variants (which I don't mind), but have added models and updated names for the legacy bed (now called red bed)

Doors (except for iron) and chests (only wooden) and wooden, gold, and iron pressure plates now use the new sounds. Could not find what mod changed it, and it is very frustrating because I don't know how to make a resource pack to replace sounds that are not supposed to exist in 1.7.10
**EDIT: Turns out I can change the chest and door sounds with a resource pack that is designed for a much newer version of the game. Don't know why that works so it must be defined in a json somewhere I cannot find. The pressure plate is still an issue though and is hard coded to be a different pitch or something.**

Iron chests mod's textures have changed that look like the minecraft texture update. They also conflict with the ironchestminecarts mod now. Not a big deal since I can just make a resource pack for it.

Critical hit sounds are present (can also be removed by making a resource pack with a silent audio file in its place (assets\minecraft\sounds\entity\player\attack\crit1 through 3)

Cow milking sounds now are added

The new cave noises have been added (Cave noise 14 that is in MC 1.9)

### Your Goal

To play this modpack without newer asset replacements being used.

### Your Vision

Modpacks that are on older versions of the game should avoid using newer vanilla assets like sounds and textures in replacement to existing blocks that are already present in the Minecraft version.

[...]

PR #21705, opened 2025-10-02
Update etfuturum config to keep parity of original 1.7.10 vanilla and modded items, blocks, and assets, but keep modern additions.

User198263321:
Keep the general spirit of 1.7.10 sounds. The ones I can see of a good argument of keeping is end portal sounds so the whole server can hear, but this is not normally present on 1.7.10 and maybe nether ambience with the new blocks.

### DIFF
--- a/config/etfuturum/functions.cfg
+++ b/config/etfuturum/functions.cfg
@@ -65,13 +65,13 @@ changes {
 
 client {
     # Bows render pulling animation in inventory [default: true]
-    B:enableBowRendering=true
+    B:enableBowRendering=false
 
     # Enables NBT tag count and item namespace label on F3 + H debug item labels [default: true]
     B:enableExtraF3HTooltips=false
 
     # Skulls render 3D in inventory [default: true]
-    B:enableFancySkulls=true
+    B:enableFancySkulls=false
 
     # Enable the /fill command. [default: true]
     B:enableFillCommand=false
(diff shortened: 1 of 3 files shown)


Dream-Master:
No why you disable every sound ? This shouldn't be the pack standard .

User198263321:
> No why you disable every sound ? This shouldn't be the pack standard .

Should we at least disable the changed sounds for doors and chests/pressure plates? Someone else in #21518 had issues with muted audio

Dream-Master:
I like the new stuff

User198263321:
> I like the new stuff

I reverted many of my changes, but only changed door, button, pressure plate and combat sounds. The new combat sounds I think don't fit the old combat system before 1.9 and the new door sounds do not effect iron doors or any other variant besides vanilla wood it seems like. Also, lever sounds were unchanged before the commit, as well as many others such as the new chest noises not affecting the iron chest mod, so after the commit, the sounds are more consistent in the game with modded counterparts.

User198263321:
I made more changes such as not rendering mob heads in the inventory in 3D since that would break parity with wolf heads in witchery, but enderio uses fancy heads as well. The config setting seems broken, so it currently doesn't do anything to the mob heads anyways.

Reverted new block textures since it conflicts with every modded item and block that uses it in their texture needing the material.

Beds and beds that are dyed are 2D in the inventory again, and the dyed beds show up properly in 2D.

Dream-Master (changes requested):
i will not approve this pr so do the changes to your game but not to GTNH standard

UltraProdigy:
I'm going to go ahead and close this on account of the fact that it appears to have been vetoed. If this is wrong feel free to re-open.

Issue #21408, opened 2025-09-23
Machine connected to a cable in an already connected state doesn't receive power


feioO:
### Your GTNH Discord Username

Feio0

### Your Pack Version

2.8-rc1

### Your Server

Private server

### Java Version

Java 21

### Type of Server

Vanilla Forge

### Your Expectation

A machine connected in this fashion to receive power properly right away.

### The Reality

Trying to connect a machine or energy hatch to a cable in an already connected state will not receive any power. I've tested waiting for 5 minutes and still no power.

<img width="620" height="372" alt="Image" src="https://github.com/user-attachments/assets/404d9011-5b96-41e5-a0c1-a1fccb7594dc" />

<img width="1294" height="456" alt="Image" src="https://github.com/user-attachments/assets/3663d0b3-01c0-4c0c-b439-3dce07c6491d" />

The machine begins receiving power if you disconnect and reconnect any cable on the energy network.

<img width="1270" height="393" alt="Image" src="https://github.com/user-attachments/assets/960d9628-0232-462b-9fb2-5fa20c622e05" />

I've not be able to figure it out and reproduce it, but sometimes the machine also starts receiving energy if there is a change to the power consumption on the network, like another machine starting or stopping. 

### Your Proposal

Fix the issue so that machine receive power normally.
While talking with people someone mentioned that this change might be related, but we didn't check into the code itself to confirm.
https://github.com/GTNewHorizons/GT5-Unofficial/pull/4865

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [x] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

boubou19:
@C-Remilian any idea?

couref:
I'll take a look at it.

boubou19:
put this speed up feature behind a flag for now: https://github.com/GTNewHorizons/GT5-Unofficial/pull/4919

Issue #4921, opened 2019-08-10
Floodlights occasionally cause server to crash


AlexBucket:
#### Which modpack version are you using?
2.0.7.5

#
#### If in multiplayer; On which server does this happen? 
ProsperCraft

#
#### What did you try to do, and what did you expect to happen?
Loaded an area with floodlights

#
#### What happened instead? (Attach screenshots if needed)
Server crashed

#
#### What do you suggest instead/what changes do you propose?
Fix floodlights

https://paste.fedoraproject.org/paste/xt~xtw9ZeULFJxeNWk0xdw

richardhendricks:
@AlexBucket Any details on the crash?

1. Consistent?  Every time those chunks are loaded?

2. Anything in particular about the floodlights setup that might trigger a crash? Light-beams crossing, etc? 

3. How are the floodlights configured? Cone or straight?

Floodlights mod is open source. https://github.com/Keridos/FloodLights/tree/1.7.10

https://www.baeldung.com/java-concurrentmodificationexception

    public void updateAllSources() {
        for (int[] source : sources) {
            TileEntity te = worldObj.getTileEntity(source[0], source[1], source[2]);
            if (te != null && te instanceof TileEntityMetaFloodlight) {
                ((TileEntityMetaFloodlight) te).toggleUpdateRun();
            } else {
                this.removeSource(source[0], source[1], source[2]);    <-- Looks like the problem!
            }
        }
    }

richardhendricks:
uh.... that code is some "weird shit"

draknyte1:
You shouldn’t update something while iterating. 
Just gotta use a concurrent data type. (Concurrent hashset of int[][][])

finkledoodle:
This is my base where the crash was happening.
1. No it doesn't seem to be consistent, sometimes it crashes when I first load the area, other times I could be returning to that chunk an hour later (still loaded with chunkloader), and then it would crash.
2. Yes, I have a series of floodlights on the ceiling, as well as a set right above the fluid level to light the surface.
3. The lights on the ceiling are cone, those on the surface are straight
Row of lights at surface level, and then in the ceiling above
![2019-08-14_13 55 58](https://user-images.githubusercontent.com/12854704/63055692-52cebe80-be9b-11e9-8ed8-d70b975f6e6e.png)

That Ztyl block was one of the PhantomLight issue blocks, the other was in the same configuration.
[...]

Issue #4919, opened 2019-08-10
Mutator Quest Inconsitant With Mutator Inputs


Masbee29:
#### Which modpack version are you using?
2.0.7.5
#### If in multiplayer; On which server does this happen? 
N/A
#
#### What did you try to do, and what did you expect to happen?
Nether stars as Mutator in alveary.
#
#### What happened instead? (Attach screenshots if needed)
Nothing. Nether stars aren't valid inputs anymore
#
#### What do you suggest instead/what changes do you propose?
Update questbook to not include nether star as a possible modifier.

Issue #4865, opened 2019-07-21
Quest Correction?


shawnbyday:
#### Which modpack version are you using?
2.0.7.5
#
#### If in multiplayer; On which server does this happen? 
Private Server
#
#### What did you try to do, and what did you expect to happen?
Made an iron hammer, excavator and lumbering axe. Expected quests to not complete since they say "Steel".
#
#### What happened instead? (Attach screenshots if needed)
Quests completed as if I had made steel tools.
#
#### What do you suggest instead/what changes do you propose?
Change the quest book to say iron or change the quest to require steel.

Dream-Master:
It was changed to ignore nbt data but i guess this dont work propertly

Dream-Master:
I think make this tool in Iron isnt realy good since you have steel at this tier to mak more durable tools.

Dream-Master:
In sp it works
![image](https://user-images.githubusercontent.com/2970688/61993703-cedd9f80-b06f-11e9-894c-045c28446adc.png)
![image](https://user-images.githubusercontent.com/2970688/61993706-d4d38080-b06f-11e9-94ef-b2aee646351e.png)
Maybe i change the quest text

PR #21725, opened 2025-10-03
Set "speedupMachineUpdateThread" to true

C-Remilian:
Changes from [pull #4865](https://github.com/GTNewHorizons/GT5-Unofficial/pull/4865) were put under this flag in [#4919](https://github.com/GTNewHorizons/GT5-Unofficial/pull/4919) due to [issue #21408](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/21408) which was presumably fixed in [pull #4921](https://github.com/GTNewHorizons/GT5-Unofficial/pull/4921). Needs testing.


Dream-Master:
we test this on zeta first

config/GregTech/GregTech.cfg
```diff
@@ -71,6 +71,9 @@ gregtech {
         # Controls the stacksize of oredicted planks. [range: -2147483648 ~ 2147483647, default: 64]
         I:maxPlankStackSize=64
 
+        # Speed up machine update thread by skipping cables. Warning: this can cause weird side effects. [default: false]
```
UltraProdigy, line 74:
```suggestion
        # Speed up machine update thread by skipping cables. Warning: this can cause weird side effects. [default: true]
```

UltraProdigy (commented):
Not sure whether this test has been going well so I won't review but either way here's a nit.

config/GregTech/GregTech.cfg
couref: The commented line is auto generated based on [GT5 definition](https://github.com/GTNewHorizons/GT5-Unofficial/blob/55c51c5afe4da92f8382669cc998bdaa9a8aa8c9/src/main/java/gregtech/common/config/Gregtech.java#L133). I'd say that's where it should be changed first.

Dream-Master: i guess we can change this after test to true in code

PR #21733, opened 2025-10-04
remove incorrect shift-clicking dissuasion

metalcupcake5:
the questbook tells the reader that they are unable to shift-click into the lead lined box, which I am assuming is from an older version where this wasn't possible


UltraProdigy (approved):

UltraProdigy:
Ty!

PR #21915, opened 2025-10-11
Bee Meteor adjustments (Add missing hives and remove marble one)

Miklebe:
image

My OCD compelled me to clean up the ordering in config.

javaw_fPalQlfBtY

Adjusted a few hives based on their coins cost and reduced pam's hives chance (not by much, but they are practically useless and can't be used anywhere with Pam's Apiary and Queen recipes disabled).

Removed Marble Hive due to not having Marble Bees to drop from them.

A few after and before screenshots: https://discord.com/channels/181078474394566657/522098956491030558/1426526725801775247

### DIFF
--- a/config/BloodMagic/meteors/BeeHive.json
+++ b/config/BloodMagic/meteors/BeeHive.json
@@ -1,5 +1,7 @@
 {
   "ores": [
+    "etfuturum:honey_block:0:100",
+    "BiomesOPlenty:hive:1:95",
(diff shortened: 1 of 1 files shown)


Miklebe:
Another strange thing is that the 4 of these hives (last 3 from Magic Bees and Ender Hive from Forestry) are completely missing from NEI, but i couldn't find them in the hide config in neither this, nor NEI repos.
image

PR #21930, opened 2025-10-11
Fix The Forest dungeon

WanderingHero:
Forest dungeon was bugged and didn't spawn correctly, this fixes it.


WanderingHero:
This problem only existed since the minor loot change, not since the start of 2.8

UltraProdigy (approved):
Makes sense.

PR #21956, opened 2025-10-13
Алюмит -> Обзинит for TC tool/part

Eldrinn-Elantey:


Nikolay-Sitnikov:
Does this work in-game?

PR #22039, opened 2025-10-16
Add Cooking for Blockheads drop-off support (yummers)

Miklebe:
I think it would be lovely to just get back to your kitchen and drop off all the crops you harvested for cleaner management (if anyone even tries that)

image

Because of substring matching it tries to interact with Nuclear Control's counters and Open File Cabinet, but first two only accept IC2 Transformer upgrades and the last is not craftable.
Some screenshots from testing:

<details>
  <summary>Big images</summary>
  
javaw_3pmffHVzk1
javaw_2XjJ25n4d6
  
</details>

### DIFF
--- a/config/bogosorter.cfg
+++ b/config/bogosorter.cfg
@@ -59,6 +59,9 @@ general {
             Barrel
             Drawer
             Crate
+            Fridge
+            Cabinet
+            Counter
          >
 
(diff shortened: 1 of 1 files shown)


AshVz:
Thank you, very useful for me!

PR #22214, opened 2025-10-24
Clarified cleanroom quest regarding power usage

tyrannus00:
The wording regarding the power usage for the cleanroom is confusing and easily misleading. It makes it sound like the cleanroom is the exception for accepting 40eu/t (1 MV amp) through an LV energy hatch, while the exception actually is referring to the energy hatch being able to draw two amps for the cleanroom, which is usually not the case.

The diff for this change is unfortunately unreadable, so here is the changed content (**bold**) with the necessary context:

Previous:
The power usage of this multiblock is a bit unusual. It starts out at 40 EU/t and then gradually goes down. But even when it uses 40 EU/t - which is usually MV - a single LV energy hatch suffices, **as long as you give it enough power**. Just keep in mind that this makes the Cleanroom an exception, not the rule.

Proposed change:
The power usage of this multiblock is a bit unusual. It starts out at 40 EU/t and then gradually goes down. But even when it uses 40 EU/t - which is usually MV - a single LV energy hatch suffices, **because it accepts two amps of LV power**. Just keep in mind that this makes the Cleanroom an exception, not the rule.


chochem (approved):
this sounds pretty similar to me, but sure why not.

PR #22275, opened 2025-10-27
Add quest for the chaos locator

GDCloudstrike:
Quest addition for https://github.com/GTNewHorizons/GT5-Unofficial/pull/5172

image

image


UltraProdigy:
Following this I might add mention of this to the chaos guardian killing quest (which I believe exists?).

PR #22301, opened 2025-10-29
Quest Patches 10.29 (2.8.X)

UltraProdigy:
### Changes:
- Actual beacon retrieval fix
- Steel barrel fix
- Cut down Iron banded wand material amount to 1 wand
- Gave info on swapping wand cores & caps
- Fix typo

Issue #22105, opened 2025-10-20
Monthly Ticket for Quest Changes (Nov 2025)


UltraProdigy:
## Overview:

The running ticket for quest changes; a ticket for any problems with any quests. Just post below and I will get to your comment as soon as I can. For more in-depth suggestions or problems, creating a [Dedicated Ticket](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues) is better. This ticket is more focused on small and clear problems and suggestions that don't really need any larger discussion.

Unaddressed items from last ticket (https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/21341) can be found below.

#### To See Latest Questbook:
1) Be sure your GTNH instance is reasonably up to date
2) Grab the most up-to-date quest files from [Here For 2.9+](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/tree/master/config/betterquesting/DefaultQuests) or [Here For 2.8.X](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/tree/release/2.8.x/config/betterquesting/DefaultQuests) and download by either pasting the URL in [here](https://download-directory.github.io/) or forking the pack and grabbing the folder from the appropriate branch.
3) Rename the folder "DefaultQuests" if needed and replace the folder in your `.minecraft\config\betterquesting`
4) Either reload your world, or if you hot-swapped the folder while in game, use `/bq_admin default load` after new folder is in (this will **NOT** impact your quest progression).

#### 2.8.2+
- [x] Prismatic Naq. Quest advice on strontium [22191](https://www.github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/22191)
- [x] Separate Ore coins quests (2.8.X ONLY) [22191](https://www.github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/22191)
- [x] Crop Automation PreReq Fix [22191](https://www.github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/22191)
- [x] Round and Round and Round Checkbox Fix [22191](https://www.github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/22191)
- [x] Add LuV assembler dependency for LuV superconductors [22191](https://www.github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/22191)
[...]

PR #22310, opened 2025-10-29
Quest Patches 10.29 (2.9.X)

UltraProdigy:
### Changes:
- Actual beacon retrieval fix
- Cut down Iron banded wand material amount to 1 wand
- Gave info on swapping wand cores & caps
- Fix Quartzite typo
- Removes mention of slicer for circuit imprints

This is a cherry pick of https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/pull/22301 for 2.8.X but also with a couple 2.9 specific changes, and not with the steel barrel req fix, as that was already in.

Contributes towards https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/22105

### DIFF
--- a/config/betterquesting/DefaultQuests/Quests/GettingAround-bAal8kZ3R8eRTwaHgbYkKA==/ElectricJetpack-AAAAAAAAAAAAAAAAAAACiw==.json
+++ b/config/betterquesting/DefaultQuests/Quests/GettingAround-bAal8kZ3R8eRTwaHgbYkKA==/ElectricJetpack-AAAAAAAAAAAAAAAAAAACiw==.json
@@ -158,3 +158,4 @@
       "taskID:8": "bq_standard:retrieval"
     }
   }
+}


chochem:
I think we might need to ask dream what the intention is with industrial diamonds

UltraProdigy:
Last time I saw them discussed I believe the intention is to make them obsolete with no intended replacement though I'm no authority. Regardless though imo this should still pass for now until something changes (since we devs take so long to decide on things :trollface: )

@Dream-Master Would love some thoughts though. (The context is that currently they are craftable in a regular compressor and not restricted to an implosion compressor hence the dependency removal on the electric jetpack quest.

chochem:
well dream at least used to call the implosion compressor his favourite multi, so removing one of its important uses seems strange.

UltraProdigy:
Tbf I agree, though it was removed 3 years ago I believe you were saying in the ticket? If you are suggesting it gets re-implemented I can take the change out of this PR for now.

Dream-Master:
@UltraProdigy @chochem where is the industrial Diamond mentioned? i dont see it in this changes?

UltraProdigy:
This should be good to go; I'm going to include any changes that spur from removing the regular compression recipe in another PR that can be cherried to 2.8

chochem (approved):

PR #22364, opened 2025-11-03
Add Quests For Vending Machine

UltraProdigy:
### Changes:
- Adds quest for Vending Machine to Stone age tab (location of quest is not permanent, I will move it when I do the coke oven quests)
- Adds quest for Vending Machine uplink hatch to AE tab
- Deletes short P2P guide quest (all duplicate info that can be found in adjacent quests)

### Vending Machine Quest: 
image

### Location: (Again, Temporary)
image

### Uplink Hatch Quest:
image

### Location:
image


Dream-Master:
@UltraProdigy thanks for add the quest but I think it is too late. Some trades like seeds need to be done after you got the bed. I think what I can change .

cubefury:
I think it's cuz people need to know how to craft tools before it, technically they can craft it alot earlier

UltraProdigy:
I see what you mean, my bad. I'll switch it over if we change things up.

PR #22377, opened 2025-11-04
Netherrack Froth Count Modification

UltraProdigy:
### Changes:
- Changes quest to reflect the output change 1 -> 8 Netherrack Froth found in https://github.com/GTNewHorizons/GT5-Unofficial/pull/5246

Doing this change standalone just in case it gets ported to 2.8 so it's an easy cherry.
(Also, I'm an S tier PR farmer how can I resist)


chrombread (approved):
a masterclass

PR #22408, opened 2025-11-07
Don't advertise curseforge on tips screen.

tomprince:


Dream-Master:
This is a joke. I guess people lost the humor a bit. Are gtnh a serious place ? @tomprince

tomprince:
> This is a joke. I guess people lost the humor a bit. Are gtnh a serious place ? @tomprince 

I can't tell what the joke is supposed to be. If I were a new player, went to the download page and installed the pack, then started up the game and saw this tip, I'd be inclined to think that I should have instead installed from curseforge. Given that we push people away from using curseforge, it seems counter productive to suggest using curseforge in the tips (particularly with no context clues to indicate that it is a joke.

FourIsTheNumber:
The joke is just that curseforge spams that message at you whenever you download mods from them.

PR #22434, opened 2025-11-08
Add back Cosmic Neutronium, Infinity Catalyst and Infinity bees to the mutatron, but with their requirements

boubou19:
This PR adds back the Cosmic Neutronium, Infinity Catalyst and Infinity bees to the mutatron, but with their requirements.

The mutation rates are the following:
- Cosmic Neutronium: 0.7%
- Infinity Catalyst: 0.3%
- Infinity: 0.1%

Those are by far the worse bees to breed in the game only because they are the only one excluded in the mutatron, which means you need to use breeding methods unlocked earlier than LuV/ZPM. This causes a severe issue in the bee breeding progression, as not only you need to go back previous breeding methods, but infinity catalyst is actually required in other bees, like kevlar bee, and those bees can be bred just like normal in mutatron.

In addition of that, if you want to breed those 3 bees, you need to have the bee requirements, which are the following:
- Cosmic Neutronium: Cosmic Neutronium block
- Infinity Catalyst: Infinity block
- Infinity: Infinity Chest

So to get those bees, you must actually get those resources first anyways, having those bees being blacklisted from mutatron just adds a time gate based on player's luck, because once you have a queen, you use the Mega Apiary in swarm mode to get princesses to then extract the species gene and then get a 13/13 gendustry template to litterally print bees to scale.


UltraProdigy (approved):
Solid change; technology shouldn't regress & changes no balance. As it stands this doesn't enrich the game just adds 20 minutes to your playtime.

Though I will say it would be cooler if Mutatron was a multiblock.

PR #22456, opened 2025-11-10
Alter Coke Oven Quests For New GT Coke Oven

UltraProdigy:
### Changes:
- Changes initial coke oven quest to use GT & alters text
- Changes 10 coke oven quest to use GT & alters text
- Adds quest for coke oven hatch
- Alters various quests mentioning the coke oven if the information was incompatible. Hopefully we can do this for the Advanced Coke Oven too in the future.

### "Getting (Char)coal" Update:
image

### Added Coke Oven Hatch Quest (Right after bronze pipe quest)
image

### "Why One Coke Oven When You Can Have Ten?" Update:
image


chochem (approved):
I am just gonna trust that calculation for wallsharing 10 xD

we might need to move the hatch quest if dream changes the recipe, but that is easy enough to do then. All good for now.

PR #22466, opened 2025-11-10
Re enable stone tools

captainjolt:
Re-enabled stone tools to accompany [Added Recipe for Stone Tools](https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1461#issue-3609155674)

Note: Sword, axe, and hoe were already enabled


UltraProdigy:
What's the point of this? I thought these were meant exclusively as crafting ingredients?

captainjolt:
> What's the point of this? I thought these were meant exclusively as crafting ingredients?

That was my intention, but DreamMasterXXL mentioned that I should enable them to be usable again since they're worse than the other tools you will have access to at the point you can make them anyway. 

I don't really mind if they are enabled or not, but if we do leave them disabled I feel that we should disable all of them instead of just the pickaxe and shovel.

UltraProdigy:
I see. I suppose there's no harm either way.

(Assuming Players can't get this for loot?)

Dream-Master (approved):

PR #22720, opened 2025-12-02
Quest Patches 12.2.25

UltraProdigy:
### Changes:
- Andradite typo in "Even More Challenges Needed?"
- Make note about Soldering Iron R-Click Ability
- HV Chemical Reactor Quest Description Improvement

#### HV Chem Change:
image


chochem:
I think that is a good quest. Only concern I have is that you can get 2 circuits for the price of pressing a button.

boubou19:
Sometimes the game could be nice to the player :joy: But no strong opinion if the circuits "for free" are removed

UltraProdigy:
Yeah.. it's a bit odd despite being the "same task" in effect, I'll have this updated in a few minutes with reduced rewards.

Dream-Master:
I hate these check box quests. There's always a better alternative.

UltraProdigy:
Given the point where PTFE is able to be made it's a bit tricky. The only reasonable alternative I can think of (that doesn't change recipes) would be to keep it retrieval for the HV Reactor & have a skip quest for the LCR. Either that or a large scale reshuffle.

chochem (approved):

PR #22794, opened 2025-12-09
Coke Oven Quest Fixes

UltraProdigy:
### Changes:
- Fixes amount in 10x Coke Oven quest (turns out I was wrong)
- Fixes amount, description & item in "Faster Charcoal"
- Makes "Benzene Tier" and "Biodiesel Alternative" OR unlock with MV again


chochem (approved):
kind of weird to make 19 when they are wallshared but I guess that is the math for the rc boilers.

UltraProdigy:
Yeah I had the same thought

PR #22851, opened 2025-12-13
Quest Patches 12.13

UltraProdigy:
### Changes:
- Switches up the reqs. on solar panel quests to account for Solar Factory
- Removes outdated "new QFT" line from "Kevlar"
- Removes incorrect "vertical" distinction from Fancy Stripes
- Corrects "computer base" to "Computer Case"


chochem (approved):
the solar quests will need even more adjustments then this, but this is fine for now. Ill get it to it when I actually make them in my current run :P

PR #22856, opened 2025-12-13
add compressed chest to shulker blacklist

chochem:
backpack/dolly/etc are all there, but this one was missed.

(was tested)


UltraProdigy:
Might want to also include the Draconic chest if that works

chochem:
will do

PR #22877, opened 2025-12-15
Quest text formatting edits 3

kamenskyi:
Multiple edits to special formatting characters in quest texts


UltraProdigy (changes requested):
"eXpert" was intentional, that can be restored

kamenskyi:
There are 17 more quests where %n is at the end. Should we include them in this pull request too?

UltraProdigy:
> There are 17 more quests where %n is at the end. Should we include them in this pull request too?

\n you mean? You can absolutely include that; I'll review it after.

UltraProdigy (approved):
Elite, TY for the PR!

PR #22979, opened 2025-12-22
Quest Patches 12.22

UltraProdigy:
### Changes:
- Adds all types of maglev to quest to allow for completion with any tier and gets rid of standalone quests with other versions that were only visible in 1/3 tabs, misleading.
- Adds helpful information to "Magnets != Magic"
- Adds detail to "Refined Firestone"


wlhlm:
> [...] removes unneeded retrieval from "Refined Firestone"

@UltraProdigy I think you forgot that part.

PR #23049, opened 2025-12-28
Quest Patches 12.28

UltraProdigy:
### Changes:
- Removes Coins, Coins, Coins reference from "Options For Finding Zinc"
- Changes "Secrets" from mandatory quest to kill quest
- Slightly reorganizes Adept Thaum tab to compensate for that ^ change
- Fixes Reqs for Compressed Chest
- Removes "Palladium" req from 2-Eth... Quest
- Removes Fluid Level Emitter Quest
- Amends Level Emitter Quest to account for fluid interaction
- Removes Super Bus mention from LMA Quest
- Adds Quest for Thermic Heating Device in Mass Processing

### Thermic Heating Device Quest:
image

### DIFF
--- a/config/Betterloadingscreen/tips/en_US.txt
+++ b/config/Betterloadingscreen/tips/en_US.txt
@@ -12,7 +12,7 @@ Material tiers play a role when breaking pipes
 If a machine catches fire, it can explode
 Adding water to an empty but hot Boiler will make a loud bang!
 Avoid Eldritch Obelisks!
-You can bind the questbook to a key to save you from using the item
+You can bind the questbook to a key to save you from using a slot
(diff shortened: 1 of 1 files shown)


chochem:
why not unify the two zinc quests?

Why are you removing palladium from the 2eth quest? Was the recipe changed?

UltraProdigy:
Right you mentioned there were two my mistake. I'll have to unify it when I'm back. Maybe in a separate patch or I can just set this to draft for now.

As for Palladium, I've had a few complaints from people getting palladium before unlocking the quest. Considering the overarching skip completes the quest anyways & the quest mentions needing palladium I figure it's no harm to just remove the req.

chochem (approved):

PR #23070, opened 2025-12-30
More meteors

EnderProyects:
- Buffed or nerfed tnt meteor to add more explosives, reduced redstone filler
image

- Meteors should fall right? Well, sand falls at least, sand + hempcrate, nothing could go wrong,.. right?
image

- Who doesnt like some vanilla ores for some astetics, or well, lets just say to decorate
image

- Do you have a problem with taint? Or do you need some vinteum? Wait no more, just spawn a taint biome in the roof of your base
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,7 @@ config/WitcheryExtras
 !config/GregTech/Unification.cfg
 !config/NEI/defaultbookmarkcontainers.cfg
 !config/NEI/collapsibleitems.cfg
+!config/NEI/hiddenhandlers.cfg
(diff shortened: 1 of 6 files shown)


Spicierspace153:
Maybe change the starfield to the orechid?

EnderProyects:
That doest work, I tried, botania flowers are orediced by name, not by tag number, so it just places a poppy 🙃🤣

EnderProyects:
> Radius 22 might be a little wild but otherwise I see no issues with this assuming officials don't have to worry about griefing in the meteor department

changed it to 16 xd

PR #23124, opened 2026-01-04
prevent adventure backpacks overlapping effects

mazunki:
If adventure backpack is equipped, tanks will overlap with effects on the bottom right corner of the screen. This moves them to the left so the effect times are visible. Seems to work fine across GUI scales.


UltraProdigy:
Can you include before and after images?

PR #23145, opened 2026-01-06
a quest for the Certus Quartz Tank

danyadev:
image

image


UltraProdigy:
I like it but it needs rewards

Also does there need to be a dash between cool-looking?

danyadev:
What about giving 4 tanks in return?

image

UltraProdigy:
Generous but it's fine by me

danyadev:
Similar two quests for gt fluid tanks give 4 large steel pipes each, which is much more valuable time-wise (if you smelt them in TC smeltery to get valuable steel back, for example) than these 4 cheap tanks, so seems fair for me

PR #23293, opened 2026-01-20
Additional JourneyMap Icons

Graymare:
• Adds 20 missing JourneyMap icons for various mobs.

• Changes the default steve.png to a Wisp icon, as (for reasons only known to Thaumaturges) Thaumcraft uses that texture call for Wisps.

• Icons have been tested locally in-game.


UltraProdigy (approved):
Yay! Thank you

Issue #23297, opened 2026-01-20
Useless item in dungeon loot table


ERRRR0R:
### Your Pack Version

2.8.4

### Your Server

SP

### Java Version

Java 25

### Type of Server

None

### Your Expectation

A useful item

### The Reality

Tin Gear Bushing from Railcraft is completely useless but it appeared in a Witch Circle (Seed -3350821674870357872 /tp 103 69 -585 but the witchery structure is random and rare so it takes a lot of tries)

<img width="431" height="168" alt="Image" src="https://github.com/user-attachments/assets/9e47774c-41e9-46d8-b5d9-c8349c3df8b0" />

### Your Proposal

Remove Tin Gear Bushing from dungeon loot

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

PR #23299, opened 2026-01-21
Remove Useless Loot

UltraProdigy:
Removes "Tin Gear Bushing" from loot table, it has no purpose in this modpack. If it gets one later we can add it back. 

Close https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/23297


Dream-Master (approved):

Dream-Master:
Maybe we can add useful loot instead ?

PR #23325, opened 2026-01-23
Replace Music Files with Dependency on NewHorizonsCoreMod

Nikolay-Sitnikov:
Also removes some kind of strange leftover file for the questbook.

Depends on https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1575.


Nikolay-Sitnikov:
This will require updating the wiki to correct the "disabling main menu screen music" steps.

It also *might* brick all other vanilla sounds, but I don't think it will. I can't test right now, since the daily is bricked.

PR #23384, opened 2026-01-28
Update configs to match new code changes

Nikolay-Sitnikov:
Requires https://github.com/GTNewHorizons/NewHorizonsCoreMod/pull/1581 to be merged.

(Also makes it possible to throw Endest Pearls - although they do bypass chunk protection, I disabled their ability to break tile entities, so hopefully the damage won't be horrible if anyone gets griefed by it. I'll look into adding chunk protection support later.)

Removes unused config entries and moves some other configs to their correct place.


danyadev (approved):
(wait until the main pr is merged)

PR #23447, opened 2026-02-05
Add salt bee quest

rockobonaparte:
This bee was useful for me when metal combs started to require hydrochloric acid and I wanted a fast solution. While watching some YouTube videos about bee breeding, I saw some complaints about how this bee was not mentioned in the quest book, so I decided to add it.

Putting it with its clay and aluminium dependencies would disrupt the layout of the entire left side of the quest book, so it was placed off to the right, near the line with apatite, with hover-over dependencies.


rockobonaparte:
I should note there was an odd problem when I tried to do a full copy of the quests folders. Just about everything ended up changing. So it sounds like something strange happened or I messed up in the transfers from the git repo's quest book to/from the game editor instance. I believe the two relevant files are the questline file and the quest file so I built the commit around that directly.

rockobonaparte:
Ah okay I figured there were some "hidden rules." I added the block of salt specifically because I had a forehead-slapping moment when I couldn't find "salt block" originally. I'll remove the drone and keep the princess per most of the quests. Still, I had a question about that. The earlier parts of the bee line do have the princess and drone, but then it stops. Is there a convention for that?

UltraProdigy:
The earlier ones show a drone so their importance is clearer to someone starting bees, but they quickly become unnecessary to include as their usefulness is more obvious to someone who's done a few breeding cycles. As for the block, it's not that I take an issue with it, it's more so that it's inconsistent; it should be all or none.

PR #23551, opened 2026-02-14
Easy Way to Change Biomes now uses Biome Page

rockobonaparte:
The retrieval task previously listed the Book of Biomes instead of the page. The recipe calls for a biome note and not the book itself.

This had personally confused me for a second while I was trying to figure out how to make the potion.


rockobonaparte:
I was intending to rearrange quests in that chapter so a neophyte like me that stumbled into it from the bee quests could figure out what the progression was going to be, but I isolated this change since it's a independent--and presumably uncontroversial--fix.

PR #23559, opened 2026-02-14
Witchery biome quests now tie into an untangled progression

rockobonaparte:
There's a lot of "the perfect is the enemy of the good" energy here. A lot more stuff can be redone. No quest test was altered to do this, but requirements were adjusted.

Changes:

1. The Easy Way to Change Biomes now has a hover-over requirement on Diamond Vapor. It's an ingredient in the quest and connects the quest to the rest of the Witchery progression.
2. The Hard Way to Change Biomes now has a hover-over requirement on Circle Magic since it requires performing a Rite of Summoning. I wasn't sure if there was a better existing quest to gate, but it at least gives a path in the Witchery progression.
3. "Fire Burn and ..." (the first quest) will unlock if the player completes Changing Biomes With Witchery. This is a short circuit, but let us assume they have the necessary level of magic and technology to get started if they managed to get into the Witchery quests through the back door.
4. Heavily rearranged all the quests on the screen. Most requirements have been untangled. I expect that the hard biome progression is still very bizarre and it still has a criss-cross.

Somebody who now reaches this chapter from the bee progression should see the Changing Biomes with Witchery quest. When completed, the first witchery quest will become active (if it isn't already). If they hover over the easy and hard progression, they'll get hidden lines to their dependencies. They can then trace how they have to get there from the first quest. The easy path has a semi-logical progression here. Just about everything along its route will become necessary.


UltraProdigy (commented):
The changes are solid, my one gripe is with the organization. I think the almost spirally webby way the tab was laid out is a unique staple of it. Not that it shouldn't have an organized aspect to it, the current one is pretty chaotic, but I think the straight lines take away from the character a bit. Also, feel free to take advantage of icon size if needed.

Also, you can lower your grid snap to differentiate placement a bit to make it look more 'organic' if that would be helpful.

Yours:
image

Original:
image

rockobonaparte:
I don't know what to do between that. My general objective was to show how to progress on biome changing if somebody came to this chapter cold from bee stuff. As it stands:

1. There is a progression in the quests to them. The easy method (the potion of shifting seasons) is just a list of ingredients. I took some note of what I had to do to get there, found the progression-wise closest pre-requisite, and linked it.
2. The distillery is not actually a pre-requisite of the spinning wheel, but that's how it is in the original progression.
3. Aesthetics-wise, I can make it more chaotic, but I don't know about icon resizing.

I was trying to avoid how people go into the help channel on the Discord and get that image macro for the shifting seasons potion when they ask for details. If the whole point is to just make the chapter rather inaccessible by theme, then yeah, this is a no-go.

Dream-Master:
You can simply make the items bigger in the edit menu. If you had done that instead of changing the shape, I wouldn't have had a problem with it. Don't get me wrong, but I think Witchery is such a special mod that it needs a special structure. You're welcome to set main quests that have a different symbol and make them a little bigger, but leave the shape as it is. Thank you.

PR #23601, opened 2026-02-18
Oblivion frame bee breeding meta notes

rockobonaparte:
An overhaul of the oblivion frame quest. I leaned on the meta for this quite a bit when breeding bees. It was some advice I sorted out from watching a few videos. So based on a personal of rule that "people shouldn't have to watch YouTube to figure out neat stuff" I took it to text.

1. The original oblivion frame quest was renamed. It retains its ID.
2. It was set as a pre-requisite on the apiary since its previous pre-requiste quest doesn't really gate keep it if you find the frames in Strongholds.


rockobonaparte:
This was a pretty rough one to sort out. My condolences to @UltraProdigy.

UltraProdigy (changes requested):
You accidentally added a duplicate of the quest file to the location folder, this is your current implementation:

Image

However, I do think that instead of changing the location of this one we should just make it a skip quest. I think it makes sense that it sits with all the other frames. The `Moran's Guide to Breeding Better Bees` quest in the Tips & Tricks section would be a good place for letting people know about the frame earlier.

rockobonaparte:
That's just busted. I'll have to figure out what got bitched before doing anything else. The quest wasn't supposed to start a space program and launch itself out the end.

rockobonaparte:
If there's still some screwball stuff going on, I think I'd just do a separate PR independent of this that's just doing the capitalization rename that betterquesting keeps doing on Linux. I had to filter through everything in the DefaultQuests folder that changed to pluck out the skip quest (since that went in a different directory and I didn't know where). I wiped my instance's quest book with this change set so it _should_ reflect what I meant.

I also rebased, and I should probably reiterate I did the skip quest and the other changes requested per Discord.

rockobonaparte:
lol nope. 

image

There's some stuff I can try:

1. My latest installation is the daily from February 15th so I'll bump that up on this machine and hope for the best/worst.
2. I got a Windows laptop up so I'll beam it over there. I'm assuming you're running on Windows.

Before I had updated it, I had verified it by doing a lot of manual copying back and forth between my game instance and my git checkout. However, I also just switched to blasting the installed betterquesting folder and symlinking directly to the folder in the repository. It's a lot cleaner and less likely to make screwups like the first attempt did. It also works that way here. I'm banking on it blowing up on the Windows laptop if it's going to blow up for me here anywhere.

UltraProdigy:
Hm okay, I'll re-try then just to make sure it wasn't some weird fluke on my end.

rockobonaparte:
I tried on a Windows install on my laptop here that loaded it okay, but the install there is older and the coins were invalid. Can you tell me exactly which build you tried? I want to stick with that on both my systems to try things.

UltraProdigy:
Okay, I did a complete reinstall and the quest is fine. Weird issue on my end. Let's put back the deleted quest and this should be ready to rock

rockobonaparte:
Ugh I saw the merge and thought the PR went in, not that it was getting updated with master, so I let it rot a second. The missing quest was a rename. The hash was preserved, but are renames a no-go too?

UltraProdigy:
It's gone from the quest tab

image

rockobonaparte:
It should be back from the dead. I think it got caught in a naming thing when I was still working on this branch without trying to push a pile of unrelated renames at the same time. I took care of that in a different, big, fat commit that I think amused you. ;)

PR #23707, opened 2026-03-01
Decrease Dream weaver buff Sleep % requirement from 100% to 25%

thenextelement:
Given that most servers set the sleepvote to 25%, leaving  this at a default of 100% makes crafting any dream weavers, aside from the nightmares one, on a server with any real player base almost entirely pointless.


boubou19:
yeah, sane defaults are always good, not like sever admins can't adjust it back to what it was anyways.

PR #23716, opened 2026-03-01
Dignifying the Old Quotes

UltraProdigy:
Farming as requested. Quest is staying though, idk why that had cusses to begin with


Iamworking12:
oh boy its harvesting season prodigy

Issue #37, opened 2015-02-06
Twilight Forest update soon


Dream-Master:
Twilight Forest get a update soon. The Mod Author is back. Fix that he said:

Crash when using magic beans on a server
Crash when using the enderbow on a server
(Maybe) Crash when using the sorting tree in some circumstances

Dream-Master:
Version TF 2.3.3 is out. i will update it soon

Dream-Master:
Updated

PR #24122, opened 2026-04-02
Enable mob.wither.spawn sound-effect when not near a wither-entity (#15967)

### LINKED ISSUE #15967: `mob.wither.spawn` sound effect does not play when mods try to play it

D-Cysteine

### Your Pack Version

2.5.1

### Your Server

SP

### Java Version

Java 20

### Type of Server

Single Player

### Your Expectation

Perform any action which should play the `mob.wither.spawn` sound effect (this one: https://minecraft.wiki/images/Wither_spawn.ogg), and the sound effect should play.

I tested and confirmed this with the Witchery Horn of the Hunt, as well as the Warp Theory fake wither warp event (this one has been fixed in dev).

### The Reality

The sound effect does not play (except when actually spawning the wither).

### Your Proposal

This one might be tricky to fix in a good way, so I'm mainly filing this issue just so we remember it, and so people will find this write-up if they notice this problem.

I think the cause of this issue is that Extra Utilities 2 will suppress any sound effect containing `mob.wither.spawn` from being played when there isn't actually a wither nearby. This was meant to prevent spawning withers annoying everyone on the server, but it also affects mods that happen to use that sound. Unfortunately, it might be tricky to fix: we either have to go modify every affected mod to use a custom copy of the sound that doesn't have `mob.wither.spawn` in the name, or maybe we can do a mixin to fix or work around the check in Extra Utilities 2.

Here's a search which tries to find affected mods:
 * https://github.com/search?q=org%3AGTNewHorizons+mob.wither.spawn&type=code

Here's the fix for Warp Theory, which uses a custom copy of the sound:
 * https://github.com/GTNewHorizons/WarpTheory/pull/37

### Final Checklist

- [X] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [X] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

WheelerShigley:
Enables the playing of the wither-spawn sound, even when not near a wither (namely for mods that happen to use this sound);
to do this, the ExtraUtities config, "disableWitherNoisesIfNotNearby," is set to false, as recommended by [Lyfts](https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/15967#issuecomment-2062038287).

As per my previous message on #15967, there isn't a [open source] mod in the pack that currently uses this sound because [WarpTheory](https://github.com/GTNewHorizons/WarpTheory) had a custom sound (which is identical) added to avoid this suppression, in #37.

---

If this pull-request is accepted, it may be worth undo-ing the changes made in  https://github.com/GTNewHorizons/WarpTheory/pull/37 since the additional feature would no-longer be needed; but this reversal would be equally an necessary as this issue.

When/if this pull-request is accepted/rejected, it will likely be able to close #15967.
The easiest thing to do, right now, would be to reject this change and close the issue, but this issue could come up again if a later mod uses the sound.


Dream-Master:
@WheelerShigley is this tested in full GTNH pack ?

UltraProdigy:
This would make it so everyone can hear whenever a wither spawns in right? I don't think we want that

WheelerShigley:
I did test it on the full pack; it was Daily 442, if I remember correctly.
On what UltraProdigy said: I'm not confident this is necessarily a desired change, so being rejected would be fine too; I just wanted to work towards closing some things.

UltraProdigy:
Yeah I think I would rather institute PRs like the one for Warptheory or take another avenue to solve the problem packwide, then disable this. Just shutting this off only works to re-enable a nuisance behavior without implementing a proper solution. 

In other words, I don't think the tradeoff of being able to hear wither noises from mods is worth the entire server being condemned to hearing whenever a player spawns in a wither regularly. 

It would be great if we can have the best of both worlds but in Lieu of that I would err towards status quo

UltraProdigy:
I'm going to go ahead and close this for the reason outlined above

PR #24250, opened 2026-04-13
Hv Chemical Reactor Quest Change

KraVoid:
I wrote about ability to hold 128 sheets in order to complete questline.
image


UltraProdigy (approved):
Yay! Thank you

PR #24252, opened 2026-04-13
Witchery BQ Changes

KraVoid:
Slightly reworked Witchery questlines:

1.Now you do not need to complete "wake up early" quest in order to unlock Spinning Wheel.

image

2.Now you can make Candelabra optionaly,beacuse you can make a required amount of power for some stuff without it.

image


UltraProdigy (approved):
Great change!

PR #24301, opened 2026-04-18
GTNH Power Planner Injects

DylanTaylor1:
Maybe I'm biased but I think it'd be better to point towards my Power Planner rather than the old "Large Turbine Calculator" or the Persus EOH Calculator.

The old Large Turbine Calculator has a few minor mistakes in its equations, has an unsortable and incomplete table of turbine materials, does not list fuel densities (EU/L), and is overall more confusing/difficult to use. Not to mention it doesn't have individual tabs for full setups.

The Persus EOH Calculator is not updated to 2.8 and therefore does not include the pity system, making it inaccurate. I believe the latest version on the GTNH Power Planner is also the easiest to understand since it lists values per success/failure instead of ideal/average. It also lists the different plasmas that it produces instead of only the amount of each.

https://docs.google.com/spreadsheets/d/1FTFdfmY_UWbTbFOyNzARKeI90pbHvPBH6M3JHJmaC-E/edit?usp=sharing


UltraProdigy (commented):
On all of these, I think it would be much preferred if the link was provided rather than alluded too. Ideally in a condensed url if possible (or moved to bottom of description) to not break up the text too much?

UltraProdigy (approved):

PR #24314, opened 2026-04-19
Fix missing Crops Lootbag Group

Dynamiczbee:
I didn't include the entire "batbuff" lootgroup when I recreated the Crops lootbags from the original IC2 Crops lootbags. Oops. This should re-add them.


Dynamiczbee:
Tested on my own, works as intended
image

Ruling-0 (approved):

PR #24337, opened 2026-04-20
Replace IC2 Rubber tree with GT variant

NeutronSelector:
Not merge without parent PR 
- https://github.com/GTNewHorizons/GT5-Unofficial/pull/6445


YannickMG:
Thanks for starting this effort! I believe this will need other PRs it at least the following repositories:
- https://github.com/GTNewHorizons/CropsNH to replace the IC2 crop
- https://github.com/GTNewHorizons/NewHorizonsCoreMod for recipes

And it may be beneficial to include compat in these repositories:
- https://github.com/GTNewHorizons/EnderIO farming
- https://github.com/GTNewHorizons/Binnie combs
- https://github.com/GTNewHorizons/Gravitation-Suite-Neo harvesting

YannickMG:
image

The regular pattern in-between blocks is pretty striking. It's not bad to be unique, but I don't think it's supposed to look less natural than vanilla trees right?

Also given that you're not meant to be able to tell the difference in-game, natural rubber logs should be named "Rubber Wood" instead of "Rubber Wood Natural".

I also notice that a treetap is still required for harvest but we're not including one in this PR. Should we either rethink the harvesting method or include a treetap? We could always do it as a separate PR of course but it could be neat to do it in one go.

LazyFlesh:
instead of having the spots be made at the moment of growth, can we mark logs with a knife so that resin bleeds from them an x amount of times? It removes a bit of rng, and makes the treetap (and a knife ig) far more useful than just growing then chopping down rubber trees

Dream-Master:
Yes more realistic rubber trees are a good idea. I remember ic2 had addon mod with a far more detailed rubber harvesting and production

Ethryan:
A GT6 style sap collection bag that you place on a cut log?

Dream-Master:
I remember the mod called IHL tools

NeutronSelector:
Regarding the textures, I completely agree they're not suitable at this stage. As I mentioned in the parent PR, I'm leaving the sound and texture aspects to other, more competent people.

These textures are awful in my opinion, but I didn't want to waste time on that.

As for suggestions to make harvesting more realistic, no problem. Perhaps we should have a validated vision before starting any changes. I simply reused the IC2 logic without reinventing the wheel.

YannickMG:
I propose we keep the scope of those mechanicsl changes extremely small for now so we can just get this into GTNH ASAP. That way it's no longer a blocker for IC2 removal, which is the real value we're after.

I like the idea of giving early game tools like the knife more uses. How about for now we keep the "naturally occurring sap spawning faces" mechanic and just add the Gregtech Knife as a supported tool for collecting the sap, on top of the IC2 Treetap? That way the tree is immediately usable in the pack, and a separate PR can fix the textures and if desired the mechanics.

NeutronSelector:
> I propose we keep the scope of those mechanicsl changes extremely small for now so we can just get this into GTNH ASAP. That way it's no longer a blocker for IC2 removal, which is the real value we're after.
> 
> I like the idea of giving early game tools like the knife more uses. How about for now we keep the "naturally occurring sap spawning faces" mechanic and just add the Gregtech Knife as a supported tool for collecting the sap, on top of the IC2 Treetap? That way the tree is immediately usable in the pack, and a separate PR can fix the textures and if desired the mechanics.

That was my vision, at least for the mechanics. As for the textures, if no one can create them quickly, that's certainly a possibility too. Adding the knife as a compatible tool is simple.

NeutronSelector:
Go check out the parent PR, I've revised the resin generation/harvesting gameplay to make it more realistic.

Issue #24353, opened 2026-04-21
Diamond spikes quest has incorrect prerequisites


pbullingtonmcguire:
### Your GTNH Discord Username

zutaca

### Your Pack Version

2.8.4

### Type of Questbook Change

Quest Amendment

### Quest Name

Diamond Spikes

### Quest ID

AAAAAAAAAAAAAAAAAAAFBQ==

### Your Proposal

The Diamond Spikes quest has the LV lathe quest as a prerequisite, but the lathe is not required to make it. Instead, the LV cutting machine should be a prerequisite, since you need a cutting machine to make the required diamond plates.

### Final Checklist

- [x] I believe this change would improve the overall questing experience.
- [x] I understand this suggestion may not be implemented.
- [x] I have searched the issue tracker and confirmed there is no similar issue.
- [x] I have confirmed this change has not already been implemented in the latest Questbook edition.

PR #24390, opened 2026-04-24
Add Cutting Machine as dependency for Diamond Spike

sudojunior:
Cutting Machine is set as required for the Diamond Spike quest, ~~Lathe currently remains as a hidden dependency, but may also be removed entirely as the quest will not unlock without it~~ - pictured below.

#24353

| Before | After |
|--------|-------|
| javaw_Tg4SDCacxo | oyjp9CRoEz |

~~If the Lathe dependency is removed, the arrangement of quests is expected to change, as only "Portable Mob Spawners" would remain a dependency of both branches.~~

image


UltraProdigy:
This is fine but just keep the lathe as a visible dependency. Sure it's not 100% required to make the spikes but it's the recommended method and 95% of players will take it

sudojunior:
Lathe quest has been returned to a visible dependency, and main PR description has been struck through regarding its initial details. No image uploaded, connecting line will no longer fade or disappear, dependencies remain unchanged.

UltraProdigy (approved):
Looks great. Ty for the PR!

PR #24461, opened 2026-04-28
Update betterloading screen lang

czqwq:


Nana-Sakura (changes requested):
ULV should not be considered as a tier

czqwq:
but why you write ULV into chinese lang?

Nana-Sakura:
> but why you write it ULV into chinese lang?

https://discord.com/channels/181078474394566657/234936569360809996/1297895964543160363

Nana-Sakura (approved):

PR #24508, opened 2026-04-30
Fix coins separation (Done up to LV)

OTPANNIEXD:
Oh god
The quests
They.... THEY ARE MULTIPLYING


UltraProdigy (approved):
Looks good! Thank you 🍾

PR #24546, opened 2026-05-03
Better description for infusion intercepter and small fertilizer quest fix

HectorHW:
The Thaumonomicon entry on this block is rather cryptic, referring the user to it is not very helpful IMO.

Added some more info about using the Intercepter to speed up the infusion process. Changed the spelling to match wiki and in-game name.

Also replaced probably accidentally added CR symbol in "Speed Up Your Plant Growth" quest description with LF. GitHub diff may not show it that well, but you can use `git diff -R` to see the removed CR symbol if you want.


UltraProdigy:
Does it need to be provided specifically to the bottom side of the block?

Ruling-0:
> Does it need to be provided specifically to the bottom side of the block?

As I recall, yes. The code specifically checks the block below.

UltraProdigy (approved):
Ty :)

PR #24666, opened 2026-05-12
Replace Artwork Main Menu Backgrounds w/NASA Photography

### LINKED ISSUE #24574: Change the AI generated main menu splash image

Rals

### Your Pack Version

2.8.4

### Your Proposal

One of the main menu images, bg7.png, was generated through AI. My proposal is that it should be changed.

Image

### Your Goal

Simply put, Ai usage in creative endeavors diminishes the quality of the project, even with the tremendous amount of effort by the developers made on the pack.

### Your Vision

By replacing the image with a community-made artwork, for example, it would make the pack feel like even more of a community-driven modpack, vastly improving its quality, and turning it into an even more beautiful love-letter to said community.

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying *I like this change please reconsider adding it* will prompt us to investigate and reopen it once we confirm your report.
- [x] I understand this change request may not attract enough attention and thus not be implemented.
- [x] I understand this change request may be rejected due to other community members think it's inappropriate.
- [x] I believe this feature would make the pack better.

PlayfulPiano:
Had a bit of discussion regarding this over in the discord, including some back and forth about image options. This PR replaces the existing main menu backgrounds (and the single AI generated one) with photos taken from NASA (we are able to use them for non-commercial use, see: https://www.nasa.gov/nasa-brand-center/images-and-media/). Twenty three of them are included (plus the iconic background as requested by Dream), and with that I also enabled the "shuffle" setting for CustomMainMenu (changes order every client instance iirc), as well as mentioning NASA as the background image source. The text color for the menu buttons have also been adjusted slightly to be made consistent for the "mods" button, and with hopefully a bit better visibility than before.

Separately thanks to @MalTeeez, the background for the menu buttons is now translucent to improve visibility for some of the backgrounds.

Moon and Earth photos in particular are sourced from the recent Artemis II mission!

Closes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/24574

Showcase of all backgrounds:

<details>
<summary>Backgrounds</summary>
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
image
</details>


config/txloader/load/custommainmenu/lang/en_US.lang
```diff
@@ -11,3 +11,4 @@ menu.gtnh.forge=Forge #forgeversion# for Minecraft #mcversion#
 menu.gtnh.mojang=Copyright Mojang AB. Do not distribute!
 menu.gtnh.aviators=Opening Music Copyright by the Aviators
 menu.gtnh.sol=Absolute Zero by Sol IX
+menu.gtnh.nasa=Background Photos From NASA
```
tiffit:
```suggestion
menu.gtnh.nasa=Background Photos from NASA
```
very minor but i think it reads better

Dream-Master:
Yes, the new photos look great and thanks for the PR, but I have three points to raise.

In GTNH, we don’t just travel within our solar system but also to alien worlds (New Horizons) so we shouldn’t include fictional planets or constellations?

In addition, the main image has been a hallmark of Gtnh for years and is therefore very iconic. I'm not sure how much of its identity would be lost if it were replaced ?

And third, we need to discuss about the image size and optimization. One file is 45 MB big.

PlayfulPiano:
> Yes, the new photos look great and thanks for the PR, but I have three points to raise.
> 
> In GTNH, we don’t just travel within our solar system but also to alien worlds (New Horizons) so we shouldn’t include fictional planets or constellations?
> 
> In addition, the main image has been a hallmark of Gtnh for years and is therefore very iconic. I'm not sure how much of its identity would be lost if it were replaced ?
> 
> And third, we need to discuss about the image size and optimization. One file is 45 MB big.

For the first point, there are exoplanet artworks/illustrations from NASA but I'm less sure about their ability to be used non commercially compared to the direct photography of like Webb or Hubble. I can definitely add some in later if they're good to use. But the whole alien worlds aspect is in part why we have the galaxy and nebulae photos included.

Which is the main image for gtnh? If we are including the exoplanet stuff I can definitely add those photos back, at least the non AI ones.

Yeah that's understandable with the file size concern. Should I change the files to pngs from jpgs so the image optimization job could work, and if so, should I just like rename the file type or go through some online file converter?

PlayfulPiano:
Ok double checking, looks like NASA does use some third party copyrighted material but that should be properly identified. So as long as those are avoided we should be good.

UltraProdigy:
Typically jpegs will have a lower size than PNGs just due to the way they are stored so converting these likely won't make them smaller. It would be best to just lower the resolution here.

Also, if you are ever converting just changing the file type does nothing to actual image data you'll want to use a converter.

PlayfulPiano:
> Typically jpegs will have a lower size than PNGs just due to the way they are stored so converting these likely won't make them smaller. It would be best to just lower the resolution here.
> 
> Also, if you are ever converting just changing the file type does nothing to actual image data you'll want to use a converter. 

Ty TIL, I'll try downloading lower res/sized versions instead and replace them when I get home later today.

PlayfulPiano:
PR updated. Thanks to @MalTeeez especially for getting the menu bg to be translucent so it's easier to see the buttons with the new backgrounds. File sizes are reduced to all be under 5MB per photo.

Also regarding the older images, I feel less inclined to include the stock ones for the same reason why I am hesitant to include exoplanets from NASA: for both cases these are artist renditions, rather than direct captured photography. It would feel a bit misleading to mix both real photographs and artistic renditions together, at least to me.

Dream, if the iconic background you're referring to happens to be one that could be replicated by like, an actual photo, I can try to add that in here if that's an alright compromise. Plus the idea of foreign worlds is sort of visualized by the galaxy and nebula photos imo.

Dream-Master:
@PlayfulPiano add back the one backround i mentioned before and we an merge this

PlayfulPiano:
> @PlayfulPiano add back the one backround i mentioned before and we an merge this

Gotcha. @MalTeeez is rn helping me improve and make the image quality more consistent across the images, so once that's done and the commit is made I'll let you know.

Dream-Master:
@PlayfulPiano ok

PlayfulPiano:
@Dream-Master ready for merge, also please add @MalTeeez as a co-author for this PR, ty <3

Dream-Master (approved):

Ranzuu:
Can we make this actually readable please

image

PR #24700, opened 2026-05-13
Deletes old config

OTPANNIEXD:
Deletes old unused config file


koolkrafter5:
Are you sure this is unused? Isn't this what sets the recipes for the inscriber? Are they set somewhere else instead?

tiffit:
This config is used. It adds the recipe to turn certus quartz plates into circuits in the inscriber. Removing the file also removes that recipe from the game (as AE2 falls back to default recipes)

PR #24706, opened 2026-05-14
Add Paintbrush to Making Sense of Spaghetti Quest

NotKaizul:
Title. Also a couple minor fixes:
- Added LV colour to toolbox quest's title
- Fixed the listed minimum distance between boosters for Gotta Go Fast quest
- Switched out Making Sense of Spaghetti to Playing with Colors in LV tab since it no longer requires an assembler


UltraProdigy (changes requested):
I don't love having both the spaghetti quest and playing with colors. I think we should combine the information and unify the quests assuming all of those paint methods work fine on hatches.

NotKaizul:
They do afaik so that should be fine. Should I remove the ebf dependency then? Don't want the quest to require it when paintbrushes are craftable in stone
Or I could switch that out for a checkbox task as a reminder, similar to the quest in zpm

UltraProdigy:
> They do afaik so that should be fine. Should I remove the ebf dependency then? Don't want the quest to require it when paintbrushes are craftable in stone Or I could switch that out for a checkbox task as a reminder, similar to the quest in zpm

Yeah good idea, let's do both

NotKaizul:
For some reason this is deleting the ebf quest in multiblock goals. ig i fucked something up, but idk what :P

PR #24734, opened 2026-05-16
Restore blue machine tint in default config

Zorbatron:
Restores the default machine color tint changed in #24698.
| Before (`0xFFFFFF`) | After (`0xD2DCFF`) |
| --------- | -------- |
|  |  |


Auynonymous (changes requested):
This pr should instead change the colors of the base textures to have the hue change built in. Keeping the config at 255 255 255 allows for better color customization.

Auynonymous:
This should be an easy fix, it seems like those hull textures were never updated alongside the config change.

@ABKQPO can you implement those textures changes to go along side that tinting removal?

Zorbatron:
If the base textures were tinted instead, wouldn't that make it impossible to have the machines look like the "before" image? 
*Without using a texture pack

Auynonymous:
> If the base textures were tinted instead, wouldn't that make it impossible to have the machines look like the "before" image? *Without using a texture pack

This would be correct yes, however the look of machines was always that of the right. So for vanilla pack's sake, leaning for the look on the right is best. Besides this tint actually breaks some resource packs on accident, some that color casings differently will all have the blue tint applied to them. Most users do not know to go to the config and set the tint to 255 255 255.

Zorbatron:
Makes sense, I'll close this PR in favor for the subsequent base texture change.

PR #24735, opened 2026-05-16
Fixes skip-quests

OTPANNIEXD:
Fixes skip quests to require 64 items, unless the item is rubber. When item is rubber it needs 128. There is also exception left for NaK coolant (2 needed) and for netherite (1 dense plate or 64 nanoparticles or 16 ingots)


LazyFlesh:
> When item is rubber it needs 128 

it was 125 items when it was rubber/plastive, becuase they were made in batches of 1500, which the first round number divisible by both 1500 and 144 was 18,000, or 125 items

UltraProdigy:
You're correct, however those weren't changed

UltraProdigy (approved):

PR #24740, opened 2026-05-16
Fix arc furnace quests

OTPANNIEXD:
All should be done
It rearanges the tab cuz someone put ABS material in controller (I blame @kuba6000 )
New look of the tab:
image
New arc furnace quest:
image
image
image
image


config/betterquesting/DefaultQuests/Quests/MassProcessing-AAAAAAAAAAAAAAAAAAAAEQ==/BaskInTheCurrent-AAAAAAAAAAAAAAAAAAAH7Q==.json
```diff
@@ -1,22 +1,22 @@
...
... to the modes, this machine has 3 different processing phases\n\n§§§lStart§lu§r§l§lp§r§r§l:§r§r§l§r§r §r§rDuring this phase the machine will use power equal to §lmaximum parrarels * number of amps per parrarel * \n(1+ startup penalty). §rThis phase lasts for 6 seconds before going to processing phase.\n\n§lProcessing:§r During this phase the machine behaves as any other multiblock. Its power usage is determined by the Amp/Parrarel statistic on the electrode that is being used. If the machine fails to find a recipe that it can use it goes into shutdown phase.\n\n§lShutdown:§r Machine enters a 6 second phase during which it does not process any recipes and does not use any power. After shutdown phase to start processing recipes you need to go through startup again.\n\n\nStatistics of this multiblock are mostly dependant on electrode that is being used. This ranges from speed, parrarels, OC\u0027s and even ammount of Amps per parrarel.\n\nIf the used electrode drops below 30% durability there is 5% chance that the machine will have to go through startup phase again.\n\nIf the used electrode drops below 10% durability there is 2% chance that the machine will §lvoid ALL§r outputs \n\n",
```
UltraProdigy:
```suggestion
      "desc:8": "§r[warn]This multiblock always runs recipes at the same tier as the energy hatches in the structure. If the machine does not have enough power supplied, it won\u0027t turn on.[/warn]\n\nAlthough this multiblock is available in IV, we recommend building one §nafter finishing the Naqline§r.\n\n\nThis multiblock has 3 different modes: §3Normal§r, §4EBF§r, and §6Ore§r.\n\n\nIn §3Normal§r mode, this multiblock §3processes normal arc furnace recipes§r.\n\nIn §4EBF§r mode, this multiblock can process §4every EBF recipe regardless of coil tier requirement§r. This feature comes at the cost of §lbase recipe voltage being increased by 2 tiers§r.\n\nIn §6Ore§r mode, this machine can only process ores and raw ores that can be smelted in a normal furnace, giving molten metal directly.\n\n\nIn addition to the modes, this machine has 3 different processing phases.\n\n§lStartup:§r During this phase, the machine will use power equal to §lmaximum parallels * number of amps per parallel * (1 + startup penalty)§r. This phase lasts for 6 seconds before going to the processing phase.\n\n§lProcessing:§r During this phase, the machine behaves as any other multiblock. Its power usage is determined by the amps/parallel statistic on the electrode that is being used. If the machine fails to find a recipe that it can use, it goes into the shutdown phase.\n\n§lShutdown:§r The machine enters a 6 second phase during which it does not process any recipes and does not use any power. After the shutdown phase, it needs to go through startup again before processing recipes.\n\n\nStatistics of this multiblock are mostly dependent on the electrode that is being used. This ranges from speed, parallels, OCs, and even the amount of amps per parallel.\n\nIf the used electrode drops below 30% durability, there is a 5% chance that the machine will have to go through the startup phase again.\n\nIf the used electrode drops below 10% durability, there is a 2% chance that the machine will §lvoid ALL§r outputs.",

```

config/betterquesting/DefaultQuests/Quests/Tier1LV-AAAAAAAAAAAAAAAAAAAAAw==/BasicArcFurnace-AAAAAAAAAAAAAAAAAAAClg==.json
```diff
@@ -8,7 +8,7 @@
   "properties:10": {
     "betterquesting:10": {
       "autoClaim:1": 0,
-      "desc:8": "The arc furnace is an alternative to macerating stuff back into dusts and for making wrought iron, annealed copper, and glass (in MV). You get ingots, and sometimes different materials back from recycling than with a macerator. You also have all 4 slots at LV. The recipes use oxygen and 3A of power.\n\nJust ignore the plasma one.\n\nAnd no, you can\u0027t just send 1A of MV. That will blow it up.",
+      "desc:8": "The arc furnace is an alternative to macerating stuff back into dusts and for making Cast Iron, Annealed Copper, and Glass (in MV). You get ingots, and sometimes different materials back from recycling than with a macerator. You also have all 2 slots at LV.\n\n[note]You can also use certain gases to boost your arc furnace processing speed.[/note]",
```
UltraProdigy:
```suggestion
      "desc:8": "The arc furnace is an alternative to macerating stuff back into dusts and for making Cast Iron, Annealed Copper, and Glass (in MV). You get ingots, and sometimes different materials back from recycling than with a Macerator. You also have both slots at LV.\n\n[note]You can also use certain gases to boost your Arc Furnace processing speed.[/note]",
```

UltraProdigy (changes requested):
Just grammar / spelling

UltraProdigy (approved):

PR #24764, opened 2026-05-17
Better main menu text color

PlayfulPiano:
Should've done this from the start, my bad. Text color now matches how it was prior to my preivous PR, wth the "mods" color matching to the lighter blue.


PlayfulPiano (commented):
ok maybe just don't try to make a PR directly from vscode

PR #24804, opened 2026-05-21
Add Paintbrush to Making Sense of Spaghetti Quest

NotKaizul:
Doesn't look like it's deleting the EBF quest this time 😃

Updated Quest:

I didn't include the OpenBlocks brush cause its crafted with the BuildCraft one, and has the same functionality but less durability (so no reason to use it)


UltraProdigy (approved):
I liked the LV changes you made earlier so I added that back in.

Also, mentioned this elsewhere but be sure to mention all of your changes in the description. It makes it much easier for me to maintain a changelog and do reviews. Non-listed changes are often mistaken for accidental or sneaky.

Looks great though, thanks for the PR :)

PR #24887, opened 2026-05-25
Restored Red Bed (Icon) for SO...TIRED...MUST...SLEEP... Quest

StoneLegion:
We can't have the white one it feels wrong!


Dream-Master:
It's only the icon right ? @StoneLegion

StoneLegion:
> It's only the icon right ? @StoneLegion

Correct, I did NOT touch the oredict, etc.

Dream-Master (approved):

PR #24952, opened 2026-05-30
Move betterquesting lang files from txloader/load to txloader/forceload

Eldrinn-Elantey:
Moved config/txloader/load/betterquesting/lang to config/txloader/forceload/betterquesting/lang so quests are force-loaded by txloader.


Ethryan (commented):
lang/lang/template.lang?
double lang folders.

Ethryan (approved):

PR #24999, opened 2026-06-02
Change Stats for Lootbag Seeds

Dynamiczbee:
Unless there is a way to randomize NBT when lootbag drops them, by making them all 10/10/10 they can act like a kickstart. Not totally useless (given the ease of getting 31/31/31) given they're through the worst part at the start.

https://discord.com/channels/181078474394566657/1303129128085753898/1511158931836240013


UltraProdigy (approved):
This feels like a good change to me

PR #25016, opened 2026-06-03
Remove magic frame and soul comb qb rewards

LazyFlesh:
title, to go along with https://github.com/GTNewHorizons/GT5-Unofficial/pull/6926

Did this through dev env, not ingame editing. Idk how this works for qb, so @UltraProdigy, if this is wrong, please advise.


UltraProdigy:
@LazyFleshWasTaken Why are we removing them entirely instead of lowering their chances?

LazyFlesh:
> @LazyFleshWasTaken Why are we removing them entirely instead of lowering their chances?

As neoknou had pointed out, it's currently an 80% chance to get  magic frame, wax, or a necrotic frame, whose quest gives magic frames, and therefore a punishment for the unlucky.

Lowering the chance doesn't fix that feeling, and people would still be rolling for them, even if they have to roll backups, because "you used to be able to get it" or smth. It's not a sentiment I haven't seen, or felt.

It also makes it line up more with how tech tiering works for skips for a mod, like id mentioned in the main pr. We already have oblivion frames, for rng-based bee help, anyway.

UltraProdigy (approved):
Makes sense

PR #25074, opened 2026-06-06
Add mechanical armor quests

FourIsTheNumber:
Explains mechanical armor and walks through all the tiers and augments.

image


UltraProdigy (approved):
Yay!

PR #25099, opened 2026-06-07
update iguanatweaks config

LazyFlesh:
Needs: https://github.com/GTNewHorizons/IguanaTweaksTConstruct/pull/35

implements the logic i discussed in meta dev yesterday. Modifies it from only increasing the level of a bronze-level tool to diamond-level to instead increasing the mining level by one per diamond/emerald applied, up to diamond/bronze. 

Applies durability boost only once.

Balance concerns: low. 
There are many ways to skip mining levels, including just making a gt diamond hammer to mine obsidian with (cost: 6 diamonds) or making/finding/buying a thaumium pickaxe (needs thaumcraft research, and source of thaumium), or going through a roguelike for higher tier materials, like cobalt or ardite, or alloys of them like Queen's gold.

Tinkers' isn't isolated enough to be able to enfornce mining level progression by itself.


Miklebe (approved):
:shipit:

PR #25207, opened 2026-06-11
Fix OC notation in questbook

Pxx500:
Adjusts quests to reflect the common community notation (Speed/Power)


UltraProdigy:
I've gotten several complaints for using "OC" which is why I like to write out overclock. I think it would be wise to continue the pattern. Though I'm sure I'll also get yelled at for the numbers :trollface:

Pxx500:
I think I explained it pretty well in the multiblock revolution quest, so should be fine

UltraProdigy:
Sure, I prefer OC as well anyways. I'll wait to see if anyone says something.

UltraProdigy (approved):

PR #25491, opened 2026-06-22
Server Migrate

MassAnarchyy:
Migrating to Paradox from challenged gamers, also going to be self-host by end of the month (waiting on last of hardware to arrive) for long-term stability =)


Dream-Master:
@MassAnarchyy are this change need to be done now or can set to draft until ist ready for a merge ?

Dream-Master (approved):

PR #25494, opened 2026-06-22
Fix eggplant damaging chickens when consumed

zaru-noodles:
Config update for https://github.com/GTNewHorizons/AngerMod/pull/19


UltraProdigy (approved):
Ty!

PR #25524, opened 2026-06-25
ci: bundle GTNH-Guide-Pack into nightly releases

Caedis:
Auto pulls the latest GTNH-Guide-Pack release into the release zip (not in the repo itself)
Skips if the config repo tag does not have "nightly" in it (stable/beta configs are handled manually)
If the config repo tag is -pre, it pulls the latest release, even if its -pre.
If the config repo tag is not -pre, it pulls the latest non-pre release


UltraProdigy (approved):
If I'm being super nitty, using `createdAt` here instead of `publishedAt` could lead to improper version selection if we ever started using draft releases, but I don't think I've ever seen us use that feature once here so that is more than likely inconsequential.

Looks good 😎

PR #25539, opened 2026-06-26
Niobium-Titanium-Quest-Reworking

KraVoid:
image

Made Titanium-Niobium Quest be "MAIN"

image

Deleted Niobium-Titanium Quest from "IV Age"

image

Rewrite description by a bit.


UltraProdigy:
This is great, can you also mention that it needs TPV coils to heat and set this quest as a prereq for the IV superconductors and crystal processors quests. Then on the Niobium Titanium quest can you also put a dependency on TPV coil quest

UltraProdigy (approved):
Thank you!

PR #25571, opened 2026-06-28
Keep questbook config on pack upgrade

wlhlm:
I have to reconfigure my questbook on every pack update (most notably the theme for me). This change tries to preserve the questbook settings.

Additionally, GTNH 2.9 will add many new config options for quest notifications and I think it's appropriate to keep these on update, so it's a good time to add this to `localconfig.cfg`.


UltraProdigy (approved):
Might need to change this from a wildcard to specifics later but should be good for now

PR #25847, opened 2026-07-13
Update Legacy Ores Resource Pack

koolkrafter5:
Adjust to changes made in https://github.com/GTNewHorizons/GT5-Unofficial/pull/7314. Adds the compress script from https://github.com/GTNewHorizons/BloodMagic/pull/139 to automatically rebuild this RP whenever changes are made to the raw files to make it easier to keep updated. Also change the pack icon to not be the same as five of the other RPs we ship.
image
ores


UltraProdigy (changes requested):
This looks good but can you add the raw folder to the exclusions in release-tags

UltraProdigy (approved):
💰

PR #26552, opened 2026-08-31
add dust for the ore tasks

Nana-Sakura:
## Summary

currently for iv players, they still need to get crushed ores for tasks, which is dumb for those who have automated ore proc
effects:
Europa:
Screenshot 2026-09-05 at 17 39 11

Callisto:
Screenshot 2026-09-05 at 17 42 49

Ceres:
Screenshot 2026-09-05 at 17 41 15

Venus:
Screenshot 2026-09-05 at 17 42 27

Mercury:
Screenshot 2026-09-05 at 17 40 35

Io:
Screenshot 2026-08-31 at 19 38 32

## Checklist
- [ ] I have tested this PR in DevEnv
- [x] I have tested this PR in Fullpack
- [x] This PR is in compliance with the [GTNH AI Policy](https://github.com/GTNewHorizons/GTNH-Dev-Doc/blob/master/AI_POLICY.md)
- [ ] This PR requires another PR in order to merge


UltraProdigy:
Maybe on the Naquadah but I'm not fond of this change because it lowers the space/meteor restriction they sit behind. The rewards were designed around them being GC quests.

I don't mind naquadah because they should not prevent progression. But they're bonus quests so they should require effort.

Nana-Sakura:
> Maybe on the Naquadah but I'm not fond of this change because it lowers the space/meteor restriction they sit behind. The rewards were designed around them being GC quests.
> 
> I don't mind naquadah because they should not prevent progression. But they're bonus quests so they should require effort.

For these changes:
Ledox: Can be acquired from pure bee infra, but needs mutatron(LuV), atp the task reward would be useless; else needs gc. Cannot be acquired from crops. Crops can only provide 4x bonus.

Callisto ice: bee part same as above, for Crops, needs Space Flower, requires gc blocks, but could bypass t3. Anyways not bad.

Naquadah: can be acquired from far end asteroids, which is T0, so basically no influence.

Quantium, Titanium, Naquadah: The quest is asking for these 3 stuffs, only can finish the latter 2 without gc or higher tier tech. For quantium, crops part is same as ledox, for bees, needs gc or mutatron(LuV).

Infused gold, niobium, samarium:
Infused gold could be bypassed by infusion lol but for samarium, the only way is gc or meteor, and if without t3/T4, then requires the monaline. Maybe change to require Samarium conc?

Deep iron: same case as ledox.

So I think they are not that bad as your thoughts lol

UltraProdigy (dismissed):
Let's get rid of the IV lootbags in these quests then we can move forward

Nana-Sakura:
> Let's get rid of the IV lootbags in these quests then we can move forward

done.

Dream-Master (approved):

PR #26843, opened 2026-09-16
Allow floral fertilizer skip for botania initial task

fehling135:
## Summary
As the title said, players can get to floral fertilizers from witchery or draconium ores, no need for mushrooms.

## Checklist
- [ ] I have tested this PR in DevEnv
- [ ] I have tested this PR in Fullpack
- [x] This PR is in compliance with the [GTNH AI Policy](https://github.com/GTNewHorizons/GTNH-Dev-Doc/blob/master/AI_POLICY.md)
- [ ] This PR requires another PR in order to merge


UltraProdigy (changes requested):
Let's keep the pre-req. Also just make the OR task for a flower instead of the fertilizer and the added note can dictate how the flower can optionally be attained before the mushroom via the floral fertilizer

fehling135:
what about adding flower as a skip-task? to lift the limit which players should go to twilight then start botania.

UltraProdigy:
Messaged in quest dev, but not sure how flowers would be unlocked pre-twilight here

fehling135:
> Messaged in quest dev, but not sure how flowers would be unlocked pre-twilight here

With witchery u can get flower fertilizer without twilight need, as I described in title

Nana-Sakura (approved):

